PR #1, opened 2016-02-28
Refactor/GitHub migration

itsrachelfish:
These changes remove all of the old unused subprojects and make it possible to simply clone this repo to start using OpenKore.

Also changed ```loadPlugins``` to ```2```, only allowing a whitelist of plugins to be loaded by default.


allanon:
Looking great! Re-organize plugins, or at least put the macro plugin in the right spot to get loaded, since it's enabled?

PR #3, opened 2016-02-28
Refactor/macro plugin

itsrachelfish:
As per @allanon's comment in my first pull request (#1) I've updated the plugins folder to load the macro plugin by default. All other plugins have been moved into a "needs-review" folder and the tests for the macro plugin have been moved to a "tests" folder in the root of the project.


jefferson-dab:
It's probably a good idea to move the macro plugin to its own subfolder like:

```
plugins/
- macro/
-- macro.pl
-- macro/
- needs-review
-- ICU/
-- [...]
-- xconf/
```

To avoid a messy folder as more plugins get reviewed.

allanon:
Putting `macro.pl` in `plugins/macro` works (because the code recursively scans the plugins directory exactly one level deep), and groups the files together. Seems like a good suggestion to me. Most plugins probably don't need to be in their own directory, but macro is special in that it has several perl modules in separate files.

itsrachelfish:
I just tested locally and it looks like @allanon is right. OpenKore will look in any first-level directory to load plugins so long as the filename matches.

For example, this is my sys.txt

```
loadPlugins 2
loadPlugins_list macro,test
```
Then I put the macro plugin inside a folder called "SUP_GUYS"...

```
Selectively loading plugins...
Loading plugin plugins/test/test.pl...
Loading plugin plugins/SUP_GUYS/macro.pl...
```

I was not aware of this behavior originally, so I'll make a new pull request using folders instead. 

I looked through several of the old plugins and there are quite a few with other files besides the main plugin. Documentation, SQL files, unit tests, etc.

PR #4, opened 2016-02-28
Add client for RagexeRE_2013_12_23 to fix decryption of 'area_spell' packets

kliewkliew:


itsrachelfish:
@kliewkliew Cool! :+1: 

What server is this for / how were these files generated?

kliewkliew:
I tested these on my own server using the Hercules server emulator. I made these by matching known values (ie. spellcast x-location) to values in the Wireshark hex stream. Then I constructed the decode string by using the offsets of the hex values.

ya4ept:
see:..\Receive\ServerType0.pm:
> '09CA' => ['area_spell_multiple3', 'v a*', [qw(len spellInfo)]], # -1

kliewkliew:
It looks like I was working on an old version. It seems to have been fixed in https://github.com/OpenKore/openkore/commit/8ca404f4e73909f2dd2ab91fa33b62ed9c2935cf

Slipxxliose:
It will only works inside official. private severs will warn about it.

PR #6, opened 2016-02-29
fix #1 detect git revision

allanon:
- [x] Code Review
- [x] Test on Windows
- Re-build start.exe and wxstart.exe (unnecessary, because we're no longer adding any new module requirements)

Git version detection is actually pretty easy... but the hash isn't very informative by itself. Getting the commit date is much harder... but seems like a useful way of communicating whether the code is massively out of date.

This pull request generates a git revision which looks like SHA_YYYY-MM-DD. For example:
```
$ ./openkore.pl -v
*** OpenKore what-will-become-2.1 ( version 436478c_2016-02-29 ) - Custom Ragnarok Online client ***
***   http://www.openkore.com/   ***
```

### DIFF
--- a/openkore.pl
+++ b/openkore.pl
@@ -104,7 +104,11 @@ sub __start {
 sub parseArguments {
 	eval {
 		if (!Settings::parseArguments()) {
-			print Settings::getUsageText();
+			if ($Settings::options{version}) {
+				print "$Settings::versionText\n";
+			} else {
+				print Settings::getUsageText();
+			}
 			exit 1;
 		}
 	};
--- a/src/Git.pm
+++ b/src/Git.pm
@@ -1,0 +1,138 @@
+package Git;
+
+use strict;
+use warnings;
+
+use IO::Uncompress::Inflate qw( inflate );
+
+sub detect_git {
+    my ( $file ) = @_;
+
+    my $git_dir = find_git_dir( $file );
+    return if !$git_dir;
+
+    my $fp;
+    my $head_file = open( $fp, '<', "$git_dir/HEAD" ) && <$fp> =~ /ref: (.*)/ && $1;
+    return if !$head_file;
+
+    my $sha = open( $fp, '<', "$git_dir/$head_file" ) && <$fp>;
+    return if !$sha;
+    chomp $sha;
+
+    my $timestamp = commit_timestamp( git_get_commit( $git_dir, $sha ) );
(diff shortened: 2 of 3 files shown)


jefferson-dab:
Just tested it on Windows. 

Running with Strawberry Perl works fine. But there's a missing dependency problem when running with the binaries "start.exe" and "wxstart.exe". 

```

Can't locate IO/Uncompress/Inflate.pm in @INC (@INC contains: src src/deps C:/Us
ers/Jefferson/Desktop/openkore-master/src/deps C:/Users/Jefferson/Desktop/openko
re-master/src C:/Users/Jefferson/Desktop/openkore-master .) at src/Git.pm line 6
.
BEGIN failed--compilation aborted at src/Git.pm line 6.
Compilation failed in require at C:/Users/Jefferson/Desktop/openkore-master/src/
Settings.pm line 551.
BEGIN failed--compilation aborted at C:/Users/Jefferson/Desktop/openkore-master/
src/Settings.pm line 551.
Compilation failed in require at openkore.pl line 33.
BEGIN failed--compilation aborted at openkore.pl line 33.

```

allanon:
Now to figure out how to rebuild `start.exe` and `wxstart.exe`. :)

itsrachelfish:
Are you sure we'd need to rebuild ```start.exe```? Can't we just include those libraries into the ```src/deps/```?

allanon:
I believe we would need to rebuild the exe, because `IO::Uncompress::Inflate` uses zlib, which is a dll on Windows. Can we put the dll in `src/deps`? I'm woefully ignorant when it comes to Windows development. That module is also part of the core libraries on more recent Perls, so just updating the Perl that `start.exe` is built with would probably be enough.

allanon:
Also, it looks like `Compress::Zlib`, which is an older zlib library (already in use by `Field`, for decompressing `.fld.gz` files) can do the work too. It's more difficult to use because I can't just pass it a filehandle, but that might get rid of the  `IO::Uncompress::Inflate` requirement and save us from having to rebuild `start.exe`.

allanon:
Ok! This pull request now uses `Compress::Zlib`, so we shouldn't be adding any new module requirements.

itsrachelfish:
Tested the newest version on Linux and Windows, everything looks good. Very cool implementation of parsing the git history without a library!

PR #12, opened 2016-03-01
Fix tests

allanon:
- [x] Code Review

Fix `src/test/unittests.pl`. To run it, do `(cd src/test && ./unittests.pl)`.

There are a few fixes here:
* The smartmatch operator `~~` was used, and that breaks perl 5.8, which I use, so it broke tests (for me).
* Some FileParsers tests were breaking because the `1` in `"Bowman Scroll 1" 2` was not being recognized as part of the item name. This feature appears to be aspirational: it never worked as far as I can tell.
* The Network tests were breaking because kRO has more than 100 base classes. And twRO uses kRO.
* The Network tests were also breaking because the idRO and pRO server types do bad things. As pRO is definitely closed (not sure about idRO), fixes seem impossible to test, so just ignore those server types.
* The eAthenaTest wasn't even compiling properly. If anybody cares about the things it did, they should probably re-write it.


itsrachelfish:
Ran the tests locally to compare master and your branch, everything looks good.

Not sure if I like "Bowman Scroll 1" [being added into the test verbatim](https://github.com/OpenKore/openkore/pull/12/files#diff-d6ad721d9dbb61762a2a08347512c8dbR64) instead of something more generic, but oh well.

PR #13, opened 2016-03-01
add support for !include in all files parsed by Utils::TextReader

allanon:
- [x] Code Review

Note that the old unit tests won't pass until https://github.com/OpenKore/openkore/pull/12 is merged.


allanon:
Open question: Should we make failure to find or read a config file a non-fatal error? Right now a missing file causes a crash.

itsrachelfish:
If we end up gitignoring config.txt, crashing without it would be a pain!

itsrachelfish:
Code looks good, tests passed for me. :+1:

PR #15, opened 2016-03-01
Removing autoupdate script/binaries

jefferson-dab:
These files no longer work, since they depended on the SVN repository.


allanon:
It would be nice to provide a feature like this in the core openkore code... perhaps via a plugin? It would have to run `git pull` and restart openkore if any changes were made.

PR #17, opened 2016-03-01
Split config files

allanon:
- [ ] Product Review
- [x] Code Review
- [x] Merge https://github.com/OpenKore/openkore/pull/13 (without which, this pull doesn't work)

Separate `items.txt` into three different files:
* `items-official.txt` contains a nearly exact copy of the file from the GRF (except that we have to clean up the broken UTF-16).
* `items-openkore.txt` contains openkore modifications, intended to allow us to disambiguate items which have identical names.
* `items.txt` includes the other two, and provides a place for local modification without having to change either the official version or the openkore version.

Assuming that this is approved, I would like to:
* Add a .gitignore for `items.txt` so that local modifications don't show up as changes to commit.
* Split up other config files for the same reasons.


itsrachelfish:
Great commits! A+++++++++ would merge again

is this an acceptable product review? :smile:

itsrachelfish:
On a more serious note, if / when the files openkore writes are added to the gitignore, wouldn't we have to do this same update for all of the other servers as well?

ya4ept:
Friends, I do not understand why we need to split files in "tables" folder?
how did you get the file "recvpackets-openkore.txt" ???
in this folder should be only the files from the official client ..

allanon:
Here's my rationale for this change: https://github.com/OpenKore/openkore/issues/25

PR #19, opened 2016-03-01
fields storage format (.fld -> .fld.gz)

ya4ept:
OpenKore supports gzip-compressed field files (.fld.gz).I suggest fields file is stored in compressed form. Size OpenKore is reduced by 70 MB!


allanon:
Oooo, I like a 70 MB size decrease.

PR #23, opened 2016-03-02
Untracked local tables

allanon:
- [ ] Code Review

Move local changes into files untracked files (ie, files which are not checked into git). This gives us three places to make changes.

For example, by priority:

1. `items-custom.txt` - local modifications which shouldn't ever be checked into git
2. `items-openkore.txt` - openkore modifications to fix things like whitespace, spelling, and non-unique names
3. `items-official.txt` - an exact copy of the upstream version of this file (ie, extracted from the official client's GRF file)

The `items.txt` file itself should never change unless we want to include another file for some reason.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,12 @@ src/auto/XSTools/utils/perl/HttpReader.cpp
 src/auto/XSTools/utils/perl/Rijndael.xs.cpp
 src/auto/XSTools/utils/perl/Whirlpool.c
 src/test/http-reader-test
+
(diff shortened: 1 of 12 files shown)


allanon:
Ugh, as it turns out we need more changes to make this work: if `portalsLOS.txt` doesn't exist anywhere, it won't be written anywhere.

allanon:
Closing this pull, splitting it into two pieces, and removing the change to `recvpackets.txt`, as how exactly it should be managed is unclear.

Issue #25, opened 2016-03-02
Managing tables file updates is a pain


allanon:
* The official maintainers of this game don't always make sane decisions about what goes in the GRF files. They put random utf8 characters in there and don't notice because the client doesn't render them. They put spaces after item names. They also seem to be somewhat unaware of how to spell and the rules of grammar.
* The official maintainers of this game consider icons to be of equal value to text when deciding what to name things. OpenKore, on the other hand, uses item names for item management. While item IDs _can_ be used in config files, they aren't nearly as easy to read. And macros can't easily select items based on item id. And humans looking at their inventory may have difficulty understanding the difference between two stacks of "Transparent Plate".

OpenKore can fix these things, but managing changes by multiple groups in a single file is difficult. Also, if the end user decides to make changes, they'll end up with merge conflicts which they may not notice, or may be unable to fix, preventing them from updating in the future.

To avoid this problem, I suggest we split the iRO files into four different files, each of which has a separate purpose. Taking `items.txt` as an example, we read three files containing item definitions, with the following priority:

1. `items-custom.txt` - local modifications which shouldn't ever be checked into git, openkore users can put their changes here and not worry about having to deal with merge conflicts when they update the code from git
2. `items-openkore.txt` - openkore modifications to fix things like whitespace, spelling, and non-unique names
3. `items-official.txt` - an exact copy of the upstream version of this file (ie, extracted from the official client's GRF file)

The `items.txt` file itself just includes these three files, and should never change unless we want to include another file for some reason.

Other servers may choose to follow this example, or not, as they please.

allanon:
The core feature here is that git updates can be done without worrying about merge conflicts. Which means any local changes must be in a file which isn't checked into git. And must be referenced by a file which _is_ checked into git.

An alternative layout which preserves that core feature would be:

* `tables/iRO/items.txt` - include `tables/iRO/official/items.txt`, then have openkore modifications (which are currently in `tables/iRO/items-openkore.txt`), then include `tables/iRO/items-custom.txt`
* `tables/iRO/official/items.txt` - an exact copy of the upstream version of this file (ie, extracted from the official client's GRF file)
* `tables/iRO/items-custom.txt` - local modifications which shouldn't ever be checked into git, openkore users can put their changes here and not worry about having to deal with merge conflicts when they update the code from git

This strategy would reduce the number of files by one, and move the official file into a subdirectory so users don't have to think about it.

itsrachelfish:
[...]

PR #27, opened 2016-03-03
support !include_create_if_missing, use it to allow for user-specific…

allanon:
… untracked files

- [x] Code Review
- [x] QA testing

Fix https://github.com/OpenKore/openkore/issues/25

This pull request also undoes the splitting of `recvpackets.txt`, as it's not clear that there's an "official" version of the file.


itsrachelfish:
Code looks good, QA tested by logging into kore and checking if the custom files were created. Tests passed as well :+1:

itsrachelfish:
@allanon  By the way, I noticed you've been putting your new tests first. Any reason for this? It might be nicer if new tests were put at the end so when we run the tests we can see the output of new tests without needing to scroll up 1000 lines.

PR #30, opened 2016-03-03
Changing the default  behavior of auto-storage

jefferson-dab:
- [x] Code review
- [x] Product review

There are an handful of items that cannot be placed in the storage. Sometimes the user doesn't remember to set the proper flag for them on `items_control.txt` before setting the bot to auto-storage. This usually generates a loop. Openkore tries to store the item, but the item is pushed back on the inventory.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1282,6 +1282,14 @@ sub processAutoStorage {
 					next if $item->{equipped};
 					next if ($item->{broken} && $item->{type} == 7); # dont store pet egg in use
 
+					if (defined($args->{lastInventoryCount}) &&  defined($args->{lastNameID}) &&
+					    $args->{lastNameID} == $item->{nameID} &&
+					    $args->{lastInventoryCount} == @{$char->inventory->getItems()}
+					) {
+						error TF("Unable to store %s.\n", $item->{name});
+						next;
+					}
+
 					my $control = items_control($item->{name});
 
 					debug "AUTOSTORAGE: $item->{name} x $item->{amount} - store = $control->{storage}, keep = $control->{keep}\n", "storage";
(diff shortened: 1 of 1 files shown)


itsrachelfish:
Thanks for the pull request! This change has been long overdue...

I've reviewed the code and everything looks good, but I don't want to merge until I get home and can test in game. If anyone else would like to test, feel free.

allanon:
What happens if the bot is trying to store multiple items with the same ID (eg, a bunch of weapons), and the `ai_storageAuto` timeout is smaller than the server lag? I suspect this will stop early. But maybe autostorage will break in general if the timeout is too small, I haven't tried it. :)

What happens if there are multiple character-bound items with different nameIDs that are in inventory? For example, "Eden Group Armor" and "Eden Group Boots I" and "Culinary Wine" and "Green Apple Ring".

I think the best thing to do would be to do the analysis of what items to store when initiating the autostorage (or perhaps right after opening storage, to avoid inventory renumbering issues), then iterate over that list, attempting to put each one in storage exactly once.

itsrachelfish:
I agree that precalculating a list of items to be stored is a better solution than recalculating what items should be stored constantly.

I wrote my own plugin for putting items in storage quickly (https://github.com/ragnastats/plugins/blob/master/plop/plop.pl#L23-L32) and it literally just loops through all of my items and stores them sequentially. I never have to worry about getting stuck in a loop because it tries once and then it's done. Obviously, this isn't the behavior people want from auto-storage, but the principle is the same.

jefferson-dab:
@allanon 

It works with fairly unreasonable values for `ai_autoStorage` like `0.3`. The results are the same with or without this modification: the bot disconnects if the value is too small. But if the value is safe, then it won't have problems with weapons, for example, because even though the nameID will be the same (in case of multiple weapons), the number of items on the inventory will be smaller.

It doesn't have a problem with multiple items that cannot be stored either. I tested with character-bound equipment, consumables and misc items and it worked as expected.

The issue with analyzing the inventory beforehand is that the bot don't know which items it can and cannot store if it wasn't previously told so on `items_control.txt`. That's the whole problem after all. The bot will only know he cannot store the item once he tries storing it.

itsrachelfish:
@jefferson-dab I think what allanon and I were trying to say is that a more robust solution would be:

1. Initialize auto-storage
2. Get a list of all the items in your inventory
3. Try to insert those items based on your cached list instead of re-checking your inventory

That being said, because your solution checks the number of items in your inventory it seems like it should work fine regardless of if you have multiple items with the same ID. In the end, this change is far simpler than completely rewriting the autostorage behavior as we were describing. :+1:

allanon:
Agreed, this is certainly better than what we had before.

PR #31, opened 2016-03-04
Enable profiles plugin

allanon:
- [x] Code Review
- [x] QA testing

Turn on the profiles plugin by default. This requires at least that we don't prompt if there are no profiles.

As a stretch goal, I want to be able to pass in a profile via a command line option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ errors.txt
 benchmark-results.txt
 logs
 
+# Plugin files
+/profiles
+
 # Backup files
 *~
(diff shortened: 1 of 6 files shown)


itsrachelfish:
This is great! Now we can configure / use our bots normally directly from the git repo without needing to worry about automatic changes to config.txt

ya4ept:
we can do it without a special plug. I do it like this:
start.exe --control=profiles\prompt --logs=profiles\prompt\logs

PR #34, opened 2016-03-04
New hooks in files loading

Henrybk:
- [x] Code review
- [ ] QA testing

This is a request to add 3 hooks to Settings::loadByHandle, the purpose
of this request is to make it possible for plugins to change the filepath
and parsing method of any file just before it's loading. Be aware that
hook names depend on the loading file.

### DIFF
--- a/src/Settings.pm
+++ b/src/Settings.pm
@@ -431,13 +431,15 @@ sub loadByHandle {
 	my $internalFilename = $object->{internalName} || $object->{name};
 
 	#hooks of type 'pre_load_' make it possible to change the filename before openkore searches for it (this filename must contain file name and file path)
-	my %pre_load = (internalFilename => $internalFilename, filename => \$filename);
-	Plugins::callHook('pre_load_'.$internalFilename, \%pre_load);
-	if (!$pre_load{return} && $object->{autoSearch} && $object->{type} == CONTROL_FILE_TYPE) {
+	my $pre_load = {internalFilename => $internalFilename, filename => $filename};
+	Plugins::callHook('pre_load_'.$internalFilename, $pre_load);
+	if ($pre_load->{return}) {
+		 $filename = $pre_load->{filename};
+	} elsif ($object->{autoSearch} && $object->{type} == CONTROL_FILE_TYPE) {
 		$filename = _findFileFromFolders($object->{name}, \@controlFolders);
-	} elsif (!$pre_load{return} && $object->{autoSearch} && $object->{type} == TABLE_FILE_TYPE) {
+	} elsif ($object->{autoSearch} && $object->{type} == TABLE_FILE_TYPE) {
 		$filename = _findFileFromFolders($object->{name}, \@tablesFolders);
-	} elsif (!$pre_load{return} && !$object->{autoSearch}) {
+	} elsif (!$object->{autoSearch}) {
 		$filename = $object->{name};
 	}
 
--- a/src/Settings.pm
+++ b/src/Settings.pm
@@ -428,14 +428,26 @@ sub loadByHandle {
 	assert(defined $object) if DEBUG;
 
 	my $filename;
-	if ($object->{autoSearch}) {
-		if ($object->{type} == CONTROL_FILE_TYPE) {
-			$filename = _findFileFromFolders($object->{name}, \@controlFolders);
-		} else {
-			$filename = _findFileFromFolders($object->{name}, \@tablesFolders);
-		}
+	my $internalFilename;
+	if ($object->{'autoSearch'}) {
+		$internalFilename = $object->{'name'};
 	} else {
-		$filename = $object->{name};
+		$internalFilename = $object->{'internalName'};
+	}
+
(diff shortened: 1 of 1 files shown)


itsrachelfish:
Do you have an example plugin using these hooks we can use for testing?

Henrybk:
Sure, here is my actual use of this Pull request code: http://pastebin.com/84X3rjny

src/Settings.pm
```diff
@@ -428,14 +428,26 @@ sub loadByHandle {
...
+	my $internalFilename;
+	if ($object->{'autoSearch'}) {
+		$internalFilename = $object->{'name'};
+	} else {
+		$internalFilename = $object->{'internalName'};
+	}
+
+	#hooks of type 'pre_load_' make it possible to change the filename before openkore searches for it (this filename must contain file name and file path)
+	my %pre_load = (internalFilename => $internalFilename, filename => \$filename);
+	Plugins::callHook('pre_load_'.$internalFilename, \%pre_load);
```
allanon: I'm on a crusade against deeply nested logic. :)

How about:
```
my $pre_load = {internalFilename => $internalFilename, filename => $filename};
Plugins::callHook('pre_load_'.$internalFilename, $pre_load);
if ($pre_load->{return}) {
	$filename = $pre_load->{filename};
} elsif ($object->{autoSearch} && $object->{type} == CONTROL_FILE_TYPE) {
	$filename = _findFileFromFolders($object->{name}, \@controlFolders);
} elsif ($object->{autoSearch}) {
	$filename = _findFileFromFolders($object->{name}, \@tablesFolders);
} else {
	$filename = $object->{name};
}
```

Also, should we pass the `$object->{type}` as well, so plugins can make the same sort of decisions that this code does?

src/Settings.pm
```diff
@@ -428,14 +428,26 @@ sub loadByHandle {
 	assert(defined $object) if DEBUG;
 
 	my $filename;
-	if ($object->{autoSearch}) {
-		if ($object->{type} == CONTROL_FILE_TYPE) {
-			$filename = _findFileFromFolders($object->{name}, \@controlFolders);
+	my $internalFilename;
+	if ($object->{'autoSearch'}) {
```
allanon: Why does `$internalFilename` depend on `$object->{autoSearch}`? I don't know the context well enough to understand why this isn't just `my $internalFilename = $object->{internalName} || $object->{name};`.

src/Settings.pm
allanon: Oh right, the hook is on a per-file basis so presumably the receiver already knows the object type.

src/Settings.pm
Henrybk: $internalFilename depends on $object->{autoSearch} because if it is '1' then $object->{name} holds the file name, if it is '0' then $object->{name} holds a file path and $object->{internalName} holds the file name. Even so your idea sure is better than just a bunch of ifs.

src/Settings.pm
Henrybk: $filename = $pre_load->{filename}; is not necessary since in my code $pre_load{'filename'} holds a reference to '$filename' so it can be changed inside the plugin itself without any extra source code. About the nested logic, i guess since we are at it there could be no bad at doing some cleaning.

Henrybk: Also it's a good time to add the proper verification for 'TABLE_FILE_TYPE' instead of assuming everything that isn't a 'CONTROL_FILE_TYPE' is a 'TABLE_FILE_TYPE', this adds support for new file types in the future.

src/Settings.pm
```diff
@@ -428,13 +428,16 @@ sub loadByHandle {
...
-		if ($object->{type} == CONTROL_FILE_TYPE) {
-			$filename = _findFileFromFolders($object->{name}, \@controlFolders);
-		} else {
-			$filename = _findFileFromFolders($object->{name}, \@tablesFolders);
-		}
-	} else {
+	my $internalFilename = $object->{internalName} || $object->{name};
+
+	#hooks of type 'pre_load_' make it possible to change the filename before openkore searches for it (this filename must contain file name and file path)
+	my %pre_load = (internalFilename => $internalFilename, filename => \$filename);
```
allanon: I'm not a big fan of passing scalar refs around as in/out variables. Perl doesn't make them easy to work with; when using them, one has to use the clunky double sigil `${$pre_load->{filename}}` to read and write them. While it takes an extra line here to extract the modified `$filename`, it becomes easier for any hook handlers to work with the data.

OpenKore does use this idiom elsewhere though. Whatever you prefer is fine. :)

Henrybk:
Plugin example updated to work without the scalar refs: http://pastebin.com/84X3rjny

Henrybk:
Perl script to convert current portals.txt into JSON format to test with the plugin I gave: http://pastebin.com/Vb3f3Nqa

allanon:
Please don't merge without checking all the boxes. I'm assuming you tested it first, so QA should have been checked?

PR #36, opened 2016-03-05
Commands support for .fld.gz

Henrybk:
Since we are now using .fld.gz files we should change these commands
also.

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -3287,7 +3287,7 @@ sub cmdMove {
 				# map
 				$map_or_portal =~ s/^(\w{3})?(\d@.*)/$2/; # remove instance. is it possible to move to an instance? if not, we could throw an error here
 				# TODO: implement Field::sourceName function here once they are implemented there - 2013.11.26
-				my $file = $map_or_portal.'.fld.gz';
+				my $file = $map_or_portal.'.fld';
 				$file = File::Spec->catfile($Settings::fields_folder, $file) if ($Settings::fields_folder);
 				$file .= ".gz" if (! -f $file); # compressed file
 				if ($maps_lut{"${map_or_portal}.rsw"}) {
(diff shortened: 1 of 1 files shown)


src/Commands.pm
```diff
@@ -3287,7 +3287,7 @@ sub cmdMove {
 				# map
 				$map_or_portal =~ s/^(\w{3})?(\d@.*)/$2/; # remove instance. is it possible to move to an instance? if not, we could throw an error here
 				# TODO: implement Field::sourceName function here once they are implemented there - 2013.11.26
-				my $file = $map_or_portal.'.fld';
+				my $file = $map_or_portal.'.fld.gz';
 				$file = File::Spec->catfile($Settings::fields_folder, $file) if ($Settings::fields_folder);
 				$file .= ".gz" if (! -f $file); # compressed file
```
allanon: Doesn't this line already support `.fld.gz`? We need to keep compatibility with non-gzipped `.fld` files too.

src/Commands.pm
```diff
@@ -4025,9 +4025,9 @@ sub cmdPortalList {
 		#Command: portals add y_airport 143 43 y_airport 148 51 0 c r0 c r0
 		print $args."TEST\n";
 		my ($srcMap, $srcX, $srcY, $dstMap, $dstX, $dstY, $seq) = $args =~ /^add ([a-zA-Z\_\-0-9]*) (\d{1,3}) (\d{1,3}) ([a-zA-Z\_\-0-9]*) (\d{1,3}) (\d{1,3})(.*)$/; #CHECKING
-		my $srcfile = $srcMap.'.fld';
+		my $srcfile = $srcMap.'.fld.gz';
```
allanon: Should write compatibility code so the old `.fld` will still work. See line 3292 above for an example.

src/Commands.pm
Henrybk: I tested it with 'portals add' and it won't work

src/Commands.pm
```diff
@@ -5989,4 +5991,4 @@ sub cmdDeadTime {
 	message $msg, "list";
 }
 
-1;
\ No newline at end of file
+1;
```
allanon: :heart:

allanon:
:+1:

ya4ept:
you can do it in a single commit, not three?
Please, applying this [patch](https://gist.github.com/ya4ept/559965eebd0262a598ad) additionally

PR #41, opened 2016-03-06
New quest kill count db

Henrybk:
This new file will be used to know (or guess at least) the kill count
goal for any known quest. This was created because some servers never
send this info, or at least not when we want it :)

The file will be updated when new info is sent from server.

Needs more testing.


allanon:
This should be a plugin.

Henrybk:
I also thought that but this code needs some deep network managing. For it to work as a plugin some (many?) hooks need to be added.

allanon:
I only see three network hooks in the current code. While I agree that for this to reach its full potential, there will probably be many more, all that logic is a great reason to make it a plugin.

itsrachelfish:
I agree with @allanon, the last thing we need in the core networking code is more features.

No offense intended, your code is good and this feature is something that people would definitely benefit from. That said, ```src/Network/Recieve.pm``` is over 2000 lines long already and ```src/Network/ServerType0.pm``` is almost 7500 lines.

Moving forward I think we should try to keep as much as we can in plugins.

Henrybk:
I'll change it into a plugin and add the necessary hooks into source.

PR #42, opened 2016-03-06
update bRO official files

lututui:
Just some tables files.


itsrachelfish:
I didn't test this in game, but the only files modified were tables files and they all appear to be in the correct format, in Portuguese. :+1:

Issue #21, opened 2016-03-02
Plugins missing


windhamwong:
Some of the plugins are optional for installation however we should think about where can we put them.

itsrachelfish:
What do you mean?

Currently all of the plugins have been moved into the "[needs-review](https://github.com/OpenKore/openkore/tree/master/plugins)" folder except for the macro plugin, which is now enabled by default. When we made the migration to GitHub, we changed the default behavior of the plugins folder to only load a [whitelist of default plugins](https://github.com/OpenKore/openkore/blob/master/control/sys.txt#L40-L52). This lets us have many "official" plugins directly in the repo without needing a subproject.

If there are any plugins you want enabled by default, please review them and move the "trunk" files into their own folder, exactly the same way as the macro plugin is set up.

Let me know if you have any questions, and I'm always on IRC if you wanna chat there too. :)

ya4ept:
I propose to move workers plugins from "../plugins/needs-review/" folder in "../plugins" folder, like macro plugin. For example, I always use the plugin "alert sound". I can not use it by default.

PR #43, opened 2016-03-06
Plugin killcount and necessary hooks

Henrybk:
- [x] Code review
- [x] QA testing

Turned pull request #41 into a plugin. Haven't tested it.


Henrybk:
Tested and seems to be working well. Pic:[https://i.gyazo.com/37661c845d304638c748398fc10d201b.png](url)

itsrachelfish:
Hey, would you mind putting the plugin into its own folder? OpenKore supports loading plugins that are nested 1 folder deep, and we decided that keeping plugins in their own folder now will probably help us keep things more organized down the road.

Some plugins use multiple files or have their own tables / documentation / unit tests, and we can keep all of these things within the plugin folder. For example, instead of placing the quest killcount file inside of tables, you could include it within the plugin folder.

You can read our previous discussion about this:
https://github.com/OpenKore/openkore/pull/3
https://github.com/OpenKore/openkore/issues/21

allanon:
I actually like having the config file in the tables folder. Because:
* `tables` is server-specific, and quests will be server-specific too
* in a "everything is plugins" world, tables will have to have plugins files too, or it'll be empty

itsrachelfish:
As it stands, the killcount file is just in the root of tables, not within a specific server. Should this be changed, then?

Henrybk:
Example of what happens if the user never move/copy the table file:
https://i.gyazo.com/b6a8e3802f603cd26cb8b7fbc361ac94.png

It wll be auto-generated and will add all info sent by the server when needed.

allanon:
@itsrachelfish - I suspect it's fine to have a top-level default killcount file, but each server should be able to override it. This tends to be necessary because official servers can implement their own quests (within limitations) and private servers can do anything they want. And the quest IDs are numbers and can easily overlap.

However, I'm assuming that we want to be able to share, which maybe we don't.

I'd love to eventually get to a world where openkore can do this out of the box:

Log in on a brand new character, type "quest run newbie_castle". Openkore finds a quest called `newbie_castle` (by `label`), and begins execution of it. The `newbie_castle` quest involves performing sub-quests:

1. Walk around.
2. Talk to NPCs.
3. Kill monsters (eg, Poring, Fabre).
4. Gather items (eg, Apple).

Each of these is a potentially complex task. Walking around is mostly solved, but has some wrinkles:
* Standing location is often important. Some NPCs are triggered by walking near them.
* The routing code still gets confused sometimes and doesn't handle another player standing where the bot wants to be very well. This is particularly important for things like newbie castles, where everybody is doing the same thing.
* Sometimes characters are sent to dangerous locations, where they may need to use skills or fight to get to the target location. This isn't a big problem for the newbie castle, but can certainly be an issue for quests intended for higher level characters.
* Sometimes the best way to get there is with a Butterfly Wing, but sometimes walking there is the best choice (eg, when the NPC which gives the quest is on the same map or a nearby field).

Talking to NPCs is relatively straightforward, but answering can sometimes be tricky. Some NPCs want the character to say their name, for instance, which can't be done by talknpc.

Killing monsters is a complex task which involves:
1. Knowing what maps include the monster.
2. Choosing one of those maps. Some interesting information which informs this decision: number of the monster on the map, how close it is, how many monsters need to be killed, how dangerous the map is to this particular bot, available gear, available consumables.
3. Picking up appropriate gear and consumables.
4. Going to the target map (see "walking around").
5. Killing the target monsters. Potentially avoiding all other monsters, if those monsters might kill the bot.

Gathering items is a shockingly difficult task. In the end, it's easy to say "get an Apple from storage", but making that decision is very hard, and even performing it can be hard.
* Find the item in inventory.
* Find the item in the cart.
* Find the item in storage.
* Buy the item from an NPC.
* Buy the item from a player vend.
* Find the item in another bot's inventory and trade for it (in bot network scenarios). This may be common for repeatable quests.
* Kill monsters to get the item (see "killing monsters").

And there are also some meta concerns like:
* Taking multiple quests and attempting to do all of them at once.

In the end, this plugin provides one piece of one step of one of these behaviors: knowing how many monsters need to be killed to complete a killcount quest. I'd love to see a set of plugins which work together to solve the overall problem, one piece at a time. :)

Henrybk:
Could someone review the code and test it? I got it to work without any problems.

itsrachelfish:
I was talking with @Henrybk on IRC and it looks like bRO is one of the few servers that still uses the old quest packets. Because of that, I think someone from bRO should test this plugin. 

Ping: @lututui @flacrasec @orlandomf :smile:

Henrybk:
Ping 2: @AlexandreBR3N @emsjustin

flacrasec:
When I tested #43 i have got this error:
error

lututui:
@flacrasec install the JSON module via ppm or cpan (or manual)

flacrasec:
it will be a bit annoyng if everyone needs to do these steps you are seeing for yourself lututui! =P

itsrachelfish:
I agree, if we're going to include this as an officially supported plugin the JSON module should be added into kore's ```src/deps/``` folder

Henrybk:
@itsrachelfish @flacrasec Now it should be working

flacrasec:
Good afternoon Henry, it works only with openkore.pl, start.exe and wxstart.exe can't be used with the plugin killcountFix.pl. is the result expecting this of the picture?
image

Henrybk:
The point is that this plugin adds a quest hunting database to openkore, in situations where kore wouldn't normally be able to know how many and which mobs to kill for a certain quest this plugin would serve it's purpose.

Example of situations when kore won't know much about the hunting mission:
1- when it first gets the quest
2- when map is reloaded (map exchange or teleport)
3- when it relogs
4- anytime before killing the first mob

To test this plugin you first need to copy the databse file ( the .txt) to the tables folder, after that get a hunting mission on eden or any other, now still in front of the npc type 'quest list' in the console, you should see the quest and the mob ID/Name you need to kill, but not the necessary kill count goal. Now activate the plugin in sys.txt and repeat the process with the plugin activated, you should in the same situation see not only the quest ID and Mob ID/Name but also the kill count goal for that mob.

You can also try that after killing a few mobs for the quest, log into your character with the plugin not active and type 'quest list' you should see only the kill count, but when you repeat with the plugin you should see the kill count goal also. 

I hope I helped.

flacrasec:
I have put the database archives at tables/bro(right?) and the same was told me that the new quest was added but it doesn't show the mob kills only if I use quest list and quest list was already showing the total mob that I killed!

Henrybk:
So, can anyone confirm it is working?

flacrasec:
the last updates was tested and functional congratulations Henry I approve this.

allanon:
Let's merge!

PR #47, opened 2016-03-09
Split resnametable

allanon:
- [x] QA Review

Split up `resnametable.txt`, which OpenKore uses for map aliasing, and alias izlude (and friends) to `izlude_a.fld`, which is the new Izlude Academy version of izlude.


itsrachelfish:
Works great, tested in Izlude on iRO Chaos.

PR #54, opened 2016-03-15
Plugins sort

Henrybk:
Change to getFilesFromDirs so it's guaranteed to return a sorted plugins
list


Henrybk:
Wow, i didn't see i had all those merges, i'm gonna fix that later

itsrachelfish:
Why is it necessary that plugins are sorted alphabetically?

Henrybk:
Because we may need in the future to load plugins in a specific order. I stumbled upon this need when creating an extension to the profiles plugin

allanon:
Can you give a more concrete, near-term example? I don't want to commit code we don't need.

That said, I can see a completely separate argument for loading plugins in alphabetical order because the console output looks nicer.

itsrachelfish:
Relying on the order plugins are being loaded sets a bad precedence. Relying on the order based on the plugin name sounds even worse. Do we really want to name plugins "00_profile" "01_macro"?

OpenKore is an event-based application and any potential inter-plugin dependencies should be handled by using hooks or passing references to shared variables.

I don't think it's wise to introduce code like this without specific examples of the problems you were facing so that we can try to come up with other solutions that don't rely on the order plugins are being loaded.

PR #57, opened 2016-03-21
macro improvements

allanon:
DO NOT MERGE - This is for review only, at this point.

### DIFF
--- a/plugins/macro/Macro/Automacro.pm
+++ b/plugins/macro/Macro/Automacro.pm
@@ -5,10 +5,10 @@ use strict;
 
 require Exporter;
 our @ISA = qw(Exporter);
-our @EXPORT_OK = qw(releaseAM lockAM automacroCheck consoleCheckWrapper);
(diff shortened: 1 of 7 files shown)


itsrachelfish:
Can you provide some insight into what improvements were made and why it shouldn't be merged?

allanon:
This is a raw dump of the code of the version of the macro plugin which I currently use. However:

*  as you noted, the additional features are not documented
* the additional features are not split out into individual feature pull requests
* there is some debugging code left in here
* this breaks backwards compatibility with previous macro features (the one I can think of off the top of my head is what happens when one passes arguments to a macro on the command line; eg, `macro foo arg1 arg2`)

This pull request only exists to allow others to look at the code and think about the feature changes I made. In case they see something they really want, a separate pull request for that specific feature should be made.

I don't have time to properly split this up into separate pull requests right this second. I wanted to expose the code anyway, specifically so Henrybk could look at it, but also so anybody else could look at it.

kaliwanagan:
Can we close this PR?

What we should do instead is to create an issue referencing the branch you're interested in, so we can have the discussion in there

PR #66, opened 2016-04-06
Chat Commands Auth

lututui:
- [x] Code review
- [x] QA review

There should be a way to prevent in-game auth to Chat Commands by sending a message with adminPassword, in order to prevent people from abusing an easy adminPassword

http://openkorebrasil.org/index.php?/topic/499-aonde-posso-pegar-atualiza%C3%A7%C3%A3o-segura/
http://openkorebrasil.org/index.php?/topic/490-logout/

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -40,6 +40,7 @@ adminPassword
 callSign
 commandPrefix ;
 callSignGM 0
+inGameAuth 0
 
 pauseCharServer 0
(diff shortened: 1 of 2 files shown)


Henrybk:
I agree with the idea, especially the part it is set to 0 by default

itsrachelfish:
Tested in game, works great! :+1:

Issue #37, opened 2016-03-05
Managing inventory lists


Henrybk:
The way openkore manages inventory lists is very (awful) strange. Character Inventory is a class, storage inventory are two arrays and cart is a hash. I suggest we turn inventoryList into a abstract class with at least 3 subclasses: inventory, storage and cart, and maybe other ones later like vender inventory, deal inventory, buy_store_inventory, etc.

I made something like that last year, it works (at least for me), and it fixes many bugs.
Patch: http://pastebin.com/URRkTnLf

Bugs I remember it fixes:
Wrong item count checking when openkore teleports/changes map (all is 0)
ai_auto_storage not activating when character inventory is empty
character starting auto_storage because of wrong item checking

Note: This patch was apliable for openkore before git migration, it may need some changes now, it was just as axample and needs lots of rewritting and testing, I also never added support for xkore for it.

Henrybk:
Fixed in #119

PR #67, opened 2016-04-06
New inventorylist class

Henrybk:
This is not to be merged as it will probably cause errors with some plugins and some code is still to be tested.

This is a solution I made for issue 37 (https://github.com/OpenKore/openkore/issues/37). With this patch all character, cart and storage inventories will be handled in the same exact way, leaving the code way cleaner.

It fixes some bugs as empty inventories not being recognized, map change inventory and cart bugs and unwanted autostorage due to bad  inventory tracking. All the code needs to be reviewed and some parts rewritten.

I also have not made any work on Wx interface.

Nonetheless the code works with a clean openkore with no plugins installed (at least it did a few versions back), that's why I changed sys.txt so I won't forget about it.

This pull request was mostly created due to me wanting to see your opinion about this idea.


src/AI/CoreLogic.pm
```diff
@@ -993,14 +993,14 @@ sub processStorageGet {
 sub processCartAdd {
 	if (AI::action eq "cartAdd" && timeOut(AI::args)) {
 		my $item = AI::args->{items}[0];
-		my $i = $item->{index};
-		my $invItem = $char->inventory->get($i);
+		my $index = $item->{index};
+		my $amount = $item->{amount};
+		my $invItem = $char->inventory->getByServerIndex($index);
```
allanon: This might be more succinct:
```
my $invItem = $char->inventory->getByServerIndex($item->{index});
if ($invItem) {
    $messageSender->sendCartAdd($item->{index}, min($invItem->{amount}, $item->{amount} || $invItem->{amount}));
}
```

src/AI/CoreLogic.pm
```diff
@@ -1777,48 +1779,44 @@ sub processAutoCart {
...
+				} elsif ($invItem->{amount} < $control->{keep}) {
+					$amount = $control->{keep} - $invItem->{amount};
 				}
 				if ($amount > $cartItem->{amount}) {
 					$amount = $cartItem->{amount};
 				}
 				if ($amount > 0) {
 					my %obj;
-					$obj{index} = $i;
+					$obj{index} = $cartItem->{index};
```
allanon: I haven't looked over this whole pull request yet, but it seems like this is a breaking change. `AI::cartGet` expects to get the client index, not the server index. And while you've changed the implementation of `processCartGet` to match, so it doesn't actually break this particular use of `cartGet`, any other code which uses `cartGet` will break. Is there a benefit to using the server index? Does this protect us when we get disconnected in the middle of the sequence or something, for instance? (Since I know that will re-order the client indices.)

src/InventoryList/Cart.pm
```diff
@@ -0,0 +1,56 @@
...
+	return $self->{exists};
+}
+
+sub info {
+	my ($self, $args) = @_;
+	$self->{items} = $args->{items};
+	$self->{items_max} = $args->{items_max};
+	$self->{weight} = int($args->{weight} / 10);
+	$self->{weight_max} = int($args->{weight_max} / 10);
+	$self->{exists} = 1 if (!$self->{exists});
```
allanon: `$self->{exists} ||= 1;`? Or just `$self->{exists} = 1;`, since it doesn't actually matter what it was set to before.

src/InventoryList/Inventory.pm
```diff
@@ -0,0 +1,49 @@
...
+
+use constant {
+	MAP_LOADED => 0,
+	RECV_STAT_INFO2 => 1
+};
+
+sub new {
+	my ($class) = @_;
+	my $self = $class->SUPER::new('Inventory');
+	$self->{hooks} = Plugins::addHooks (
```
allanon: Do these hooks ever get removed? It's not safe to assume `DESTROY` will remove them either, because the garbage collector may not be run immediately when the object is freed. And because the hook is a closure, this object is a memory leak. Does it make sense to assume that only one inventory can ever exist? If so, the hook could be a global and this object could hide the global state. If we don't want to do that (and I'd love to see a day where a single instance of OpenKore could manage multiple characters), then we need logic to remove these hooks when we allocate a new `$char` object.

src/Actor.pm
```diff
@@ -628,7 +628,7 @@ sub statusActive {
 sub cartActive {
 	my ($self) = @_;
 	
-	if ($cart{exists} ||
+	if ($self->cart->{exists} ||
```
allanon: `$self->cart->isReady`

src/Interface/Console/Curses.pm
```diff
@@ -592,7 +592,7 @@ sub updateStatus {
 		hp       => { title => 'HP',     cur => $char->{hp},        max => $char->{hp_max},     color1 => 'bold|red',  color2 => 'bold|green', threshold => 15 },
 		sp       => { title => 'SP',     cur => $char->{sp},        max => $char->{sp_max},     color1 => 'bold|blue', color2 => '',           threshold => 0 },
 		weight   => { title => 'Weight', cur => $char->{weight},    max => $char->{weight_max}, color1 => 'cyan',      color2 => 'red',        threshold => 50 },
-		cart     => { title => 'Cart W', cur => $cart{weight},      max => $cart{weight_max},   color1 => 'cyan',      color2 => '',           threshold => 0 },
+		cart     => { title => 'Cart W', cur => $char->cart->{weight},      max => $char->cart->{weight_max},   color1 => 'cyan',      color2 => '',           threshold => 0 },
```
allanon: Should we make accessors for `weight` and `weight_max`? Might be nice to be able to write `$char->cart->weight`.

src/InventoryList.pm
```diff
@@ -76,6 +76,7 @@ sub new {
 
 sub DESTROY {
 	my ($self) = @_;
+	Plugins::delHook($self->{hook}) if $self->{hook};
```
allanon: Does anything ever set `$self->{hook}`? `InventoryList::Inventory->new` sets `$self->{hooks}` instead.

src/InventoryList/Cart.pm
```diff
@@ -0,0 +1,56 @@
...
+package InventoryList::Cart;
+
+use strict;
+use Globals;
+use InventoryList;
+use base qw(InventoryList);
+
+sub new {
+	my ($class) = @_;
+	my $self = $class->SUPER::new('Cart');
```
allanon: I don't think passing a parameter to `InventoryList->new` does anything. Might as well remove it? `my $self = $class->SUPER::new;`

PR #83, opened 2016-05-03
add attackAuto -1

allanon:
- [x] Product Review (is this something we should do?)
- [x] QA Review (does it work?)

Add option to turn off the `attackAuto` logic entirely.

Without this, the bot will spend time thinking about attacking nearby monsters even if it ultimately chooses not to attack. And more importantly, if a monster attacks the bot, the bot will stop and try to attack back. This is often a bad idea, particularly for priests trying to run past aggressive monsters.

With `attackAuto -1`, priests (and other support characters) can easily run past aggressive monsters without stopping.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2666,6 +2666,9 @@ sub processAutoEquip {
 
 ##### AUTO-ATTACK #####
 sub processAutoAttack {
+	# Don't even think about attacking if attackAuto is -1.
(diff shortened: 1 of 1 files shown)


itsrachelfish:
I like this idea, it's simple and definitely an issue I've run into before. I'd be interested in running kore in debug mode to see what the benchmarks are for processing auto attack.

itsrachelfish:
Tested in game, seems to work fine :+1:

PR #84, opened 2016-05-03
fix updating config.txt with !include

allanon:
- [x] Code Review
- [x] QA Review

Making a config change inside kore auto-saves the config. Since the recent change to `Utils::TextReader`, that re-save would cause all included files to be inlined into the config, instead of being preserved as includes.

This pull request fixes that.

### DIFF
--- a/src/FileParsers.pm
+++ b/src/FileParsers.pm
@@ -716,7 +716,7 @@ sub parseRODescLUT {
 	undef %{$r_hash};
 	my $ID;
 	my $IDdesc;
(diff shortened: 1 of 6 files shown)


itsrachelfish:
Awesome job, you changed the behavior and updated the tests. :wink: 

The code looks great, but I'll have to give it a try in game first.

itsrachelfish:
Updated my config file to use ```!include account.txt``` and compared between master and this branch.

On master, my include block was replaced with the original text, on your branch it preserves the original include line. :+1:

PR #105, opened 2016-06-05
let the server know that we're done buying/selling so we can move

allanon:
- [x] QA Review

iRO now does the same thing fRO does: after talking to an NPC vendor, we must send a 09D4 packet before we can move.

This pull request solves the problem in a different way from the fRO solution, so that it will (hopefully) work in more use cases. Here we keep track of whether we started talking to an NPC vendor and send the packet along with the next move packet, regardless of whether we tried to buy or sell something.

### DIFF
--- a/src/Network/Receive/iRO.pm
+++ b/src/Network/Receive/iRO.pm
@@ -63,6 +63,15 @@ sub received_characters_info {
 	$self->received_characters($args);
 }
 
+sub npc_store_begin {
+	my $self = shift;
+
+	# The server won't let us move until we send the sell complete packet.
+	$messageSender->{sell_mode} = 1;
+
+	$self->SUPER::npc_store_begin(@_);
+}
+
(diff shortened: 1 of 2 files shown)


allanon:
Tested by ZiriusPH on IRC. (Thanks ZiriusPH!)

Mathspy:
I would like to say that although the issue was fixed. While using XKore2 the server seems to disconnect after finalizing a deal although Openkore keeps working as intended so it's a rather weird inconvenience having to relog in (I know it takes less than 3 seconds)

I don't mind it, but thought I should probably report it in

marcelothebuilder:
@Mathspy Is it happening when manually controlling your character (and clicking on menus, etc.) or when viewing an autobuy/sell sequence?

Mathspy:
@marcelothebuilder Manually controlling

PR #114, opened 2016-06-11
add --command command line option

allanon:
- [x] Code review
- [x] QA testing

Add `--command=...` command line option. Useful for completely automated processes like:

```
openkore.pl --command="macro do_something_and_quit"
```

### DIFF
--- a/src/Settings.pm
+++ b/src/Settings.pm
@@ -115,6 +115,7 @@ our $dead_log_file;
 our $interface;
 our $lockdown;
 our $starting_ai;
+our $command;
 our $no_connect;
 
 
@@ -181,6 +182,7 @@ sub parseArguments {
 		'interface=s',        \$interface,
 		'lockdown',           \$lockdown,
 		'ai=s',               \$starting_ai,
+		'command=s',          \$command,
 		'help',	              \$options{help},
 		'version|v',          \$options{version},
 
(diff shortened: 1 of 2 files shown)


Henrybk:
Shouldn't this be done with CmdOnLogin plugin? Maybe we could rework it to use commands on more situations during login, like on character screen, on account login, on files load, etc.

allanon:
Hm. For reference, you're talking about: https://github.com/OpenKore/openkore/blob/master/plugins/needs-review/cmdOnLogin/trunk/cmdOnLogin.pl

That plugin requires a config option, and I want to be able to do it via command line. But now that plugins can inject command line options, it should be possible to enhance the plugin.

itsrachelfish:
@allanon Can you give an example of what you're using this for? I'm curious what macro you need to run and then quit the game...

allanon:
Two of them:
1. gramps_and_quit
2. daily_quen_and_quit

The first gets the current Gramps TI quest for the character's level range. It does not attempt to do the quest. This is useful because characters can go into TI as long as they have the current quest, even if they're out of the level range. So I can collect all the quests and go in when I feel like it.

The second runs the Daily Quen quest, which can net some free Enriched Elunium/Oridecon. Useful stuff.

itsrachelfish:
Tested and working. :)

Are there any drawbacks to having this functionality in a plugin like @Henrybk was talking about?

PR #119, opened 2016-06-15
new InventoryList class system

Henrybk:
- [x] Code review
- [x] QA testing

This is a big code review and rework that focus on getting cart, storage and character inventory to be managed in the same way, it seems to work well. I added support for macro plugin but haven't tested it yet.

I have done no work with the Wx interface, so if someone likes this idea and wants to help, feel free.


Henrybk:
In the last commit I fixed a bug with the line ending character, now it should be way better to check the code changes.

It would be of great help it this could be tested, I'll try later to comment on all fixed bugs.

src/Actor/You.pm
```diff
@@ -117,6 +122,24 @@ sub inventory {
 }
 
 ##
+# InventoryList $char->storage()
+# Ensures: defined(result)
+#
+# Get the storage list for this character.
+sub storage {
+	return $_[0]->{__storage};
+}
```
farrainbow: Do we really need an indirection there or it's just for writing `$char->storage->...` instead of `$char->{storage}->...`?

src/InventoryList/Cart.pm
```diff
@@ -0,0 +1,80 @@
...
+
+sub changeType {
+	my ($self, $args) = @_;
+	$self->{type} = $args;
+}
+
+sub weight {
+	my ($self) = @_;
+	return $self->{weight};
+}
```
farrainbow: Comment about misdirection applies there as well and in the similar places. Maybe keep it simple?

src/InventoryList/Storage.pm
```diff
@@ -0,0 +1,59 @@
...
+	$self->{items_max} = 0;
+	return $self;
+}
+
+sub isOpenedThisSession {
+	my ($self) = @_;
+	return $self->{openedThisSession} == 1;
+}
+
+sub isOpened {
```
farrainbow: Maybe name it `isReady`, as in Inventory and Cart if they all have the same purpose?

src/InventoryList/Cart.pm
```diff
@@ -0,0 +1,80 @@
...
+	$self->{exists} = 1;
+}
+
+sub onMapChange {
+	my ($self) = @_;
+	$self->{exists} = 0;
+	$self->clear();
+}
+
+sub release {
```
farrainbow: Maybe name it `close` as in Storage if it has the same purpose?

src/Actor/You.pm
Henrybk: It was just done to be able to use `$char->storage->`, I don't think directly accessing these vriables is a good practice.

src/InventoryList/Cart.pm
Henrybk: I will review this later today, you are probably right but I am not home right now to be able to properly consider it.

src/InventoryList/Cart.pm
Henrybk: This one I agree is a little too much misdirection.

src/InventoryList/Storage.pm
Henrybk: Same as above, I'll review this later today.

Henrybk:
@farrainbow These are the changes I would like the most to see implemented, thanks for helping me and reviewing my code :)

Henrybk:
Wow I did some crazy wrong stuff but it's alright now.

Henrybk:
I just remembered the problem with this pr, i never updated the wx interface code, i will have to see about that

farrainbow:
I can do Wx stuff, maybe push this as a branch to main repository?

Henrybk:
@farrainbow I have, in the last commit, implemented most of your suggestions.

farrainbow:
`storage->wasOpenedThisSession` is a nice one.

farrainbow:
Merge those: https://github.com/OpenKore/openkore/tree/new_inventory_class_system

Henrybk:
@farrainbow Done

Henrybk:
I believe this is now complete

Henrybk:
Let the bugs come

PR #122, opened 2016-06-18
fix for autostorage bug

Henrybk:
When kore puts an item into storage but keeps some of it in inventory we
receive the error message "Unable to store", but the item was stored,
this is due to an error in the error message verification. this change
should fix it.

Image of error:
https://i.gyazo.com/f694972338c497f6b7dc9cef21b83432.png

After change: 
https://i.gyazo.com/2dd1c91d968cf410139edc35a3612313.png

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1282,9 +1282,10 @@ sub processAutoStorage {
 					next if $item->{equipped};
 					next if ($item->{broken} && $item->{type} == 7); # dont store pet egg in use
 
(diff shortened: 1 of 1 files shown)


Henrybk:
Could someone try to reproduce this bug?

lututui:
It works! Tested on RMS Renewal Test Server

items_control.txt -> http://i.imgur.com/v1fIeV8.png

Before -> http://i.imgur.com/stv23IO.png
After -> http://i.imgur.com/RxFvZ2E.png

PR #123, opened 2016-06-19
Remove tasks raiseStat and raiseSkill

Henrybk:
raiseStat and raiseSkill are both unstoppable Tasks which are started during initialization even if the user has disabled them in config.txt.

This pull request focus on completely removing RaiseSkill and RaiseStat from source code, and changing it into a plugin.

I have only made RaiseStat so far, but I would like to have your opinions.


allanon:
What do you think about speculatively turning on the `AI_pre` handler whenever something happens that _might_ mean we can raise stats, then turn it off if we can't actually do so? That might reduce the amount of logic we need.

I've implemented something like that as a Simple plugin here: https://github.com/allanon/openkore-addons/blob/master/plugins/raisestat.spl

Henrybk:
I made the plugin with the idea of using the least cpu power possible without being silly, the hooks adding/removing and the complicated logic is a consequence of that. Nonetheless I think I should have added some more speculative hooks, thanks.

plugins/raiseStat/raiseStat.pl
```diff
@@ -0,0 +1,185 @@
...
+#  look for opportunities to raise stats.
+#  During status == 1 the plugin will be active and will
+#  have 'speculative' hooks added to try to look for
+#  opportunities to raise stats.
+#  During status == 2 the plugin will be active and will
+#  have 'AI_pre' hook active, on each AI call the plugin
+#  will try to raise stats if possible.
+sub changeStatus {
+	my $new_status = shift;
+	Plugins::delHook($active_hooks) if ($status == ACTIVE);
```
allanon: Why is this conditional separate from the one with the same condition below?

plugins/raiseStat/raiseStat.pl
```diff
@@ -0,0 +1,185 @@
...
+	}
+	message "raiseStat has no more stats to raise; disabling statsAddAuto\n", 'success';
+	return 0;
+}
+
+################################################################
+#  canRaise() is the function responsible for checking if
+#  we have enough free points to raise '$next_stat'.
+sub canRaise {
+	return 1 if ($status != INACTIVE && $char->{points_free} && $char->{"points_".$next_stat} && $char->{points_free} >= $char->{"points_".$next_stat});
```
allanon: Unless you need the values to be exactly `0` and `1`, you can just use perl's conditional result. `return` is unnecessary (but optional if you prefer it). For example:

```
sub canRaise {
	$status != INACTIVE and
	$char->{points_free} and
	$char->{"points_".$next_stat} and
	$char->{points_free} >= $char->{"points_".$next_stat};
}
```

plugins/raiseStat/raiseStat.pl
Henrybk: 'changeStatus' makes sure '$adding_hook' will only be active during 'ADDING' and '$active_hooks' will only be active during 'ACTIVE', so deleting them this way is okay.

And the conditions are different, first one is 'if ($status == ACTIVE)' and second one is 'if ($status == ADDING)'.

plugins/raiseStat/raiseStat.pl
Henrybk: I actually never used this Perl feature so I didn't even know it existed haha.

plugins/raiseStat/raiseStat.pl
allanon: Ah, I see it now. Removal is predicated on `$status`, while adding is predicated on `$new_status`.

allanon:
Looks good to me, code-wise. But it pains me to see the localized messages get deleted.

Henrybk:
What do you mean by localized messages? The translations?

allanon:
Yeah, I meant that it pains me to see the translations get deleted.

Henrybk:
raiseStat is complete and commented, raiseSkill must still be tested, and we must find a way to deal with impossible skills add lists (because people set them wrongly), since kore will still keep trying to add them even after receiving errors.

plugins/raiseSkill/raiseSkill.pl
```diff
@@ -0,0 +1,172 @@
...
+	$messageSender->sendAddSkillPoint($next_skill->getIDN);
+}
+
+################################################################
+#  on_configModify() is called whenever config is changed, it
+#  checks if something important to us was changed, and can 
+#  change plugin status.
+sub on_configModify {
+	my (undef, $args) = @_;
+	return changeStatus(ACTIVE) if ($args->{key} eq 'skillsAddAuto' && $args->{val} && $config{skillsAddAuto_list} && validateSteps($config{skillsAddAuto_list}));
```
allanon: I think we should go straight to `ADDING` instead of `ACTIVE`. Let the `AI_pre` hook downgrade back to `ACTIVE` if we can't actually add skills.

plugins/raiseSkill/raiseSkill.pl
```diff
@@ -0,0 +1,172 @@
...
+sub changeStatus {
+	my $new_status = shift;
+	Plugins::delHook($active_hooks) if ($status == ACTIVE);
+	Plugins::delHook($adding_hook) if ($status == ADDING);
+	if ($new_status == INACTIVE) {
+		undef $next_skill;
+		undef @skills_to_add;
+	} elsif ($new_status == ACTIVE) {
+		$active_hooks = Plugins::addHooks(
+			['map_loaded',           \&on_possible_raise_chance],
```
allanon: Should probably just listen for `stat_info` with `type=0x0c` (`VAR_SPPOINT`).

plugins/raiseStat/raiseStat.pl
```diff
@@ -0,0 +1,186 @@
...
+sub changeStatus {
+	my $new_status = shift;
+	Plugins::delHook($active_hooks) if ($status == ACTIVE);
+	Plugins::delHook($adding_hook) if ($status == ADDING);
+	if ($new_status == INACTIVE) {
+		undef $next_stat;
+		undef @stats_to_add;
+	} elsif ($new_status == ACTIVE) {
+		$active_hooks = Plugins::addHooks(
+			['map_loaded',           \&on_possible_raise_chance],
```
allanon: Should probably just listen for `stat_info` with `type=0x09` (`VAR_POINT`).

Henrybk:
on v7 I used the same hooks as the original tasks, so it should be just as safe as they were.
Now I only have to find a way to disable skillsAddAuto if the list is incorrect.

Henrybk:
These plugins should now work as intended.

It would be of great help if anyone tested it.

kaliwanagan:
@Henrybk just need the merge conflicts resolved and we're good to go

Henrybk:
@kaliwanagan Resolved

PR #136, opened 2016-07-10
Character zeny change moved to before hook call

Henrybk:
Hook calls and $char changes should be consistent across servertypes, in
ServerType0 $char is changed before the hook call, in Sakexe_0 it should
also be done this way.


itsrachelfish:
:+1: for hook consistency

PR #137, opened 2016-07-11
add wildcard search to conf command, eg "conf sit*"

allanon:
- [x] Code review
- [x] QA testing

Add wildcard searching to conf command, to make it easier to find the config option the user is trying to change.

For example, to find all "sit" related options, do: `conf sit*`

Some patterns produce a lot of results. For example, `conf skill*`. The conf options are sorted to hopefully make these easier to browse.

Multiple wildcards are allowed. For example, `conf tele*lock` finds all options which include "tele" followed by "lock". Which for me finds three options: teleportAuto_maxDmgInLock, teleportAuto_minAggressivesInLock, and teleportAuto_totalDmgInLock.

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -1321,8 +1321,7 @@ sub cmdCloseBuyShop {
 
 sub cmdConf {
 	my (undef, $args) = @_;
-	my ($arg1) = $args =~ /^(\w*\.*\w+)/;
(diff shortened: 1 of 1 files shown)


itsrachelfish:
This is an awesome feature! Glad to have it in OpenKore after all these years :+1:

Issue #145, opened 2016-07-19
autoBreakTime.pl is not working


kittygrill:
Breaktime is not working.

itsrachelfish:
@kittygrill The breakTime plugin is now enabled by default in the newest version of kore. Please reply here if you're still having any issues. Thanks! :+1:

PR #154, opened 2016-07-25
Enable the breakTime plugin

allanon:
- [x] Code review
- [x] QA Review

* this might fix https://github.com/OpenKore/openkore/issues/145 
* move plugin to `plugins`
* enable plugin in `sys.txt`
* add documentation to plugin's `README.md` (should this be somehow auto-sync'd with the wiki?)
* add unit tests
* fix a couple of bugs:
  * break times spanning midnight would not trigger if the current time was after midnight but before `stopTime`
  * the plugin would reset the login time if the user had deliberately logged out for a long time (longer than `stopTime`) - it should only extend the time spent logged out, not shorten it


itsrachelfish:
I tested this branch and configured my bot to use the autoBreakTime options. Everything seems to be working, except it looks like the plugin description is messed up:

breaktime-description-bug

PR #173, opened 2016-08-04
support packet 0A3B (extended misc_effect)

allanon:
- [ ] QA Review

I'm not actually sure how to cause packet 0A3B to trigger, but it does happen in prontera iRO Chaos periodically.

Thanks to snop222 for making a twRO implementation of it, which is promoted to a global handler by this pull request.


allanon:
https://github.com/OpenKore/openkore/pull/175 is better.

PR #175, opened 2016-08-05
Added support for Hat Effect Info

### LINKED ISSUE #164: Packet 0xA3B is for HatEffectInfo
Well, I just gonna put the message here first.
I forgot where someone mentioned this packet, and there was answer if it was `misc_effect`, but I found that it's for hateffectinfo (if in client side it's in luafiles514/lua files/hateffectinfo/hateffectinfo.lub).
I can do a little work around for this by follow the changes in https://github.com/rathena/rathena/commit/5da49b87b5a2ae77852d1f697b62f5bfea7bcde8#diff-85d990dde8caf2d72d1e2b0f7dcf9786.
The packet is support for multiple `HatEFID` that be sent at once.

cydh:
* Received packet: `0xA3B`
  * Structure: `['hat_effect', 'v a4 C a*', [qw(len ID flag effect)]], # -1`
* Added table files:
  * `hateffect_id_handle.txt`
  * `hateffect_name.txt`
  * Files extracted from kRO: `data/luafiles514/lua files/hateffectinfo/hateffectinfo.lua`
* Credits for detailed packet info https://github.com/rathena/rathena/commit/5da49b87b5a2ae77852d1f697b62f5bfea7bcde8

As well for:
* Fixed #164
* So for #173
  * Just need to add the entry for iRO's `recvpackets.txt`
  * Don't mixed `misc_effect` with `0xA3B`

Tested on @rAthena by using `@send` command. (If using actual client, there are many things to do, since OpenKore may incompatible for kRO 2015-05-13)
screenshot_49

Signed-off-by: Cydh Ramdh <cydh@pservero.com>


allanon:
Nice. Looks like you figured out what was going on better than I did. Should we remove the `Network::Receive::twRO` implementation of this packet? (I think we should.)

cydh:
Yeah, I think should remove twRO implementation.

PR #180, opened 2016-08-09
Update CoreLogic.pm

systeman:
processPartySkillUse do have Alternating between the comparison(||) with the calculation(or), Openkore cann't use party skill on smooth.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2504,14 +2504,14 @@ sub processPartySkillUse {
 				if (
 					( # range check
 						$party_skill{owner}{ID} eq $player->{ID}
-						|| inRange(distance($party_skill{owner}{pos_to}, $player->{pos}), $config{"partySkill_$i"."_dist"} || $config{partySkillDistance} || "0..8")
+						|| inRange(distance($party_skill{owner}{pos_to}, $player->{pos}), $config{"partySkill_$i"."_dist"} or $config{partySkillDistance} or "0..8")
 					)
 					&& ( # target check
 						!$config{"partySkill_$i"."_target"}
-						or existsInList($config{"partySkill_$i"."_target"}, $player->{name})
-						or $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
-						or $char->{homunculus} && $player->{ID} eq $char->{homunculus}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@homunculus')
-						or $char->{mercenary} && $player->{ID} eq $char->{mercenary}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@mercenary')
+						|| existsInList($config{"partySkill_$i"."_target"}, $player->{name})
+						|| $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
+						|| $char->{homunculus} && $player->{ID} eq $char->{homunculus}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@homunculus')
+						|| $char->{mercenary} && $player->{ID} eq $char->{mercenary}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@mercenary')
 					)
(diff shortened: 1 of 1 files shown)


src/AI/CoreLogic.pm
```diff
@@ -2504,14 +2504,14 @@ sub processPartySkillUse {
...
+						|| inRange(distance($party_skill{owner}{pos_to}, $player->{pos}), $config{"partySkill_$i"."_dist"} or $config{partySkillDistance} or "0..8")
 					)
 					&& ( # target check
 						!$config{"partySkill_$i"."_target"}
-						or existsInList($config{"partySkill_$i"."_target"}, $player->{name})
-						or $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
-						or $char->{homunculus} && $player->{ID} eq $char->{homunculus}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@homunculus')
-						or $char->{mercenary} && $player->{ID} eq $char->{mercenary}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@mercenary')
+						|| existsInList($config{"partySkill_$i"."_target"}, $player->{name})
+						|| $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
```
allanon: This is wrong, I believe. The operator precedence of `or` is very low, so `if (A or B or C && D or ...)` is equivalent to `if (A || B || (C && D) || ...)`.

The logic was "if there is no target, or the target includes this player, or the target is me, or the target is me, or the target is my homunculus, or the target is my mercenary" then do stuff.

This changes the logic to "if (((((there is no target or the target includes this player or the player is me) AND @main is in the target list) OR i have a homunculus) AND the player is my homunculus and @homunculus is in the list) OR i have a mercenary) AND the player is my mercenary and @mercenary is in the list", which is satisfied only for mercenaries.

I'd totally believe that there's a performance problem here, since we're doing a scan of every player, homunculus, and mercenary nearby for every configured partySkill. Fixing that performance problem is a good idea. But I think the logic is correct, and this change breaks it.

systeman:
that's not really, (or, and) good for do calculate or do object, but (||, &&) good for do comparison. such as this line.
|| inRange(distance($party_skill{owner}{pos_to}, $player->{pos}), $config{"partySkill_$i"."_dist"} or $config{partySkillDistance} or "0..8")

that's you want  value of '$player->{pos}), $config{"partySkill_$i"."_dist"} or $config{partySkillDistance} or "0..8' 

'||' don't work this above line, becuase you do want any value of these you don't want comparison its.

if You do want logic calculate,  '$Y = $a and ($b or $c)' good enough.

but you do want comparison (||, &&) good performance for 'if (a && (b || c))'.

that's recommended on perl tutorial website http://perldoc.perl.org/perlop.html.

maybe if you do want 'or' these lines. 
                                       (
 						!$config{"partySkill_$i"."_target"}
-						or existsInList($config{"partySkill_$i"."_target"}, $player->{name})
-						or $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
-						or $char->{homunculus} && $player->{ID} eq $char->{homunculus}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@homunculus')
-						or $char->{mercenary} && $player->{ID} eq $char->{mercenary}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@mercenary')

it's mean you do want value !$config{"partySkill_$i"."_target"}
-						or existsInList($config{"partySkill_$i"."_target"}, $player->{name})
-						or $player->{ID} eq $char->{ID} && existsInList($config{"partySkill_$i"."_target"}, '@main')
-						or $char->{homunculus} && $player->{ID} eq $char->{homunculus}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@homunculus')
-						or $char->{mercenary} && $player->{ID} eq $char->{mercenary}{ID} && existsInList($config{"partySkill_$i"."_target"}, '@mercenary')
)

That do you want any value for 'if', you don't want comparison it.

Ah' that why they is in () maybe good enough than, i see.

You have got idea and understand my desciptions.

PR #198, opened 2016-08-19
iRO: support the new 0A09, 0A0A, 0A0B, 0A0C, 0A0D, 0A0F, 0A10, 0A2D p…

allanon:
…ackets, and support the changed 133 and 136 packets

Episode 15.2 adds "options" to items. (A new kind of card-like thing?) This changes item packets, adding a bunch of new ones. And also, quite annoyingly, modifying some old ones.

- [ ] Code review
- [x] QA Review

The code for the modified vend packets in the iRO server overrides the one in ServerType0. Maybe this should be a flag in `servers.txt` instead?


cydh:
In emulator we call it **random options** as what the client-side's lua named.

The `options` parts (as so we call it Random Options), you can unpack them again as `unpack('v2 c v2 c v2 c v2 c v2 c', $item->{options});`

You can found in https://github.com/rathena/rathena/blob/03adcd9bce2d8700ab742a7689b2fa37902a49b7/src/map/clif.c#L2439-L2449
```c++
void clif_add_random_options(unsigned char* buf, struct item *it) {
#if PACKETVER >= 20150226
	int i;

	for (i = 0; i < MAX_ITEM_RDM_OPT; i++) {
		WBUFW(buf, i*5 + 0) = it->option[i].id;		// OptIndex
		WBUFW(buf, i*5 + 2) = it->option[i].value;	// Value
		WBUFB(buf, i*5 + 4) = it->option[i].param;	// Param1
	}
#endif
}
```


And, the `unknown` on this code, is `nRandomOptionCnt` (that's what we believed) https://github.com/rathena/rathena/blob/03adcd9bce2d8700ab742a7689b2fa37902a49b7/src/map/clif.c#L2615
```
'0A0C' => ['inventory_item_added', 'v3 C3 a8 V C2 a4 v a25', [qw(index amount nameID identified broken upgrade cards type_equip type fail expire unknown options)]],
```

However, as long as this PR is work, I can just add this expanded feature later.
Just tell the important info so OpenKore also can support this.
(So, if I'm going to add this, I'll test on local with @rAthena and please do review in iRO @allanon @itsrachelfish)

N3evin:
It works fine!

itsrachelfish:
@N3evin and David___ on IRC tested these changes. Both said everything works! 

I just reviewed all the code and it looks good. Thanks @allanon :smile:

PR #200, opened 2016-08-19
add very limited support for random options in item names

allanon:
- [x] QA Review
- [x] Code Review (this code is at best incomplete - only the options which are currently implemented in iRO are expected to work)

This shows the new Crimson weapons with "Fire", "Ice", etc in the name.

I've also added a data file which was extracted from `data\luafiles514\lua files\datainfo\addrandomoptionnametable.lub`. (Speaking of which, does anybody have an easy way to decompile a lub? `luadec` gave me header errors.)


cydh:
@allanon for luadec, use for 5.1.4 https://github.com/ROClientSide/Translation/blob/master/Dev/Tools/luac5.1.4.exe

and the question for crimson weapon on iRO, the element given since the item is dropped or while identifying?
I meant, by using this update, at least if you got the crimson weapon u can look into the item list it has randomoption or not yet.
Thank you

allanon:
I'll have to see if luadec5.1.4 has a Linux version, since I don't use Windows.

The options are in the item packet even before the item is identified, so yes, decisions could be made on whether to pick up (or drop, sell, store, etc) the item without having to identify it first.

deado:
Started testing this pull on iRO for Crimson weapon Elements support. This pull request was made before the recent addition/changes of packets 09DB, 09DC, 09DD, etc. so those are showing up Unknown Switch... However, it seems to read all the elements nicely (at least all the ones I currently have in a cart). Works if the item is identified or not. IMO these changes for the crimson element support could be merged with the main repo

PR #201, opened 2016-08-21
Improve attack loot monster performance.

systeman:
Improve performance attack monster looter.

### DIFF
--- a/control/consolecolors.txt
+++ b/control/consolecolors.txt
@@ -38,6 +38,7 @@ partychat brown
 selfchat green
 schat yellow
 list white
+looter yellow
 info white
 equip grey
 teleport yellow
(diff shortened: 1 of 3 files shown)


itsrachelfish:
Hi @systeman. Thank you for proposing changes to the project. Could you please give some more information on the problems you're trying to solve and how these changes fix them? Thanks!

systeman:
Actually, I have resolved problems of Openkore by the private solve and this is with one the problem, I resolve to achieve real and quick when the monster came to loot. Openkore can indicate that item is the aim, there can maintain to fast and properly.

________________________________

itsrachelfish:
This cannot be accepted without detailed comments explaining the intended purpose of your logic changes

PR #214, opened 2016-09-01
Removed deal action check from processDead sub.

deezar:
Removed "deal" action check from processDead sub. Is that relevant here?

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -946,7 +946,7 @@ sub processDead {
 			}
 		}
 
-	} elsif (AI::action ne "dead" && AI::action ne "deal" && $char->{'dead'}) {
+	} elsif (AI::action ne "dead" && $char->{'dead'}) {
 		AI::clear();
 		AI::queue("dead");
 		$char->setStatus('Dead', 1);


itsrachelfish:
Maybe this was added by someone who was trying to deal their bot Yggs before they auto-respawn?

deezar:
I'm not sure either. 
But isn't Yggdrasil Leaf only usable if a player isn't dead?
Moreover the processDead should be independent from respawn action. If that's the case it should check just before respawn.
A quick glance what's inside this IF statement has nothing to do with "deal" action.

itsrachelfish:
@deezar I just did a bit of digging into the code history. This change was intentionally added 11 years ago to ensure that dealAuto works even when the player is dead.

https://github.com/OpenKore/openkore/commit/36d1d3cecc0612dd40f380415f5aeeaa0a26e784

Is there a reason that you think this check should be removed?

allanon:
While it's true that dealing an Ygg Leaf _to_ a dead player is useless, dealing an Ygg Leaf _from_ a dead player to a live player would allow the live player to resurrect the dead player.

farrainbow:
It's not the best implementation but the purpose makes sense for me.

kaliwanagan:
I'll be closing this PR for now. I think the original scenario was e.g. a priest that had lots of ygg leafs, following a tank (e.g. a blacksmith or a knight) that only loots. If the priest dies, the priest can deal the tank an ygg leaf so the tank can resurrect the priest.

PR #220, opened 2016-09-12
support packet 0441 (skill_delete)

allanon:
- [x] QA Review (ie, does it work?)

iRO uses this packet when a skill given by equipment (eg, Heal given by a Vitata Card) is removed.


itsrachelfish:
Tested with a sight clip, works as expected!

PR #253, opened 2016-09-26
Support for 02F1

lututui:
Adds support for 02F1 (PACKET_CZ_PROGRESS), packet which notifies the server that the progress bar has been completed.

Reference (pt-br): http://bit.ly/2dmpUhg


lututui:
Note: we don't really need to deal with this packet, since openkore uses a timer in order to know when the progress bar was completed. This pull request deals only with the Packet Parser warning.

emsjustin:
Nice update.

Henrybk:
I made some changes to the progress bar support back in r8978 and r8995 but back then I made sure that the servers didn't send anything to the client when the progress bar was over, guess I forgot to add support for the pack haha.

Also, wouldn't it be a good idea to add it to Sakexe_0.pm also?

lututui:
Since its counterparts are listed in Sakexe, yes.

Henrybk:
I guess it's okay now, thanks for the update @lututui

PR #258, opened 2016-10-12
Poseidon compatibility with bRO

lututui:
Adds missing charblocksizes to poseidon server
Adds bRO_2016-10-11a as a poseidon servertype
Adds '09D0' as a gameguard sync packet


emsjustin:
Awesome lututui!
With this modifications is possible to use poseidon in bRo servers.
But, this update just respond one client per openkore ...

Henrybk:
Can someone confirm this works?

emsjustin:
This works.
But, need update every week because the bRo client change...

lututui:
Updating the poseidon servertype is not mandatory. You can use older clients, as long as they are charblocksize 124.

PR #264, opened 2016-10-13
storage_gettocart changes

AlexandreBR3N:
- [x] Code review
- [x] QA testing

Check if cart exists before send item straight from storage to cart !

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -4661,6 +4661,7 @@ sub cmdStorage_addfromcart {
 	if (!defined($amount) || $amount > $item->{amount}) {
 		$amount = $item->{amount};
 	}
+	
 	$messageSender->sendStorageAddFromCart($item->{index}, $amount);
 }
 
@@ -4702,6 +4703,11 @@ sub cmdStorage_gettocart {
 
 	if (!defined($amount) || $amount > $item->{amount}) {
 		$amount = $item->{amount};
+	}
+	
+	if ($char->statusesString =~ /.*?Cart.*?/ig) {
+		error TF("Cart wasn't found.\n");
+		return;
 	}
 	$messageSender->sendStorageGetToCart($item->{index}, $amount);
 }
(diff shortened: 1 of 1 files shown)


itsrachelfish:
Looks good, pretty simple fix. @AlexandreBR3N what happens currently? Does the server disconnect you? Does the item not move? An explanation of the current behavior would be helpful while testing.

src/Commands.pm
```diff
@@ -4703,6 +4703,11 @@ sub cmdStorage_gettocart {
 	if (!defined($amount) || $amount > $item->{amount}) {
 		$amount = $item->{amount};
 	}
+	
+	if ($char->statusesString !~ /.*?Cart.*?/ig) {
```
allanon: Please use `$char->cartActive` instead.

ChrstphrR:
Regarding the behaviour, when I tested on iRO's renewal server, with a character that had no cart (and had the wrong class, and thus, no ability to EVER have a cart...
*Surprisingly*

- No disconnect on sending a "storage gettocart # #" command
- The item(s) in question don't disappear, nor get duplicated

It just fails out silently, and I should also note, on the openkore/client side:

We are NOT warning the user that they're trying to do this on a non-cart equipped character.

AlexandreBR3N:
I'm using on bRO. 
Before i put my merchant to sell, he has 600 oridecon minerium on storage. I sold only 250. After this, he tried to put straight from storage to cart, but the item, for my surprise, disappeared. I don't know why. But i really checked if i had 600 minerium and my log don't lies. Then i tried on the second account, for my surprise, i did with lesser amount and then the item disappeared. Don't appears nothing. Just look like the item was moved to cart, but i had no cart and the item was lost.

itsrachelfish:
@AlexandreBR3N Can you update this pull request to use the ```$char->cartActive``` variable that allanon mentioned? Once you do that I'll be happy to merge this.

itsrachelfish:
Thanks @AlexandreBR3N I just tested and it's working as expected :+1:

PR #267, opened 2016-10-14
bRO's skillnametable.txt from skillinfolist.lua

lututui:
Updates bRO's skillnametable from its decompiled skillinfolist.lub (path in data.grf: data\luafiles514\lua files\skillinfoz\skillinfolist.lub).

All skills are present, although many were shifted around due to the differences between the sequences in the new lub file and the old text file.

This solves the missing skills problem, mainly after the latest episode patches.


flacrasec:
Wow this is great lututui thanks to make these changes!

Henrybk:
Seems okay. thanks for the update.

Issue #271, opened 2016-10-16
NPC Talk Responses output truncated to 36 characters, response numbers are left justified.


ChrstphrR:
***Synopsis:***
This is an issue that hails back to the commit for a [message/formatting update in August of 2014](https://github.com/OpenKore/openkore/commit/f8afd7c750022fa59dfee8c651fdca557dd8aaeb#diff-979dcabd09d1189f5069504f05191ad2R4718)

*Issue A*
A longer response (longer than 36 characters) is sent to the user when talking to an NPC, the response text is truncated.   When this commit was created, many of the changes were to tabular output, formatting the output to conform to the bounds of the chart output.   The same was done to the NPC talk responses, but some official server response lines are longer than 36 characters.

*Issue B*
The second issue that arises from this, when there are more than 10 responses possible (it happens!), the numbers are left justified, when right justification is the norm for tabular number fields.

***Example of the issue***

As it happens, there is one NPC on Renewal servers that displays both issues with ease!
On a renewal server, go to payon_in01, and talk to "Inventor Jaax" (5, 134) in the northwest corner of the room.

Part of the NPC conversation, that displays Issue B (happens first in the conversation, so B comes first... deal with it)

` ------ Responses (Inventor Jaax) -------`
`#  Response`
`0  Quiver`
`1  Iron Arrow Quiver`
`2  Steel Arrow Quiver`
`3  Oridecon Arrow Quiver`
`4  Fire Arrow Quiver`
`5  Silver Arrow Quiver`
`6  Wind Arrow Quiver`
`7  Stone Arrow Quiver`
`8  Crystal Arrow Quiver`
`9  Shadow Arrow Quiver`
`10 Immaterial Arrow Quiver`
`11 Rusty Arrow Quiver`
`12 Holy Arrow Quiver`
`13 Cancel Chat`
`----------------------------------------`
`Inventor Jaax: Type 'talk resp #' to choose a response.`

The Issue A occurance, where the response get truncated:

`------ Responses (Inventor Jaax) -------`
`#  Response`
`0  Store as many arrows in quivers as p`
`1  Purchase 1 quiver`
`2  Cancel`
`3  Cancel Chat`
`----------------------------------------`

The truncated response here should be "`Store as many arrows in quivers as possible`"

***Info***
**OS:**
>Ubuntu 16.04.1 LTS

Using the server distro, if that matters.
Kernel issue from "uname"

> 4.4.0-42-generic GNU/Linux

**Perl:**
[...]

PR #273, opened 2016-10-16
Fix formatting for NPC talk response choices #271

ChrstphrR:
***Fixed two issues outlined in Ticket #271:***

**Issue A** - Text for the NPC talk response choices were cut off if they
were longer than 36 characters.

**Issue B** - Numeric response codes for the choices need to be right
justified because up to two digit codes (i.e. more than 0-9 choices)
are possible.

Fixed by referencing: http://perldoc.perl.org/perlform.html

ticket:   #271
branch:   talk_response
modified: src/Commands.pm

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -4828,7 +4828,7 @@ sub cmdTalk {
 			TF("#  Response\n");
 		for (my $i = 0; $i < @{$talk{'responses'}}; $i++) {
 			$msg .= swrite(
-			"@> @*",
+			"@< @*",
 			[$i, $talk{responses}[$i]]);
 		}
 		$msg .= ('-'x40) . "\n";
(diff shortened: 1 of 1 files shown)


ChrstphrR:
Tested changes to formatting on iRO, using the same NPC mentioned in the steps-to-reproduce listed in ticket #271 

- [x] Numeric response codes are now right justified, up to two digits
- [x] Response text length is not restricted to the width of the "Response" header/footer

Please test independently to confirm.

itsrachelfish:
@ChrstphrR Can you provide an example of an NPC that gives more than 9 response options so I can test? Thanks!

ChrstphrR:
Yes, that's the one - but to save anyone searching:   on a renewal server:   payon_in01 5, 134 (Inventor Jaax) tests out both formatting changes in two of the response choice lists it provides.

itsrachelfish:
@ChrstphrR As per my comment in the issue, I think we should keep the left alignment for now to remain consistent with all of the other list commands.

If you really think that all lists should use right alignment, that should be a separate pull request.

Other than that, I tested and confirmed that the truncation issue is now fixed!

PR #195, opened 2016-08-19
Add support for RMS Test Server

lututui:
- [x] QA
- [x] Code Review

Adds initial support for RagexeRE_2014_10_22b, based on RMS Test Server. This server used to be supported, and is still listed in servers.txt (entries 20 and 21). recvpackets.txt and cryptKeys were generated by PEEK

RMS Test Server: http://play.ratemyserver.net/ragnarok/

Known issues:
* ~~Picking up items doesn't work every time ('item_take')~~ 5ded68f

PR #275, opened 2016-10-17
Servertype ragexe 2014 10 22b

Henrybk:
This pull request adds support for the servertype Ragexe_20141022b which many private servers use nowadays. I was only able to test it in one or 2 servers but it seemed to work fine.

The servertype was created based on the source code from rAthena.

The recvpacket were generated from a Ragexe_2014-10-22b I diffed myself.


itsrachelfish:
@Henrybk Do you have any examples of what pservers this pull request can be tested on? Thanks :)

Henrybk:
I could not find the servers I tested at haha.

Well it's working here and the packets aparently are right by this:
https://github.com/brAthena/brAthena/blob/master/src/map/packets.h
And this
https://github.com/rathena/rathena/blob/master/db/packet_db.txt

I guess I will have to find another server.

itsrachelfish:
@Henrybk How is this different from the pull request @lututui created before?

https://github.com/OpenKore/openkore/pull/195

Henrybk:
@itsrachelfish There is a difference between ragexe and ragexeRE, if you try to use this servertype to log into RMS server you won't be able to do it, and if you use #195 code to log into a server diffed with 2014-10-22ragexe hexed you won't be able to also.

itsrachelfish:
Thanks for the clarification! :+1:

magimagica:
I got this message when someone using a skill
Packet Parser: Unknown switch: 09CB

m3kgin3er:
09CB now was fixed using https://github.com/OpenKore/openkore/issues/516

a2362998:
[iRO] Packet Tokenizer: Unknown switch 09DB, 09DD  could someone help me pls  having problem with this

m3kgin3er:
You should be banned of github.

magimagica:
I get Unknown Switch: 09CA from time to time when hunting, not sure because of monster's skill or something else.

LoseracK:
Well the correct way maybe is just to change it inside sT 0

Slipxxliose:
What timestamp you've got?
What sT is u using ?

kaliwanagan:
@Henrybk I'll close this for now. When those servers pop up again then we'll either rework on the new packets, or revisit this PR (if we remember)

PR #285, opened 2016-10-21
Fix in function loadByRegexp and add loadFiles

Henrybk:
- [x] Code review
- [x] QA review

Now only files that match the regex will be hooked


itsrachelfish:
Tested these changes, seems to be working fine!

PR #287, opened 2016-10-22
fix #269 - add more useful error when XKore2 has a port conflict

### LINKED ISSUE #269: XKore_listenPort: Cryptic, ungraceful failure when port number is already in use.
**Overview:**
A cryptic, ungraceful failure occurs when XKore_listenPort number is already used as a TCP/IP listening port on the same machine.   The message that results seems to be due to an unhandled exception, so there's no sanitized message that would help the user correct their configuration issue.

**Symptoms:**
Tail of my logs shows what the hapless user sees:

> Loading tables/hateffect_id_handle.txt...
Loading tables/hateffect_name.txt...

> [Network::Receive] Network::Receive::iRO  (mode: new)
[Network::Send] Network::Send::iRO  (mode: new)
XKore 2 Map Server started
[Network::Receive] Network::Receive::iRO  (mode: new)
[Network::Send] Network::Send::iRO  (mode: new)
IO::Socket::INET: Address already in use
Benchmark results saved to benchmark-results.txt
macro plugin unloading, cleaning up
*** OpenKore what-will-become-2.1 ( version bc07ae7_2016-10-14 ) - Custom Ragnarok Online client ***
***   http://www.openkore.com/   ***
Bye!

**Cause:**
In the control file _config.txt_ when XKore is configured (XKore 2 in my/this case)

> XKore_listenPort PORTNUMBER_HERE

When PORTNUMBER_HERE is already used as a listening port (for example, in another openkore instance using the same port number), the above IO::Socket::INET error will occur.

Now, before someone just tut-tuts and sweeps this away, saying that it's good enough...

The related code for this is in  [src/Network/XKoreProxy.pm:](https://github.com/OpenKore/openkore/blob/master/src/Network/XKoreProxy.pm#L306-L317) 

> 		# (Re)start listening...
		my $ip = $config{XKore_listenIp} || '127.0.0.1';
		my $port = $config{XKore_listenPort} || 6901;
		$self->{proxy_listen} = new IO::Socket::INET(
			LocalAddr	=> $ip,
			LocalPort	=> $port,
			Listen		=> 5,
			Proto		=> 'tcp',
			ReuseAddr   => 1);
		die "Unable to start the X-Kore proxy ($ip:$port): $@\n" . 
			"You can only run one X-Kore session at the same time.\n" .
			"And make sure no other servers are running on port $port." unless $self->{proxy_listen};

It appears that the IO::Socket::INET exception is raised, and not caught gracefully because we're checking for a return code with that trailing **die** statement, and this is not catching the error created.

**Solution:**

For the user:

- Check which ports are in use on your machine you are running openkore.
 - Using "netstat" from the console, 
 - Checking your multiple openkore bot control files to ensure you're not just copy-pasting the same settings for XKore
- Correct any duplicate XKore_listenPort settings that conflict

For the devs:
- Find out why/how this error is generated
- Fix how openkore handles this situation, so the error is trapped, and then we can warn the user about the port conflict in a more graceful way.

allanon:
- [x] QA Review (does this work?)
- [x] Code Review (does the code look good?)

This PR fixes #269.

### DIFF
--- a/src/Network/XKore.pm
+++ b/src/Network/XKore.pm
@@ -49,8 +49,7 @@ sub new {
 		Proto		=> 'tcp');
 	if (!$self->{server}) {
 		Network::XKore::CannotStart->throw(error => TF("Unable to start the X-Kore server.\n" . 
-			"You can only run one X-Kore session at the same time.\n" . 
-			"And make sure no other servers are running on port %s.\n", $port));
+			"Make sure no other servers are running on port %s.\n", $port));
 	}
 
 	$self->{incomingPackets} = "";
--- a/src/Network/XKore2.pm
+++ b/src/Network/XKore2.pm
@@ -20,6 +20,7 @@ package Network::XKore2;
 use strict;
 use Globals qw(%config %rpackets $masterServer $char);
 use Utils qw(makeCoordsDir makeCoordsXY makeCoordsFromTo calcPosition);
+use Utils::Exceptions;
 use Plugins;
 use Base::Ragnarok::SessionStore;
 use Network;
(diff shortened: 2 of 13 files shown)


itsrachelfish:
@allanon This looks good, I tested it locally and it outputs an informative error message saying the address is already in use, but don't you think the XKore 3 error message should be updated as well?

PR #289, opened 2016-10-25
Add profile changing support

Henrybk:
Add initial support for profile changing during runtime.


kaliwanagan:
Merged. Thanks!! 👍

PR #313, opened 2016-11-08
Set loadPlugins back to 1

itsrachelfish:
Originally we wanted to only load certain plugins by default and those would be considered "officially supported and tested"

Unfortunately this means if you want to use your own plugins you have to modify the sys.txt config file which can potentially lead to merge conflicts.

Because all "untested" plugins are currently in the "needs-review" folder, there is no reason not to load all plugins by default.

### DIFF
--- a/control/sys.txt
+++ b/control/sys.txt
@@ -44,7 +44,7 @@ panelTwo_fontsize 8
 #   1 : load all plugins
 #   2 : only load plugins that are listed in loadPlugins_list
 #   3 : load all plugins except those listed in skipPlugins_list
-loadPlugins 2
+loadPlugins 1
 
 # loadPlugins_list <list>
 #   if loadPlugins is set to 2, this comma-separated list of plugin names (filename without the extension)


allanon:
:+1: :heart:

flacrasec:
the first time I' ve started openkore with all plugins on at bRO, I had some problems with them, soon I´ll try again and bring news.

lututui:
@flacrasec if the plugins are kept in the needs-review folder, they are deep enough that openkore won't load them

Henrybk:
Well, the only problem I see in setting it back to 1 is that all 'reviewed' plugins will be loaded, the commom user may not like that if we keep reviewing more plugins.

c4c1n6kr3m1:
just because "officially supported and tested" , doesn't mean i need it
so, no !!! keep loadPlugins 2

but must give a good readme.txt or something, so noob know how to use plugins and edit sys.txt

itsrachelfish:
I came up with a couple alternative ideas based on the feedback so far:

**Idea 1** What if the plugins directory was set up like a traditional apache / nginx configuration with ```plugins-available``` and ```plugins-enabled``` folders?

The contents of ```plugins-enabled``` would be gitignored so we could change which plugins were enabled without worrying about what's being tracked in git.

**Idea 2** The profiles plugin could automatically reload plugins based on the sys.txt options in a profile. We could have different sys.txt configurations loaded per profile, and profiles are already gitignored so there would be no need to change the default.

This could be especially helpful in situations where you want to use specific plugins for different characters.

Henrybk:
@itsrachelfish I tried doing that once but sys.txt is loaded before plugins that change the control folder.

farrainbow:
First we really should set `loadPlugins 1` and then consider other things.
>Because all "untested" plugins are currently in the "needs-review" folder, there is no reason not to load all plugins by default.

itsrachelfish:
@farrainbow Feel free to merge this if you think it should be set to 1 before we consider alternatives. :wink:

kaliwanagan:
@farrainbow are we good with loading all plugins by default? Otherwise we'll close this and leave it to documentation to say that if you want to load all plugins, you need to set this option.

Mortimal:
i think this is good idea

Nipodemos:
I think we should close this PR, since it's not being discussion about this in more than one year

But I liked the idea1 of plugins enabled and plugins available

Nipodemos:
closing this because of Inactivity

PR #315, opened 2016-11-08
Remove unused dependency

lututui:
This was used by the autoupdater tool and only by it, which is now deprecated. Removing this dependency makes no difference whatsoever.

autoupdater source ->https://svn.code.sf.net/p/openkore/code/openkore/trunk/autoupdate.pl
wiki article -> http://wiki.openkore.com/index.php/AutoUpdater


flacrasec:
Cleaning unused files.

PR #321, opened 2016-11-17
[2016-11-17] Connection Files for bRO

lututui:
http://openkorebrasil.org/index.php?/topic/1855-atualiza%C3%A7%C3%A3o-17-11-2016/


flacrasec:
thank you as usual it is working just fine! =)

PR #327, opened 2016-11-24
Add Route from Airport to Einbroch at bRO

flacrasec:
added the correct route from airport to einbroch at portals.txt


itsrachelfish:
Cool, thanks :smile:

PR #328, opened 2016-11-24
add new version of splendide map (with access to bif_fild01)

allanon:
Update the splendide map to allow access to bif_fild01 (and thus, eventually, mora).


flacrasec:
very nice.

PR #331, opened 2016-11-28
Fix crash when server config option is invalid

farrainbow:
Message with server info should be displayed only if specified server exists in the list.


allanon:
Can you please give an example of what caused this crash, so I can test it?

farrainbow:
Manually set `server` in config to an excessively big value.

That happens in normal workflow when servers are merged or closed and server list goes down in size.

itsrachelfish:
Works great! 
invalid-server

PR #332, opened 2016-11-28
Remove support for deprecated encrypt config option

farrainbow:
This is really long unused option which plagues the network code.


itsrachelfish:
Looks good! Tested this branch on iRO and it seemed to work fine.

Issue #330, opened 2016-11-25
Poseidon - GameGuard


AlexandreBR3N:
http://forums.openkore.com/viewtopic.php?f=36&t=211651&p=274718#p274718

Urgent help.. Almost 80% code fixed.
To emulate the GG, is only missing this part of code. And it was encrypted by themida. I need help to translate.
I know there are monsters in ASM here ! :D Really ! It's part of kore. Heartbleed packets

AlexandreBR3N:
Sorry for that. i've already solved this asm problem. this is the blowfish_encrypt routine. 

sub GGBE {	#GameGuard::Blowfish_Encrypt
	my $cipher = new Crypt::Blowfish $inca_salt;
	$output_keys[0] ^= 0xD253FC2E;
	$output_keys[1] ^= 0x630FC331;
	unpack("H*", $cipher->encrypt(pack("H16", $output_keys[0], $output_keys[2])));
	$output_keys[2] ^= 0xCC0C559F;
	$output_keys[3] ^= 0x1AC6186C;	
	unpack("H*", $cipher->encrypt(pack("H16", $output_keys[1], $output_keys[3])));
	GLOG("GameGuard::BE called");
}

PR #337, opened 2016-12-01
[2016-12-01] Connection Files for bRO

lututui:
http://openkorebrasil.org/index.php?/topic/1929-atualiza%C3%A7%C3%A3o-01-12-2016/


flacrasec:
The poseidon is at the same situation from last week? thx again for the update lututui ^^

itsrachelfish:
@lututui Can you explain what happened with poseidon last week?

I can't read that Openkore Brasil forum page without registering. :(

flacrasec:
since the last update the ragexe.exe of bRO became incompatible with the poseidon the servertype.txt does not work at all, so the solution to use poseidon was to make a copy from the last ragexe.exe that was compatible and keep using it that is how we are using until now, but for any more details @lututui can help more than me.

flacrasec:
ah @itsrachelfish but this is the weekly update, and can be merged, it is working fine!

lututui:
@itsrachelfish basically, bRO's Gameguard updated and any client older than bRO_2016-11-08aRagexe does not answer the query packet. That might be caused by a new answer packet we're not expecting (best scenario) or the packet structure changed completely.

Actually, bRO_2016-11-08aRagexe (or any 2016 client for that matter) works perfectly, and anyone can get their hands on the Ragexe.exe from Lu!G (http://openkorebrasil.org/index.php?/topic/1857-obtendo-o-ragexeexe-do-poseidon-2016-11-08a/).

The best solution would be helping @AlexandreBR3N with emulating GameGuard (#330). If just the answer packet has changed, it's something easy to fix, and I'll be looking into it. Otherwise, people can stick with bRO_2016-11-08aRagexe just fine, at least for now.

At last, this pull request just updates the old boring packets for bRO, nothing to do with Poseidon.

AlexandreBR3N:
[quote]
No packet has actually changed. The problem is that the cryptKey for the packet '09D0' has something extra. I'll open an issue just for this later.
[/quote]

I was seriously thinking in one new param to identify the bot ! :D

AlexandreBR3N:
I was seriously thinking in one new param to identify the bot ! :D 

dwIndex, dwValue1, dwValue2, dwValue3, poseidonIdentification

PR #341, opened 2016-12-03
Add hook 'AI_start'

Henrybk:
With this we can hook something to all AI states and even execute stuff when AI is off or in manual, and only when we are connected and $timeout{ai} is off.

I think this is an awesome hook and may be way better than 'mainLoop_pre' in many situations.


allanon:
:+1:

PR #343, opened 2016-12-05
Fix building on OS X

farrainbow:
Can't get Benchmark module to build correctly, so disabled that on OS X for now.

Run `make clean all` first.


farrainbow:
Tested on Debian 8 and OS X 10.9.

PR #349, opened 2016-12-08
Map Property Info Updates

cydh:
Map Property Info Updates
* Updated MAPPROPERTY_INFO_id_handle.txt with correct entries. The current entries are not properties that used to tell the client!
* Corrected the `map_property` and `map_property3` functions to process `info_table` values.

Misc
* Added maptype `0x19 MAPTYPE_SIEGE_LOWLEVEL` (25) for TE Maps

Signed-off-by: Cydh Ramdh <cydh@pservero.com>


farrainbow:
Thanks!

PR #353, opened 2016-12-09
Add 'equipped_item' and 'unequipped_item' hooks

Henrybk:
Just adding these two useful hooks, I am already using them with no problems at all, as you can see the code I made for them in eventMacro here: https://github.com/Henrybk/eventMacro/commit/09b2b25499ea7935d1f0920062512a6aaf418057


itsrachelfish:
Looks good :+1:

PR #357, opened 2016-12-12
Overload array dereferencing for ObjectList

farrainbow:
`@$playersList` is much shorter than `@{$playersList->getItems}` and IMO doesn't hurt readability.

Also, if `$list` is undefined, `@$list` would successfully evaluate to empty array, so in some situations it's not even needed to check if `$list` is defined.

- [ ] QA review
- [ ] code review
- [ ] test AI features

### DIFF
--- a/src/AI.pm
+++ b/src/AI.pm
@@ -294,7 +294,7 @@ sub ai_getAggressives {
 	my $num = 0;
 	my @agMonsters;
 
-	foreach my $monster (@{$monstersList->getItems()}) {
+	for my $monster (@$monstersList) {
 		my $control = Misc::mon_control($monster->name,$monster->{nameID}) if $type || !$wantArray;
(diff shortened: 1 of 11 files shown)


FireflySkye:
error upon opening a shop
untitled

farrainbow:
@FireflySkye it's not related to this PR, but it's now fixed.

Henrybk:
I like the idea, will test later today.

kaliwanagan:
I'm not sure. I feel it's too magical. Then again I've been away from perl for far too long to comment about idiomatic perlisms :stuck_out_tongue: 

I like the `for instead of `foreach` but the magical `@` might be a bit too much (and being explicit in the var type can be useful for debugging, or for code you want to take a look at a few years from today).

What do you think @farrainbow?

PR #358, opened 2016-12-12
Inventory hooks additions and changes

Henrybk:
Added the hooks which were in the 'TODO' comments.


FireflySkye:
untitled

farrainbow:
Maybe name them with something like `inventory_ready` and `cart_ready`? I guess it's not important that we received info, but that the thing become ready to work with.

Henrybk:
In inventory that would be true, but $cart->info is called everytime a change happens in cart state (weight, number of items, etc) so I think 'cart_ready' wouldn't be the best choice.

I agree with the inventory one.

Issue #362, opened 2016-12-13
Rethink talk and talknpc


Henrybk:
Right now we have many bugs with npc conversarions, such as auto talk continuation having no timeout, no way to talk to npcs which need to be approached, no way to use talknpc without coordinates, no way to use talk command without npc binID.

I think talknpc task should be initialized everytime a npc conversation is started, be it by command or by approaching it, and all npc management should be done by it, including all talk commands.

AlexandreBR3N:
Everything should be done until he get this objective. Close the chat...

PR #363, opened 2016-12-15
Rework on Task::TalkNPC

### LINKED ISSUE #40: autotalkcont bug
On some servers the lag sometimes makes openkore disconnect when talking to npcs if the config key autotalkcont is set to '1'

### LINKED ISSUE #367: Problems when 'talk no' is used instead of 'talk cont'
Using 'sendTalkCancel' (talk no | talknpc n) when a 'sendTalkContinue' (talk cont | talknpc c) is expected causes openkore to get disconnected on bRO and on iRO, in RMS renewal test server it causes the player to the stuck in the cell but won't disconnect it.

Since 'sendTalkCancel' is sent everytime a 'talknpc' command fails I think it should be fixed ASAP.

I would love if someone could provide me with information regarding when the client actually sends  'sendTalkCancel' to the server and in which situations it can be sent without problems.

Edit 1: Since apparently all conversations in the real client end with a 'npc_talk_cancel' after a 'npc_talk_close' shouldn't we send it ALWAYS?

Henrybk:
Changes suggested in #362 

Changes so far:
1 - Made 'talk' command use `Task::TalkNPC`
2 - Mapchanges now clear `NPC` from `AI` queue
3 - Now `TalkNPC` tasks are auto created if a npc conversation starts and we don't have one
4 - Now `autoTalkCont` is handled by `Task::TalkNPC` ( fixes #40 )
5 - Now the next conversation step will only be sent if the server has responded to the last one, this was done to avoid DC's when the server is lagging, instead we only cancel the task.
6 - Changed `$talk{'buyOrSell'}` to `
$ai_v{'npc_talk'}{'talk'} eq
buy_or_sell`
7 - Removed `talk no` and `talknpc n` ( fixes #367 )
8 - `talk cancel` will always be sent after `talk close`
9 - Moved some dupped subs to `Receive.pm`
10 - Changed some `%talk` changes in `Send.pm`
11 - Rewrote all npc talk steps logic in `TalkNPC`
12 - Added a kind of intelligent logic to try to end a conversation after a bug happens
13 - Now you can add multiple steps at once using the command `talk`
14 - Now all talknpc sequences are validated before the conversation starts
15 - Messages like `Type 'talk xxxx' to continue conversation` won't appear if the next conversation step is defined.

TODO:
1 - Test on npcs with no talk or buy/sell interaction (private healer npcs that activate on click)


Henrybk:
Also I think we should implement these ideas:
http://forums.openkore.com/viewtopic.php?f=36&t=182288
http://forums.openkore.com/viewtopic.php?f=36&t=10125

src/functions.pl
```diff
@@ -644,7 +644,10 @@ sub initMapChangeVars {
 	undef %items;
 	undef %spells;
 	undef %incomingParty;
+	
 	undef %talk;
+	AI::clear("NPC");
```
farrainbow: TalkNPC task can hook map change event and mark itself as done (`setDone`) in the next `iterate` call.

src/Task/TalkNPC.pm
```diff
@@ -70,33 +80,107 @@ sub new {
...
+			$self->{steps} = [parseArgs($self->{sequence})];
+		} else {
+			$self->{steps} = [];
+		}
+	} else {
+		if ($self->{type} eq 'talknpc') {
+			$self->{steps} = [parseArgs("x $self->{sequence}")];
+		} else {
+			push (@{$self->{steps}}, 'x');
+		}
```
farrainbow: Maybe ignore `$self->{type}` and always use `$self->{sequence}` (looks like it's empty when `$self->{type} ne 'talknpc'`?).

src/functions.pl
Henrybk: Because the task will only be cleaned in the first AI cycle after the mapchange, this can cause major bugs if another npc conversation starts before that, like a npc that starts conversation as soon as the character enters a certain map. (eg. Payon palace guards)

src/Task/TalkNPC.pm
Henrybk: Thats seems about right, I'll change it.

src/Task/TalkNPC.pm
```diff
@@ -394,4 +626,9 @@ sub findTarget {
 	return undef;
 }
 
+sub nextStep {
+	my ($self, $step) = @_;
+	push(@{$self->{steps}}, $step);
+}
+
```
farrainbow: Maybe name it `addSteps` and allow to pass many steps together in a single string argument, like when creating the task? Also it's not always would be "next" step since it just adds steps to the end of the sequence.

Henrybk: I also noticed this earlier today, thanks for pointing it out.

src/Task/TalkNPC.pm
```diff
@@ -70,33 +80,107 @@ sub new {
...
-	$self->{stage} = 'Not Started';
-	$self->{mapChanged} = 0;
+	$self->{stage} = NOT_STARTED;
+}
+
+sub find_and_set_target {
+	my ($self) = @_;
+	my $target = $self->findTarget($npcsList);
+	if ($target) {
+		debug "Target NPC " . $target->name() . " at ($target->{pos}{x},$target->{pos}{y}) found.\n", "ai_npcTalk";
```
farrainbow: If `$target` is an `Actor`, you can just do `"Target $target at..."` to get a suitable string representation (with actor type, binID and possibly other info). This can be used in translated strings too: `TF("Found %s", $target)`.

src/Task/TalkNPC.pm
```diff
@@ -70,33 +80,107 @@ sub new {
...
+	
+	if ($target) {
+		$self->{target} = $target;
+		$self->{ID} = $target->{ID};
+		lookAtPosition($self);
+	} elsif (exists $talk{nameID} && $ai_v{'npc_talk'}{'talk'} ne 'buy_or_sell') {#check if this is really necessary
+		$self->{ID} = $talk{ID};
+		$self->{target} = Actor::NPC->new;
+		$self->{target}->{appear_time} = time;
+		$self->{target}->{name} = 'Unknown';
```
farrainbow: Which NPCs use this branch?

src/Task/TalkNPC.pm
```diff
@@ -318,28 +444,110 @@ sub iterate {
...
+			# Usual end of a conversation
+			if (!%talk) {
+				$self->conversation_end;
+			
+			# No more steps but we still have a defined %talk
+			} else {
+				debug "After we finished talking with hte last npc another one started talking to us faster than we could predict.\n", "ai_npcTalk";
+				my $target = $self->find_and_set_target;
+				$self->{stage} = TALKING_TO_NPC;
+				$self->{time} = time;
```
farrainbow: Which NPCs do this?

src/Task/TalkNPC.pm
Henrybk: Done

src/Network/Receive.pm
```diff
@@ -3844,8 +3844,8 @@ sub npc_talk_close {
 		$messageSender->sendTalkCancel($ID);
 	}
 
-	$ai_v{npc_talk}{talk} = 'close';
-	$ai_v{npc_talk}{time} = time;
+	$ai_v{'npc_talk'}{'talk'} = 'close';
+	$ai_v{'npc_talk'}{'time'} = time;
```
farrainbow: Please next time do style changes in a separate commit, since it gets in the way of review in other places where style changes interleave with actual changes.

src/Task/TalkNPC.pm
Henrybk: NPCs scripts that start conversation without an actual npc 'body', like automatic bot checkers and automatic npc conversations on map changes.

src/Task/TalkNPC.pm
Henrybk: Any conversation that starts right after another one ends, 'payon 158 246 Guardsman' made this necessary before I added the deletion of the task at mapchange, don't know of any other.

src/Network/Receive.pm
Henrybk: Sorry, did it without thinking :P

src/Network/Receive.pm
```diff
@@ -3824,4 +3824,154 @@ sub hat_effect {
...
+
+sub npc_talk_continue {
+	my ($self, $args) = @_;
+	my $ID = substr($args->{RAW_MSG}, 2, 4);
+	my $name = getNPCName($ID);
+
+	$ai_v{'npc_talk'}{'talk'} = 'next';
+	$ai_v{'npc_talk'}{'time'} = time;
+	
+	message TF("%s: Type 'talk cont' to continue talking\n", $name), "npc";
```
farrainbow: Maybe put this message into TalkNPC task as well (and display it only when `autoTalkCont` is not set, and also when there are no more steps so user interaction is needed)?

Actually, maybe display all of those "type ... to continue talking" messages only when there are no more steps in the current sequence, so they don't appear at all when conversations are fully automated?

src/Task/TalkNPC.pm
Henrybk: Done

src/Network/Receive.pm
Henrybk: Good idea

src/Task/TalkNPC.pm
```diff
@@ -121,81 +211,83 @@ sub iterate {
...
+	# This is where things may bug in npcs which have no chat (private healers)
+	} elsif (!$ai_v{'npc_talk'}{'time'} && timeOut($self->{time}, $timeResponse)) {
 		# If NPC does not respond before timing out, then by default, it's
 		# a failure.
 		$messageSender->sendTalkCancel($self->{ID});
 		$self->setError(NPC_NO_RESPONSE, T("The NPC did not respond."));
 
-	} elsif (timeOut($ai_v{npc_talk}{time}, 0.25)) {
+	} elsif ($self->{stage} == TALKING_TO_NPC && timeOut($ai_v{'npc_talk'}{'time'}, 1.5)) {
 		# 0.25 seconds have passed since we last talked to the NPC.
```
farrainbow: Comment became wrong...

src/Task/TalkNPC.pm
```diff
@@ -206,103 +298,137 @@ sub iterate {
...
+			# Get the buy list in a shop.
+			} elsif ($step =~ /^b$/i) {
 				$messageSender->sendNPCBuySellList($talk{ID}, 0);
-			# Bulk buy solution.
-			} elsif ($step =~ /^b(\d+),(\d+)/i) {
+				
+			# Click the cancel button in a shop.
+			} elsif ($step =~ /^e$/i) {
+				$ai_v{'npc_talk'}{'talk'} = 'close';
+				$self->conversation_end;
```
farrainbow: "e" shouldn't click the cancel button and it doesn't seem it's done there; so, wrong comment?

src/Task/TalkNPC.pm
```diff
@@ -318,28 +444,110 @@ sub iterate {
...
+	# After a 'npc_talk_done' hook we must always send a 'npc_talk_cancel' after a timeout
+	} elsif ($self->{stage} == AFTER_NPC_CLOSE) {
+		return unless (timeOut($self->{time}, 1));
+		#Now 'n' step is totally unnecessary as we always send it but this must be done for backwards compatibility
+		if ( $self->{steps}[0] =~ /^n/i ) {
+			shift(@{$self->{steps}});
+		}
+		$self->{time} = time;
+		$self->{stage} = AFTER_NPC_CANCEL;
+		$messageSender->sendTalkCancel($self->{ID});
```
farrainbow: I wonder if all clients do this.

Henrybk: That needs confirmation, I tested on bRO and RMS renewal test server, it is up for changes.

src/Task/TalkNPC.pm
Henrybk: It's an old comment from the original code I haven't updated.

src/Task/TalkNPC.pm
Henrybk: Thanks, will fix.

src/Task/TalkNPC.pm
Henrybk: Actually in all servers I tested that is right, we don't send anything to the server when the client clicks the cancel button in a shop, so we just have to assume it's closed. (I ay be wrong, needs more testing)

src/Task/TalkNPC.pm
Henrybk: Done

src/Task/TalkNPC.pm
Henrybk: this code in receive.pm makes me think it is also important in other servers such as pRO and iRO.
`# I noticed that the RO client doesn't send a 'talk cancel' packet
	# when it receives a 'npc_talk_closed' packet from the server'.
	# But on pRO Thor (with Kapra password) this is required in order to
	# open the storage.
	#
	# UPDATE: not sending 'talk cancel' breaks autostorage on iRO.
	# This needs more investigation.
	#if (!$talk{canceled}) {
	#	$messageSender->sendTalkCancel($ID);
	#}`

src/Task/TalkNPC.pm
```diff
@@ -351,20 +559,44 @@ sub target {
...
+		$ai_v{'npc_talk'}{'talk'} = 'close';
+	
+	} elsif ( $ai_v{'npc_talk'}{'talk'} eq 'store' ) {
+		$self->conversation_end;
+		$ai_v{'npc_talk'}{'talk'} = 'close';
+		
+	} elsif ( $ai_v{'npc_talk'}{'talk'} eq 'sell' ) {
+		$self->conversation_end;
+		$ai_v{'npc_talk'}{'talk'} = 'close';
+
```
farrainbow: I'm not sure this is a good idea to have by default.

What happens if I have incomplete conversation sequence in talknpc command? Would I be able to finish it manually?

Which scenarios we want to accomplish with auto-finishing the conversation, if, for example, we already have `autoTalkCont` set?

We can have this but maybe there should be a config option to enable auto-finishing the conversation, and options to set text and number.

src/Task/TalkNPC.pm
farrainbow: Purpose of "e" is to just suspend the NPC conversation and allow other AI parts to do their stuff (buy/sell items, for example). So just the comment is wrong.

src/Task/TalkNPC.pm
farrainbow: So maybe we need to put that whole comment right there.

src/Task/TalkNPC.pm
Henrybk: This is only called when there's a bug in the sequence or the npc fails to respond correctly.

When the sequence is not complete it will only wait for a new step.

farrainbow: Well, when we usually want to have that? Is this just for random NPCs who automatically open a conversation?

farrainbow: Well, when we usually want to have that? I can't imagine useful scenarios of just ignoring faulty NPC sequences if they were purposely configured. Also NPC conversation would timeout after a while, wouldn't it?

farrainbow: I'm fine with hiding it behind config option and further testing, though.

src/Task/TalkNPC.pm
Henrybk: I will move it.

src/Task/TalkNPC.pm
Henrybk: What will we do when the sequence fails though? Before it would send Talk cancel ans get openkore DC'ed, this code prevents that, should it just clear all remaining steps and next the task open for new steps?

src/Task/TalkNPC.pm
Henrybk: Done

src/Task/TalkNPC.pm
farrainbow: Well anything definitely would be improvement over DC'ing. Clearing the remaining steps sounds okay for me (after that either something would add more steps or NPC will timeout). We need input from more people on that.

farrainbow:
>TODO:
>3 - Create a way to add multiple steps in a single command.

Just allow to do `->addSteps("c c r0")` and parse it the same way as task does when starting.

src/Task/TalkNPC.pm
```diff
@@ -70,33 +80,119 @@ sub new {
...
+		$target = undef;
+	}
+	
+	if (exists $talk{nameID}) {
+		$self->{steps} = [parseArgs($self->{sequence})];
+	} else {
+		$self->{steps} = [parseArgs("x $self->{sequence}")];
+		undef $ai_v{'npc_talk'}{'time'};
+		undef $ai_v{'npc_talk'}{'talk'};
+	}
```
farrainbow: What about this? (obv. addSteps needs to be able to handle this first)

```perl
	unless (exists $talk{nameID}) {
		$self->addSteps('x');
		undef $ai_v{'npc_talk'}{'time'};
		undef $ai_v{'npc_talk'}{'talk'};
	}
	$self->addSteps($self->{sequence});
```

src/Task/TalkNPC.pm
farrainbow: @itsrachelfish @allanon what do you think about this?

src/Task/TalkNPC.pm
Henrybk: Done

Henrybk:
@farrainbow Feel free to make any changes you please, maintainers can edit this pull request :P

farrainbow:
>TODO:
>2 - Add new ways to start a conversation (talk => binID, talknpc => coordinates, talkname => name, talkregex => regexOnName, etc)

A single command can handle all the cases.

>3 - Rework talk command to be able to give binId and steps in only one command like talknpc (eg. talk 0, resp 1, cont, text hello, resp 5)

Commas not really needed even.

I think commands should be changed separately and not with this PR - https://github.com/OpenKore/openkore/issues/390

src/Task/TalkNPC.pm
Henrybk: I added the config key and 4 ways to deal with the problem.

src/Task/TalkNPC.pm
```diff
@@ -121,81 +221,93 @@ sub iterate {
...
-			if ($target) {
-				$self->{target} = $target;
-				$self->{ID} = $target->{ID};
-				$self->{steps} = [parseArgs("x $self->{sequence}")];
-				undef $ai_v{npc_talk}{time};
-				undef $ai_v{npc_talk}{talk};
-				lookAtPosition($self);
-			}
+			if ($target || %talk) {
+				unless (exists $talk{nameID}) {
```
farrainbow: I guess `unless (%talk) {` would do the same thing there?

Henrybk:
I found a bug that needs to be fixed, when using `Task::TalkNPC` to route to the npc and the character approaches and auto-talk npc and it auto-starts a conversation the task won't be able to end it.

Henrybk:
Another bug: If Task::TalkNPC creates a subtask MapRoute, which also creates a subtasks TalkNPC (eg. to use kafra teleport), all subtasks hooks will be heard by the original Task::TalkNPC::handle_npc_talk.

Henrybk:
3 More bugs:
1 - When using Task::TalkNPC to route to the npc and the character approaches and auto-talk npc and it auto-starts a conversation the task won't be able to end it.
2- If Task::TalkNPC creates a subtask MapRoute, which also creates a subtasks TalkNPC (eg. to use kafra teleport), all subtasks hooks will be heard by the original Task::TalkNPC::handle_npc_talk.
3 - Sometimes when using npcs to change maps (eg. kafra) 2 `npc_talk_done` will be sent by the server, bugging the conversation

Henrybk:
The only big problems I still see are:
1 - Dealing with npcs with no interaction
2 - Dealing with unwanted conversation when there's no talknpc task
3 - Dealing with unwanted conversation when there's a talknpc task

farrainbow:
I need to review this more, but I can't say I like the latest changes.

IMO in general this task (and all other tasks) should be able to have several independent instances.

Regarding approach argument: it should modify distFromGoal of MapRoute task instead, otherwise things may fail on strange maps.

Regarding approaching the NPC we want to talk to: since we can't command to approach "non-existing" NPC and there are always (map)+x+y arguments, we can just check whether an NPC which starts talking with us is the NPC we're already have a task to talk with (which didn't start yet - that can be done with creating a hook which NPC tasks would listen to and set a return value if it's "their" NPC), and only create a new task if it's a different one (and we encountered it while trying to go to another NPC). That could solve some of remaining bugs.

Checking for all subtasks to see something isn't a thing we like to do. It smells bad and TalkNPC task can be used in other tasks as well, and we don't really want to go find all instances of it. Also consider a setup when there's a Task::Chained with two Task::TalkNPC inside of it, or more complicated setups. It can help if we move adding hooks to task's activate method, or maybe even later, if possible (maybe not with the current code).

Also I've managed to do all pre-talking stuff (routing, waiting for movement to stop, finding target, initializing steps) task-way, resulting changes look simple, so later I can add that.

Henrybk:
Oh, well, thanks. I didn't think the approaching code looked so bad, the other changes are a bit sly :P.

farrainbow:
Maybe it would be easier to not do any routing inside of `Task::TalkNPC` and let `talknpc` command just create a `Task::Chained` with `Task::MapRoute` and `Task::TalkNPC` in the chain (if needed). If we then move hook addition to the `activate()`, then only one instance of `Task::TalkNPC` would catch them, even if on a way to some NPC we encounter another one. For NPCs which need approaching that would work too with some tweaks (the specified conversation steps would be activated and used only after routing is done). What do you think about this?

farrainbow:
Example of what I mean:

---
**Task::Interaction** *(map, x, y, distance, task)* - route to the specified actor, find it, and do a *task*-based specified interaction with it

What it will do:
- make a `MapRoute` subtask
- listen to `actor_display` packet, compare with coordinates provided and save actor ID if we found the actor we need
- when routing is done, activate the provided in arguments *task* (that would be TalkNPC task for now, but we can make different ones later)
- listen to `npc_talk` packet, in case of "approaching NPC", and ONLY IF it's the same NPC ID as we've found previously, stop the routing subtask and activate the talking task the same way as in the previous point, but also provide it with the data of initial packet we received (just call some method); if NPC ID don't match, it should be ignored by this task and other things should be set up in such a way so new task would be created to handle the talking with that unwanted NPC (we can return values from hooks to do that easily)
---
If you like that and it handles all the cases we need, I can implement it.

The only problem I can see if we try to approach NPC which doesn't exist (so no info about its coordinates) but which still starts talking to us. Considering there may be various NPC like this, we can maybe add a *actorID* argument to specify ID of NPC we want to approach, if we really need to handle this case.

farrainbow:
[18:07] <allanon> if we use coordinates, we need to ignore hidden npcs
[18:07] <allanon> iRO stacks the event and non-event npcs on the same cell

farrainbow:
If needed: npc conversation step which would wait for NPC to start the talk first (for edge cases and lags with "approaching").

farrainbow:
I'll create another branch for TalkNPC rework itself, and another one for dealing with routing.

PR #370, opened 2016-12-20
Use one handler for all map_property packets

farrainbow:
Also move to Receive.pm and tweak unpack strings a bit.

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -1954,27 +1954,65 @@ sub actor_status_active {
 	}
 }
 
-#099B
-sub map_property3 {
+# 0199, 01D6, 02E7, 099B
+sub parse_map_property {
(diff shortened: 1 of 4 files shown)


itsrachelfish:
@farrainbow What's the best way to test this? I reviewed the code and it looks good. Do I need to go to a PVP map in order to make sure this is working?

farrainbow:
Enable related config options (status_mapType, status_mapProperty) and visit different maps? PVP would be good to check since I didn't really visited one with this.

kaliwanagan:
I think this is good to merge; the only change I will advise is to add a message log to output the map property. This way when people start using it and just in case there are problems, their screenshots or logfiles will contain usable information.

@farrainbow what do you think?

LastRoze (approved):

PR #373, opened 2016-12-22
Fix character creation

farrainbow:
Other servers which use new system where you don't select stats should tweak their serverTypes in the same way.

### DIFF
--- a/src/Network/Send.pm
+++ b/src/Network/Send.pm
@@ -408,6 +408,35 @@ sub sendCharLogin {
 	debug "Sent sendCharLogin\n", "sendPacket", 2;
 }
 
+# 0067, 0970
+sub parse_character_create {
+	my ($self, $args) = @_;
+
+	$args->{name} = bytesToString($args->{name});
+}
+
+sub reconstruct_character_create {
+	my ($self, $args) = @_;
+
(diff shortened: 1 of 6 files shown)


kaliwanagan:
@farrainbow does this mean changes need to propagate across all the servertypes? I wonder if we can instead have an optional switch in the config, where if the new system is enabled, then use this otherwise fallback to the old way of character creation?

moneyisgod:
Amazing. No words to describe. Who reads your parser?

alisonrag:
@sctnightcore now that i know we have 4 versions of creation 
1 - old and maybe the regular (bRO use this) when account have a fixed sex
slot, name, hair, hair_style, atributes
https://www.youtube.com/watch?v=yiDV_gzj8Jg

2 - same as regular but without attributes (used in some private servers and some time in past in some official server)
slot, name, hair, hair_style
https://www.youtube.com/watch?v=GJO1B9ogKXY

3 - used in account with 2 sex (iRO classic and restart use this)
slot, sex, hair, hair_style, attributes

4 - used in server with doram race (iRO Renewal and kRO use this)
slot, job_id, sex, hair, hair_style
https://www.youtube.com/watch?v=92p636rjV4A

alisonrag:
remake in #1982

PR #384, opened 2016-12-25
Allow using talknpc command with empty conversation sequence

farrainbow:
Useful for talking with NPCs whose sequence consists of only `c` (Next) items.


onepointfivem:
I've been using `talknpc x y c ` for this kind of situations. Don't know if there's problems with it.

lututui:
@onepointfivem It's fine to do that. This is specially useful if one has autoTalkCont activated.

PR #394, opened 2017-01-03
Add macro_defaultPriority

lututui:
Adds a new config option for the macro plugin, macro_defaultPriority. This allows the user to modify the default priority (previously hardcoded to zero/check first) for all macros that didn't specify a **priority** special condition in their automacros.

`macro_defaultPriority <number>`

### DIFF
--- a/plugins/macro/macro.pl
+++ b/plugins/macro/macro.pl
@@ -153,6 +153,11 @@ sub checkConfig {
 		warning "[macro] orphans: using method 'terminate'\n";
 		configModify('macro_orphans', 'terminate')
 	}
+	
+	if (!defined $::config{macro_defaultPriority} || $::config{macro_defaultPriority} !~ /^\d+$/) {
+		warning "[macro] default priority: using 0 (\"check first\")\n";
+		configModify('macro_defaultPriority', '0');
+	}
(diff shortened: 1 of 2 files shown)


farrainbow:
BTW does macro plugin allow to specify negative priority for a macro?

lututui:
@farrainbow yes, it does

plugins/macro/macro.pl
```diff
@@ -153,6 +153,11 @@ sub checkConfig {
 		warning "[macro] orphans: using method 'terminate'\n";
 		configModify('macro_orphans', 'terminate')
 	}
+	
+	if (!defined $::config{macro_defaultPriority} || $::config{macro_defaultPriority} !~ /^\d+$/) {
+		warning "[macro] default priority: using 0 (\"check first\")\n";
+		configModify('macro_defaultPriority', '0');
+	}
```
farrainbow: I think for most stuff we just use (and should use) the default value without warnings and modifying `%config` and config file.

farrainbow:
You kind of can get the similar effect now if you assign both positive and negative priorities, with `0` being the default one, so that feature looks superfluous for me.

PR #395, opened 2017-01-03
[2017-01-03] Connection Files for bRO

lututui:
http://openkorebrasil.org/index.php?/topic/2078-atualiza%C3%A7%C3%A3o-03-01-2017/


flacrasec:
tested and working thank you!

PR #424, opened 2017-02-02
Fixed encoding for fRO tables

Orangensaft:
The changed files were ISO-8859-15 encoded. Openkore fails to load these, as UTF8 is required.


itsrachelfish:
Thanks for catching this! :+1:

PR #425, opened 2017-02-02
poseidon.pl small changes

lututui:
Right now, poseidon.pl cannot reach the precompiled dll for Windows (XSTools.dll) in the trunk folder, therefore the user has to move it to src/, src/Poseidon or src/deps/.

Retrieve address from object. Useful when setting the port to zero in poseidon.txt

### DIFF
--- a/src/Poseidon/poseidon.pl
+++ b/src/Poseidon/poseidon.pl
@@ -42,9 +42,9 @@ sub initialize
 	# Starting Poseidon
 	print "Starting Poseidon 2.1 (26 Sep 2012)...\n";
 	$roServer = new Poseidon::RagnarokServer($config{ragnarokserver_port}, $config{ragnarokserver_ip});
-	print "Ragnarok Online Server Ready At : " . $config{ragnarokserver_ip} . ":" . $config{ragnarokserver_port} . "\n";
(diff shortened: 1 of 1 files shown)


flacrasec:
so this solves this issue? we can use XSTooks.dcc in the trunk folder?

lututui:
@flacrasec yup

itsrachelfish:
@lututui Can you provide some more information on how this can be tested and what the results should be before and after these updates? I'm assuming kore / poesidon fails to run and outputs an error message unable to load the DLL? I've never used poesidon so I'm probably not a good candidate for testing this.

@flacrasec Can you take some screenshots of what happens currently vs what happens after this code is merged? Thanks :+1:

lututui:
@itsrachelfish 

First, the XSTools.dll issue. Notice that the trunk folder is not included in @INC, where XSTools.dll is located by default:
image

Also, if one sets ragnarokserver_port or queryserver_port (in poseidon.txt) to zero (request to dynamically generate an unused port number), the poseidon.pl console will say the port is zero, instead of showing the actual generated port.

image

flacrasec:
it is like lututui posted, sorry I was with many things to do that is why I didn't posted.

itsrachelfish:
Cool, looks good! :+1:

PR #437, opened 2017-02-16
update iRO files

allanon:
- [x] QA Review

This pull request was auto-generated by patch_iro.pl: https://gist.github.com/allanon/38620af49d250005acbf


itsrachelfish:
Wow @allanon that's a LOT of changes. Did the iRO GMs recently restructure their tables files? Also, did you notice all of the ```(null)``` items?

allanon:
Yes, they updated their client past the 2012 (!) client code for item tables. In the new client, all item data comes from a single file called `System/itemInfo.lub` (from the `rgz` update files), which as you might guess is a compiled Lua binary. I'm using some trickery to decompile the binary into the various files which we all know and love.

The primary visible consequence is that the files are now in numeric order and all comments are gone.

The new `patch.pl` also properly handles some Code Page 949 (Korean) characters that weren't being handled right before, so you'll see a few curly quotes and ellipses in various spots.

More info:
* The [`rgz` format](https://github.com/HerculesWS/Hercules/wiki/RGZ)
* Some [high level info about itemInfo.lub`](http://herc.ws/wiki/Custom_Items#Defining_Items_clientside_.28For_New_Clients.29).
* The [SeparateItemInfo.lua](https://github.com/ROClientSide/Translation/tree/master/Dev/Tools/SeperateItemInfo) script used to parse `itemInfo.lub`.

PR #439, opened 2017-02-17
iRO doram update packet changes

allanon:
- [x] QA (does it work?)

The number of packet changes seems to be quite small this time. If anybody sees any other missing packets, please report them in IRC and/or as a comment here.


allanon:
Tested it myself... if this breaks something we'll roll forward. :)

PR #447, opened 2017-02-19
Use git commit hash instead of "SVN Revision"

ChrstphrR:
Found via reading user posted issues on Github, where the errors.txt
reports that were still showing a misdirecting/distracting
"SVN revision: unknown".

We let this one slip through - we're still checking the SVN revision
when we should be checking the git commit hash instead.  The code,
getGitRevision, already exists, so let's get rid of the old, onto the
new!

Modified:
src/ErrorHandler.pm
- changed out SVN references/calls for Git in the errors.txt output
src/Interface/Console/Curses.pm
- swapped out same SVN for Git references calls, minor format change
so we're not prepending an "r" on the git commit hash.
src/Settings.pm
- removed getSVNRevision - no longer called in the two places above.

### DIFF
--- a/src/ErrorHandler.pm
+++ b/src/ErrorHandler.pm
@@ -75,11 +75,11 @@ sub errorHandler {
 	$log .= "\@ai_seq = @Globals::ai_seq\n" if (@Globals::ai_seq);
 	$log .= "Network state = $Globals::conState\n" if (defined $Globals::conState);
 	$log .= "Network handler = " . Scalar::Util::blessed($Globals::net) . "\n" if ($Globals::net);
-	my $revision = defined(&Settings::getSVNRevision) ? Settings::getSVNRevision() : undef;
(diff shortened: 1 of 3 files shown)


itsrachelfish:
@ChrstphrR How does this differ from PR #448 ?

ChrstphrR:
This one was committed first, but allanon's commit in PR 448 went through the extra step of using a function that tests for the date of the files, in lieu of no git commit hash being found.   This was the tipping point, because users that download the source from a zip archive on github will still show an "unknown" on this branch, instead of a date.

Also, his set of commits removed the SVN version check from a plugin that was the 3rd place actually checked the SVN revision.

So, PR #448 has the advantage, and just needed extra commits to remove the getSVNRevision routine, once it was no longer called elsewhere.

tl;dr:
We can formally reject this branch, since 448 covers everything this one did, and and that little bit more.

PR #448, opened 2017-02-19
reformat revision to include both git and file creation date

allanon:
- [x] Product Review (is it the right feature to add?)
- [x] QA Review (does it work?)
- [x] Code Review (does the code look good?)

We were not including the git sha (if available) in `errors.txt`. This pull fixes that.

In addition, as the git sha isn't always available (many users use "Download ZIP" instead of a git client to get the code), this pull adds the creation date of `Settings.pm` to the version string. Github sets the creation date of all files in downloaded zipfiles to (near?) the date of the last commit.

Before: `SVN revision: unknown`

After: `Revision: git:9ddeda4_2017-02-19 ctime:2017_02_19`

### DIFF
--- a/plugins/needs-review/leaveMeAlone/trunk/leaveMeAlone.pl
+++ b/plugins/needs-review/leaveMeAlone/trunk/leaveMeAlone.pl
@@ -144,11 +144,6 @@ sub turnOn {
 	
 
 sub start {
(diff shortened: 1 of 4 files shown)


ChrstphrR:
Tested  over (used "die" in src/functions.pl as per allanon's suggestion to test the errors.txt output

PR #451, opened 2017-02-20
Translation Refresh

lututui:
Run src/po/update.sh, it's been three years!


itsrachelfish:
@lututui I'm a bit confused by this. Is this an automatic script that uses machine translations?

lututui:
src/po/update.sh is a script designed to update/create and compile Openkore's multi-lingual [gettext](https://www.gnu.org/software/gettext/) messages.

https://github.com/OpenKore/openkore/blob/master/src/po/update.sh
"# This script extracts strings from the OpenKore source code,
# updates openkore.pot and *.po, and compiles *.po to .mo."

It was last executed in 2014. Since then, every new commit adding messages using the [Translation](https://github.com/OpenKore/openkore/blob/master/src/Translation.pm) module could not be translated, since their entries were never added to openkore.pot. Also, if messages once translated had their original message moved around in source (remarkably due to code cleanup), the translation wouldn't work properly.

update.sh was created to update the base template file (openkore.pot), the human-readable translations (.po) and the machine-readable objects (.mo)

_<s>I hope I made sense?</s>_

itsrachelfish:
Haha yes that makes sense, thank you :+1:

PR #456, opened 2017-02-26
Fix summoner unable to sit.

chinyong-lim:
Summoner having different basic skill than novice/super novice.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2024,7 +2024,7 @@ sub processFollow {
 	}
 	if($config{'sitAuto_follow'} && (percent_hp($char) < $config{'sitAuto_hp_lower'} || percent_sp($char) < $config{'sitAuto_sp_lower'}) && $field->isCity) {
 	my $action = AI::action;
-		if($action eq "sitting" && !$char->{sitting} && $char->{skills}{NV_BASIC}{lv} >= 3){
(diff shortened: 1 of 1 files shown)


src/AI/CoreLogic.pm
```diff
@@ -2024,7 +2024,7 @@ sub processFollow {
 	}
 	if($config{'sitAuto_follow'} && (percent_hp($char) < $config{'sitAuto_hp_lower'} || percent_sp($char) < $config{'sitAuto_sp_lower'}) && $field->isCity) {
 	my $action = AI::action;
-		if($action eq "sitting" && !$char->{sitting} && $char->{skills}{NV_BASIC}{lv} >= 3){
+		if($action eq "sitting" && !$char->{sitting} && ($char->{skills}{NV_BASIC}{lv} >= 3 || $char->{skills}{SU_BASIC_SKILL}{lv} == 1)){
```
allanon: Should this be `>=` instead of `==`?

turnipgit: it is the same. because the Summoner only have 1 basic skill.

lututui:
I think this is good to merge, since SU_BASIC_SKILL only has one level.

https://github.com/HerculesWS/Hercules/commit/473ba564829a6d0153e320d1f302754442648ec7#diff-08420bab8271f3965304fee42cf4b865R11674

PR #472, opened 2017-03-05
notify users when they lose skills (typically due to equipment changes)

allanon:
- [ ] QA (does it work?)
- [ ] Product Review (is this a good idea?)

Show a notice when the bot loses a skill. This usually happens when unequipping something which confers a skill (eg, unequipping a Vitata Card removes the Heal skill).


flacrasec:
I like the idea.

PR #476, opened 2017-03-07
Fix allow ticket

Henrybk:
For as long as I can remember kore can't handle properly Free Ticket for Kafra Transportation, these days are over because after days I finally found the bug that made it not work.


allanon:
Good job tracking this down!

flacrasec:
Thank you Henry for the good job! =)

and2and2:
Afaik, iro give restriction for kafra ticket
Only first class or low level , cmiiw
Already accommodate?

Henrybk:
@and2and2 Well, I guess there's nothing we can do about it other than remove allow_ticket from iro portals.txt. Or maybe add a configuration to xConf to conf portals configuration and then redo portals_los.

PR #484, opened 2017-03-11
fix cart inventory; should clear when logging in

allanon:
This pull request undoes more of the timing changes introduced by the InventoryList refactor. The cart inventory is supposed to be cleared when we log in (this pull request), and cart existence is supposed to be set when we get a `map_changed` packet (a different pull request, already merged).

Without this code, the cart can get into a bad state, with phantom items in the cart which can cause disconnects when the user tries to do something with them. And relogging won't fix it.


flacrasec:
I was having a issue when I opened my merchant shop sometimes it just do not start.

PR #487, opened 2017-03-12
[macroinclude.pl] Rewrited it totaly. The same but better.

Mortimal:


kaliwanagan:
This looks really interesting, but I'd like @farrainbow or @Henrybk or @lututui to also check this out :)

@SabadyshynV what were the shortcomings of the original code that you were addressing in this PR?

Henrybk:
@SabadyshynV eventMacro already has include command support (although not documented in the wiki)

Nipodemos:
i think it is a good thing to merge, even if is less used

lututui:
This must drop eventMacro support before merging

Nipodemos:
i've pushed a commit that removes eventMacro support
anyone have anything against merging this?

PR #502, opened 2017-03-18
[profiles.pl] Adeed feature to "hide" profile

SabadyshynV:
Profiles with names starts with # won't display in list...
Just if u need to "comment" profile or leave some macro folder for different profiles **!include ../#macroses/Misc.txt**


kaliwanagan:
@SabadyshynV I'm not sure if changing the filename like this is the right answer. Won't it be better to just comment out the whole line?

Henrybk:
I think the biggest feature we are lacking in profiles is the ability to choose not to use any profile, just plain control folder.

PR #541, opened 2017-04-08
Fix utf-8 error for idRO tables

hellowin:
This PR fix those errors
image
image


hellowin:
@windhamwong I saw a lot of spammer from Indonesia spamming openkore issues since idRO got new licensee by this year and current standard OK is not working anymore, I can help you to at least filtering it out like marked with some label or anything if you guys allows me to reduce the number of unnecessary issues.

PR #544, opened 2017-04-09
Placing new flds to correct folders

windhamwong:


allanon:
Could you please update https://github.com/OpenKore/openkore/blob/master/tables/iRO/resnametable.txt, and any other `resnametable.txt` files which would be affected by this?

kaliwanagan:
@windhamwong ping :) organizing the files is always something we wanna happen (as long as it doesn't break anything)

windhamwong:
I am closing this until I officially merging twRO and fixing all flds

PR #546, opened 2017-04-11
[iRO] RODEX Mail System Add Request

Zelecktor:
Hello People.
As the mail system was removed on November 22, 2016. I just wandering to add a plugin or a new function to use the new mail system called "RODEX" to simplify the life without entering in my bot account and then risking ban in getting items directly.

Actually there is no commands to use it, but what i see is very easy to use (almost the same that old mail) and more confortable because you can use in any map, at any time you like.


itsrachelfish:
Please recreate this pull request without including all of the commits from master.

The only commits in this pull request should be the code you've added.

PR #550, opened 2017-04-13
[pm command] All numbers character name fix.

Mortimal:
If char name consists of only numbers. pm command would count it as list number.
So now we can "" the number so to send it to all numbers character name:
pm "12312" message


kaliwanagan:
@SabadyshynV I know it's been over a few months, but are you still interested in merging your code?

lututui:
This doesn't work. $user doesn't get quotes back from parseArgs($args, 2);

PR #580, opened 2017-05-12
support achievement packets

allanon:
Minimal support for achievement related packets.

* `actor_info` has an additional field, likely an offset into a list of achievement-related titles. The new data is ignored.
* `quest_all_list3` is a modified version of `quest_all_list2`, with 12 new bytes for each kill count. The new data is ignored.
* `achievement_list` is a new packet. It is ignored.


allanon:
Merged to get openkore working again for iRO. Additional work on the new packets will be a different pull request.

PR #633, opened 2017-06-25
[plugin/webMonitor] Show current player on sidebar's map

BubonicPestilence:
Preview: https://yadi.sk/i/WjiWckwF3KRqdf

### DIFF
--- a/plugins/needs-review/webMonitor/trunk/WWW/_sidebar.html.template
+++ b/plugins/needs-review/webMonitor/trunk/WWW/_sidebar.html.template
@@ -2,18 +2,111 @@
 	<h5 class="center">
 		<i class="icon-map-marker"></i> <span class="value_field_description">{characterLocationDescription}</span>
 	</h5>
+    
 	<div class="center">
 		<!-- why canvas? -->
-		<canvas id="map" style="background-image: url('{characterLocationMapURL}')">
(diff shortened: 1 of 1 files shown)


plugins/needs-review/webMonitor/trunk/WWW/_sidebar.html.template
```diff
@@ -2,18 +2,111 @@
...
+                var newWH = {
+                    width: "",
+                    height: "",
+                };
+                
+                if ($map.css("width") != "512px") {
+                    newWH = {
+                       width: "512px", 
+                       height: "512px", 
+                    };
```
BubonicPestilence: Yeah, currently i stuck with conversion of original coorditaions to map files. Annyn suggestions?

P.S. My local version works with party as well :)

kaliwanagan:
Merging this in anyway; we can just fix the width and height issue later.

plugins/needs-review/webMonitor/trunk/WWW/_sidebar.html.template
alisonrag: in interface wx we have the conversion of original to map, we need just search for the code

BubonicPestilence:
I'll do it this week, if none else do it faster, just busy atm

PR #673, opened 2017-07-10
Make Task::TalkNPC errors better to read and add more debug messages

Henrybk:
Title


kaliwanagan:
I'm approving this for merge, but one refactor you might optionally do is to save the text into a local var (like maybe $msg) and use that one, instead of repeating all the text multiple times :)

PR #681, opened 2017-07-11
add default timeouts to Task::TalkNPC

Henrybk:
This is already done this way for $config{npcTimeResponse} to prevent configuration errors, and seeing the great number of people forgeting to update their timeouts.txt I guess this is a good idea.

### DIFF
--- a/src/Task/TalkNPC.pm
+++ b/src/Task/TalkNPC.pm
@@ -230,6 +230,9 @@ sub iterate {
 	$self->SUPER::iterate(); # Do not forget to call this!
 	return unless ($net->getState() == Network::IN_GAME);
 	my $timeResponse = ($config{npcTimeResponse} >= 5) ? $config{npcTimeResponse}:5;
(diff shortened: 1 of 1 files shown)


hendra814:
Getting error after get this update, talk to kafra always get disconect

Henrybk:
@kaliwanagan How do you suggest we do that?

kaliwanagan:
@Henrybk 

In the config parser you can check for particular options and if they're not set, just set them.

I think this happens in https://github.com/OpenKore/openkore/blob/master/src/Settings.pm#L767 so maybe we can add a new function called `setDefaultValues` near the end, where it will have a list of default values if the config option is not set.

That way all we need to do is look at the settings if we need anything changed.

Alternatively that function can load a file called defaults.txt which will then contain default values.

I don't know if we should do this in this PR or in a separate PR; up to you. I'm pre-approving this PR so if you want to merge it now, please do so :) But if you prefer to do the refactoring of the defaults (since these would probably be the first entries in that new way of setting defaults) then that would work too.

Issue #270, opened 2016-10-16
!include NonExistantFile.txt causes error message without a filename mentioned at all.


ChrstphrR:
**Issue:**
Using !include within config.txt (and possibly in other files that are allowed to use !include) will cause OpenKore to exit graacefully, but with a useless message stating:

> Unable to load the file .

**Symptoms:**
**_Steps to recreate:**_
- Insert an include directive in control/config.txt that calls a file/path that does not exist.
  e.g.: !include foo.txt
- Save said changes, and launch openkore using that config.txt

_Nota Bene:_ If you are using the Console::Curses interface like I was, you will not catch this error unless you are logging your console using _logConsole 1_ or change the interface to plain ol' Console.

What I saw in my case:

> Use of the encoding pragma is deprecated at plugins/macro/Macro/Parser.pm line 5.
> **\* OpenKore what-will-become-2.1 ( version bc07ae7_2016-10-14 ) - Custom Ragnarok Online client ***
> **\*   http://www.openkore.com/   ***
> 
> Selectively loading plugins...
> Loading plugin plugins/breakTime/breakTime.pl...
> Loading plugin plugins/macro/macro.pl...
> 
> Loading control/config.txt...
> Unable to load the file .
> Benchmark results saved to benchmark-results.txt
> macro plugin unloading, cleaning up
> **\* OpenKore what-will-become-2.1 ( version bc07ae7_2016-10-14 ) - Custom Ragnarok Online client ***
> **\*   http://www.openkore.com/   ***

Okay, okay, I fibbed a little - I've sanitized the path to config.txt so I don't show the world my settings or account info.  But this is otherwise exactly what I saw to cause this issue.

**Info**
**_OS:**_

> Ubuntu 16.04.1 LTS

Using the server distro, if that matters.
Kernel issue from "uname"

> 4.4.0-42-generic GNU/Linux

**_Perl:**_

> > perl --version
> 
> This is perl 5, version 22, subversion 1 (v5.22.1) built for x86_64-linux-gnu-thread-multi
> (with 58 registered patches, see perl -V for more detail)
> 
> Copyright 1987-2015, Larry Wall

**_Openkore:**_

> **\* OpenKore what-will-become-2.1 ( version bc07ae7_2016-10-14 ) - Custom Ragnarok Online client

**Cause:** 

[...]

PR #692, opened 2017-07-14
provide filename so !include errors in config.txt will report it

allanon:
- [ ] Code Review

Before: `Unable to load the file .`
After: `Unable to load the file profiles/abc/included.txt.`


lututui:
Fixes #270

PR #693, opened 2017-07-14
enable map plugin by default

allanon:
- [ ] Code Review

Map plugin output with the `Console::Curses` interface (this is automatic and refreshes periodically):
```
      ####  p            $               ####
#     #### p$   $##  $      $  $ $   ##  ####
      #### p$    ##$    $           $##  ####
      #### p$    $         p$            ####
      #### p$
  ############$           $    $          ####
  ############$    $  $    p              ####
##############                 $    $     ####
##############                            ####
##############$    $                      ####
  ############$    $          $         ######
  ############        p           $       ####
   #######            p$                  ####
   #######               $   $          ######
   ######### n         @                  ####
   #######             p    $p            ####
   #######       p    p   $   p   $       ####
   #########     ##           $   $  ##   ####
   #######$   $  ##p      $          ##   ####
          $   p    p  $   $ p    $         #
          $$   p      p      $
  ##########          $   $    $     $
  ##########$     n  p$         $         ####
  ##########$         p    pp             ####
  ##########          $      $  $  $      ####
   ########  $        p  $                 ###
   ########    $  $   $   $                ###
   ############p p$   p   p         $    #####
 ############ $p  $   $      $          ######
 ############                           ######
```

It also supports manual map requests.

### DIFF
--- a/control/sys.txt
+++ b/control/sys.txt
@@ -49,7 +49,7 @@ loadPlugins 2
 # loadPlugins_list <list>
 #   if loadPlugins is set to 2, this comma-separated list of plugin names (filename without the extension)
 #   specifies which plugin files to load at startup or when the "plugin load all" command is used.
-loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill
+loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill,map
 
 # skipPlugins_list <list>
 #   if loadPlugins is set to 3, this comma-separated list of plugin names (filename without the extension)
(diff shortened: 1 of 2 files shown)


kaliwanagan:
Is it a good idea to enable it by default? won't it use up all of the terminal space (esp since  many people don't really run kore full screen -- or do they?)

allanon:
In the regular console, the map is on-demand only: users have to request that a map be printed.

The map only shows up automatically  in the `Console::Curses` interface, and in that interface the right hand side of the screen lists actors (players, NPCs,, monsters, items). The map shows up below that list, in an area that would normally be empty almost all of the time. For example, look at the lower right corner of this screen capture: https://gist.github.com/allanon/5e9d196630df793edbe460938d26ee7f

Issue #669, opened 2017-07-10
Automatic DC when talking to an NPC


FireflySkye:
[iRO server]
The recent update makes my bot dc when it talk to the NPC. 

config.txt
npcTimeResponse 100 (set it to 1 still the same)
npcWrongStepsMethod 0

timeout.txt
ai_route_calcRoute 1
ai_route_npcTalk 100
ai_npcTalk 100

ugh..

![23](https://user-images.githubusercontent.com/19757768/28026486-2e5b079a-65d1-11e7-8c0c-234caf03af3d.png)

Henrybk:
Are you using a profiles folder? They are not updated from github, if so please add these to your timeouts.txt

```
# These timeouts are used in npc conversation (Task::TalkNPC)
ai_npc_talk_wait_to_answer 1.5
ai_npc_talk_wait_after_close_to_cancel 0.5
ai_npc_talk_wait_after_cancel_to_destroy 0.5

```

FireflySkye:
Didn't dc but new error
![234](https://user-images.githubusercontent.com/19757768/28031864-99da674c-65e3-11e7-8edd-d9b69d467bd7.png)

Henrybk:
What console command are you using to talk to this npc? and which npc is it?

FireflySkye:
move [insertmap] or when it autofollow. any NPC that can teleport. Gramps, Eden Teleporter etc.

Henrybk:
Sorry for the inconvenience, many updates were made on openkore in the last day regarding this problem, would you please update your openkore to the last version, set `logConsole 1` and `debug 2` in `config.txt` and reproduce the error and after give me (post here) the `console.txt` generated in your `logs` folder.

This way I'll be able to take a closer look at the problem.

ShishieYo:
up for this, wthe new update keep disconnecting us when talking to npcs.

kaliwanagan:
```
Sorry for the inconvenience, many updates were made on openkore in the last day regarding this problem, would you please update your openkore to the last version, set logConsole 1 and debug 2 in config.txt and reproduce the error and after give me (post here) the console.txt generated in your logs folder.

This way I'll be able to take a closer look at the problem.
```
Did you do this?

ShishieYo:
[console.txt](https://github.com/OpenKore/openkore/files/1140921/console.txt)

angelwell1:
Hi sir.
I have same problem in lastest version. Hope you fix it, love you :x.

[...]

PR #704, opened 2017-07-17
Move wait and command steps check to before autotalkcont

Henrybk:
This fixes the problem @hendra814 was having in #669


Henrybk:
Oh, I will fix this right away

Henrybk:
@kaliwanagan Fixed

Issue #705, opened 2017-07-17
IRO: Getting stuck and unknown packet


Xianzz:
Server: IRO
Good day, getting a unknown packet 0AB8 and sometimes would suddenly stop moving and says can't reach the target in a map with little to no obstructions

Xianzz:
Been running with the new recvpackets, so far seems all good. Thanks

PR #706, opened 2017-07-17
add packet 0AB8 to iRO recvpackets

Henrybk:
This hopefully fixes the problems some people including me are having with dead characters not being recognized.

http://forums.openkore.com/viewtopic.php?f=16&t=212031

http://forums.openkore.com/viewtopic.php?f=16&t=212079

Also #705


sctnightcore:
<3 ty sir

PR #729, opened 2017-07-20
#RO:RE old moc_fild

sctnightcore:


xTriumph:
I think its better to add this to the Restart tables folder than the portals.txt generic.
But first we need to fix the wrong path for the folder in the servers.txt.

kaliwanagan:
@xTriumph thanks, the servers.txt has been fixed in #776 

@sctnightcore once that pr is merged, can you redo the portals.txt changes you're doing inside the iro restart portals.txt instead?

kaliwanagan:
@sctnightcore ping?

sctnightcore:
No merged. @kaliwanagan  
i will fix in tables RE:START 
ty

PR #742, opened 2017-07-24
use the public ip if set

kaliwanagan:
this allows kore running inside a natted environment to expose the correct public ip instead of the currently associated ip (this is the same behavior in XKore 2 ported over to XKore proxy)

also, my editor auto removes spaces at the end of the line :D


makaroo:
sir please help again... im connected now... but my problem is..! im connect in openkore but... i cant see my character in game.. invisible character . cant see in real map ,, ty for help

PR #752, opened 2017-07-25
Re:Start Fixed ayo_fild02 access

Onihen:


kaliwanagan:
Can anyone else verify this?

sctnightcore:
wait ..

xTriumph:
This fixes the wrong use of the tables/portals.txt instead of the Restart/official/portals.txt.
But you need to change the code from iRO/Restart;iRO to iRO/Restart/official;iRO
I tested it with the Link Christmas NPC and it works.
Im looking now if it fixes the ayo_fild02 acess.

kaliwanagan:
will be merged via #776 

thanks @Onihen !!

Onihen:
My bad. I moved portals.txt out of the official folder because I thought that folder was for unmodified tables extracted from the client (based on what was discussed here #25)

PR #754, opened 2017-07-25
Give RMS its own ServerType

lututui:
RMS now requires encryption to login (cryptKeys).

We could change it directly in kRO_RagexeRE_2014_10_22b, but that might break other servers using this ServerType


sctnightcore:
RMS is ? idk

lututui:
RMS stands for RateMyServer. They have two test servers (Renewal and Pre). Openkore has theses servers on servers.txt by default.

sctnightcore:
@lututui Thank you for answer

PR #375, opened 2016-12-23
Item renames

Henrybk:
This tries to achieve what was said in #350

Issue #746, opened 2017-07-24
Autostorage broken


k1nt4r0:
@Henrybk sir, i download the openkore today 24 july 2017
and when my char talk to npc to change map on training ground it got error and dc

> This program has encountered an unexpected problem. This is probably because
> of a recent server update, a bug in this program, or in one of the plugins.
> We apologize for this problem. You may get support from IRC or the forums.
> 
> A detailed error report has been saved to errors.txt. Before posting a bug
> report, please try out the SVN version first. If you are already using the SVN
> version, search the forums first to see if your problem had already been
> solved,
> or has already been reported. If you truly believe you have encountered a bug
> in
> the program, please include the contents of the errors.txt in your bug report,
> or we may not be able to help you!
> 
> The error message is:
> Assertion (Argument head should be defined for packet actor_look_at) failed!
>  at E:/Games/Ragnarok/BOT/Bot ReRO/src/deps/Carp/Assert.pm line 271
>         Carp::Assert::assert('', 'Argument head should be defined for packet
> actor_look_at') called at src/Network/PacketParser.pm line 192
> 
> Network::PacketParser::reconstruct('Network::Send::iRO=HASH(0x66198ec)',
> 'HASH(0x6871604)') called at src/Network/Send.pm line 600
>         Network::Send::sendLook('Network::Send::iRO=HASH(0x66198ec)', 4, undef)
> called at src/AI/CoreLogic.pm line 372
>         AI::CoreLogic::processLook() called at src/AI/CoreLogic.pm line 71
>         AI::CoreLogic::iterate() called at src/functions.pl line 769
>         main::mainLoop_initialized() called at src/functions.pl line 75
>         main::mainLoop() called at src/Interface.pm line 75
>         Interface::mainLoop('Interface::Console::Win32=HASH(0x494824c)') called
> at openkore.pl line 97
>         main::__start() called at start.pl line 136
> Press ENTER to exit this program.
> 
> how to fix this?

k1nt4r0:
@Henrybk i got fix by ur update btw https://github.com/OpenKore/openkore/commit/b56dda3a62ece0a308ddfdbba1ae1cd49e60d202

but now the openkore cant sell item

> [console] Ready to start selling items
> [ai_npcTalk] Done talking with NPC Tool Dealer (1).
> [console] Sell failed.
> [console] Sell failed.
> [console] Sell failed.
> [console] Sell failed.
> [console] Sell failed.
> [console] Sell failed.
[...]

PR #758, opened 2017-07-25
Changes in operators and servertypes for #375

Henrybk:


flacrasec:
já? seu lindo!(already? you handsome!)

Henrybk:
This hopefully fixes autostorage problems in #746

PR #767, opened 2017-07-26
Update mail read setting on reading

Henrybk:


sctnightcore:
wait

kaliwanagan:
Why is ro depending on the client to set the read flag :(

PR #768, opened 2017-07-26
add hook on processAutoAttack cleanMonsters

Henrybk:
This is useful because it is where 'new' monsters are chosen to be attacked when kore is not already attacking or being attacked. I am using it to avoid attacking monsters which can be assisted by others of its own nameID (namely Wolves), to do so I am using my plugin avoidAssisters (https://github.com/Henrybk/Plugins/blob/master/avoidAssisters/avoidAssisters.pl)


moneyisgod:
Great.

PR #769, opened 2017-07-26
Possible fix for stackables as ammunition

Henrybk:


xSorux:
Will test it as soon as it's available for pulling!
#746 should be closed in this case.

xSorux:
Made the changes manually and did some more testing. Issue is resolved.
Previous changes (?) also introduced problems/inconsistency in arrow swapping, which have been resolved with this as well. Thank you!

PR #780, opened 2017-07-27
Changes to routing to add the possibility of rerouting and custom dist maps

Henrybk:


flacrasec:
explain these changes @Henrybk please.

kaliwanagan:
It can be. It's a generic way for plugins to modify kore's routing behavior.

PR #789, opened 2017-07-28
Update in function and packets of ZC_EQUIPWIN_MICROSCOPE (show_eq)

alisonrag:
Update to support all know show_eq packets

sem titulo

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -1427,44 +1427,57 @@ sub sage_autospell {
 
 sub show_eq {
 	my ($self, $args) = @_;
-
-	my $jump = 26;
-
-	my $unpack_string  = "v ";
-	   $unpack_string .= "v C2 v v C2 ";
-	   $unpack_string .= "a8 ";
-	   $unpack_string .= "a6"; #unimplemented in eA atm
-
-	if (exists $args->{robe}) {  # check packet version
-		$unpack_string .= "v "; # ??
-		$jump += 2;
+	my $item_info;
+	my @item;
+	
+	if ($args->{switch} eq '02D7') {  # PACKETVER DEFAULT	
+		$item_info = {
+			len => 26,
+			types => 'a2 v C2 v2 C2 a8 l v',
+			keys => [qw(ID nameID type identified type_equip equipped broken upgrade cards expire bindOnEquipType)],
+		};
+		
+		if (exists $args->{robe}) {  # PACKETVER >= 20100629
+			$item_info->{type} .= 'v';
+			$item_info->{len} += 2;
+		}
+		
+	} elsif ($args->{switch} eq '0859') { # PACKETVER >= 20101124	
+		$item_info = {
+			len => 28,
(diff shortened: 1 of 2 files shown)


src/Network/Receive.pm
```diff
@@ -1427,44 +1427,57 @@ sub sage_autospell {
...
-
-		$item->{cards} = $cards;
-
-		$item->{expire} = $expire;
-
+	for (my $i = 0; $i < length($args->{equips_info}); $i += $item_info->{len}) {
+		my $item;		
+		@{$item}{@{$item_info->{keys}}} = unpack($item_info->{types}, substr($args->{equips_info}, $i, $item_info->{len}));			
+		$item->{broken} = 0;
+		$item->{idenfitied} = 1;
```
allanon: Misspelled. Should be `identified`.

src/Network/Receive.pm
```diff
@@ -1427,44 +1427,57 @@ sub sage_autospell {
...
-
-	if (exists $args->{robe}) {  # check packet version
-		$unpack_string .= "v "; # ??
-		$jump += 2;
+	my $item_info;
+	my @item;
+	
+	if ($args->{switch} eq '02D7') {  # PACKETVER DEFAULT	
+		$item_info = {
+			len => 26,
```
allanon: No need to hardcode these data lengths. Just do `$item_info->{len} = length pack $item_info->{types}` in the while loop below.

src/Network/Receive.pm
```diff
@@ -1427,44 +1427,57 @@ sub sage_autospell {
...
-		$item->{type_equip} = $type_equip;
-		$item->{equipped} = $equipped;
-		$item->{broken} = $broken;
-		$item->{upgrade} = $upgrade;
-
-		$item->{cards} = $cards;
-
-		$item->{expire} = $expire;
-
+	for (my $i = 0; $i < length($args->{equips_info}); $i += $item_info->{len}) {
```
allanon: Is the equipment in a useful order? The code for `character_equip` attempts to re-order the equipment so that it can be more easily read by a human. Maybe we could use that code to sort these results, if they're not already in a nice order.

alisonrag:
@allanon 
What about packet 0906?
> implemented

Can we delete the old character_equip packet implementation?
> we can just comment the code for future reference

src/Network/Receive.pm
alisonrag: is it really necessary pack the types just to get their length?
talking about performance what is better hardcoded or pack and get length?

src/Network/Receive.pm
alisonrag: Unfortunately not =/
we can try to re-order, but i dont have any ideia who he `character_equip` do this.
can you help me with this?

src/Network/Receive.pm
kaliwanagan: Many implementations of the other servertypes will often override the packing string (because the server sends different stuff). This way, the solution is more generic and will support all server types.

PR #790, opened 2017-07-29
Beta change to make kore notice missing portals

Henrybk:
This pull request is being made to create a discussion regarding the possible solutions for this old known problem.

Ilustration of the problem: https://i.gyazo.com/e157f69c4dda6146016db01d71c1eba6.jpg

Definition: Openkore now has no way of dealing with a missing portal and will just stay on top of the portal location in portals.txt sending loads of move packets to the server and never moving a muscle.

I showed 2 possible solutions for the problem in the files changed, first is to just delete the portal and re-route, as we do with missing npcs, and second is to try to search for near portals and route to them hoping they are the portal we actually want to reach.

### DIFF
--- a/src/Task/MapRoute.pm
+++ b/src/Task/MapRoute.pm
@@ -155,6 +155,8 @@ sub iterate {
 		delete $self->{substage};
 		delete $self->{timeout};
 		delete $self->{mapChanged};
+		delete $self->{missing_portal};
+		delete $self->{guess_portal};
 		shift @{$self->{mapSolution}};
 
 	} elsif ( $self->{mapSolution}[0]{steps} ) {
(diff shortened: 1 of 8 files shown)


flacrasec:
this is a great addition to preven unecessary bans from our bots being trapped, but to delete the portal is a little risk because sometimes the portal exist but the server is just experiencing some bug.

c4c1n6kr3m1:
good
deleting portals is not overwrite portals.txt

xTriumph:
Oh this is great. The GM's are turning off portals to catch some bots. I've seen this in payon and morroc. Good job!

_Sent from my Motorola Moto G (4) using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

src/Task/MapRoute.pm
```diff
@@ -271,8 +271,25 @@ sub iterate {
 
 	} elsif ( $portals_lut{"$self->{mapSolution}[0]{map} $self->{mapSolution}[0]{pos}{x} $self->{mapSolution}[0]{pos}{y}"}{source} ) {
 		# This is a portal solution
+		
+		
+		if ( distance($self->{actor}{pos_to}, $self->{mapSolution}[0]{pos}) == 0 ) {
+				return unless (timeOut($timeout{ai_portal_wait}));
```
allanon: `ai_portal_wait` defaults to 0.5 seconds. This could cause deleted portals and re-routes on slow servers, until eventually the bot can't go anywhere because there aren't any portals left which the bot can use.

src/Task/MapRoute.pm
```diff
@@ -271,8 +271,25 @@ sub iterate {
...
+				if ($config{route_tryToGuessWrongPortalByDistance}) {
+					my $max_guess_distance = ($config{route_GuessWrongPortalMaxDistance} || 5);
+					for my $portal (@$portalsList) {
+						next unless (distance($self->{actor}{pos_to}, $portal->{pos}) <= $max_guess_distance);
+						$self->{actor}->sendMove(map int, @{$portal->{pos}}{qw(x y)});
```
allanon: Is the `map int` part necessary?

src/Task/MapRoute.pm
```diff
@@ -271,8 +271,25 @@ sub iterate {
...
+					}
```
allanon: Maybe add a `last` here, in case there are multiple portals? Although it probably wouldn't matter in the observed behavior of the bot.

src/Task/MapRoute.pm
```diff
@@ -271,8 +271,25 @@ sub iterate {
...
 	} elsif ( $portals_lut{"$self->{mapSolution}[0]{map} $self->{mapSolution}[0]{pos}{x} $self->{mapSolution}[0]{pos}{y}"}{source} ) {
 		# This is a portal solution
+		
+		
+		if ( distance($self->{actor}{pos_to}, $self->{mapSolution}[0]{pos}) == 0 ) {
+				return unless (timeOut($timeout{ai_portal_wait}));
+				if ($config{route_tryToGuessWrongPortalByDistance}) {
+					my $max_guess_distance = ($config{route_GuessWrongPortalMaxDistance} || 5);
+					for my $portal (@$portalsList) {
+						next unless (distance($self->{actor}{pos_to}, $portal->{pos}) <= $max_guess_distance);
```
allanon: I suggest picking the closest portal, rather than the one we happen to notice first.

```
my $pos = calcPosition($self->{actor});
for my $portal (sort { distance($pos, $a->{pos}) <=> distance($pos, $b->{pos}) } @$portalsList) {
    last if distance($pos, $portal->{pos}) > $max_guess_distance;
    $self->{actor}->sendMove(@{$portal->{pos}}{qw(x y)});
    last;
}
```

We probably also need to rewrite the route, or the other route code will try to move us back to the missing portal location next cycle.

src/Task/MapRoute.pm
```diff
@@ -271,8 +273,66 @@ sub iterate {
 
 	} elsif ( $portals_lut{"$self->{mapSolution}[0]{map} $self->{mapSolution}[0]{pos}{x} $self->{mapSolution}[0]{pos}{y}"}{source} ) {
 		# This is a portal solution
-
-		if ( distance($self->{actor}{pos_to}, $self->{mapSolution}[0]{pos}) < 2 ) {
+		
+		if ($self->{missing_portal}) {
+		
+			if (!$config{route_tryToGuessMissingPortalByDistance} || $portalsList->size == 0) {
```
allanon: This doesn't obey `guess_skip`. Maybe `$portalsList->size <= keys %{$self->{guess_skip} || {}}`?

allanon: Aha, so there's a second copy of this code inside the `route_tryToGuessMissingPortalByDistance` branch. Maybe we don't need the `$portalsList->size == 0` check here at all, then?

allanon:
Nice! I think it'll work from reading the code, but haven't tested. This should mostly solve the "portal moved slightly" and "portal disappeared entirely problems". This could still fail if there are two portals on screen, and the one we want is the farther of the two... but players would probably also be confused in that case, so that's okay.

kaliwanagan:
Someone that runs a private server (ahem @smallvill3) can maybe help out by actually removing one of the portals from one of the maps?

Henrybk:
I tested it by adding new non existing portals to portals.txt and making openkore try to use them just to be disabled by this script, it worked.

Also tested the 'moved slightly' by changing a portal position by like 7 cells in portals.txt, it worked aswell.

Henrybk:
@allanon All good now? hehe

flacrasec:
anything else needed to be merged?

Henrybk:
Testing

rayleigh21:
Hello do you have PRO bot can you give me active bot for pro

PR #791, opened 2017-07-29
Update portals.txt

alisonrag:
Brasilis Portals

based on rAthena portals:
https://github.com/rathena/rathena/blob/master/npc/re/warps/cities/brasilis.txt
https://github.com/rathena/rathena/blob/master/npc/re/warps/dungeons/bra_dun.txt
https://github.com/rathena/rathena/blob/master/npc/re/quests/quests_brasilis.txt

Working well in bRO and others brazilian private servers


kaliwanagan:
Cool! Just would like to get a second opinion from anyone playing in this server :)

PR #798, opened 2017-07-29
fix izlude in iRO Restart

allanon:
- [ ] Code Review

This uses the iRO official `resnametable.txt` file, for now, until iRO Restart gets its own version.


sctnightcore:
i love you @allanon

PR #808, opened 2017-07-30
[bRO] top10 ranks

lututui:
As reported by @guilhermeferreira1

By the way, does any official server still use the old top10 packets in serverType0?


kaliwanagan:
Should we have this specific for bRO ? I agree with @lututui to first confirm if there's another server that does this, so we can put this in a more generic location.

PR #815, opened 2017-08-01
Event macro

allanon:
Bring Henrybk's `eventMacro` into OpenKore.


madalilng:
is there a equivalent console condition from eventmacro like in macro?
for example like this.
console /That person is in another deal/i

Nipodemos:
No. 
Console is a bad condition , it's not good for Openkore 
There several other way of do things, like in your example 
You can send a pm to the other one confirming the deal started

madalilng:
okay thanks for the reply ...

madalilng:
can i suggest something ?
in the old macro we have the console right
for example
console /(Marthilda|Yosuke|Rubalkabara)/

we use $.lastMatch1to get the string that triggers the condition..
output : Rubalkabara

in eventMacro we have
NpcMsg /(Marthilda|Yosuke|Rubalkabara)/

we use $.NpcMsgLastMsg that returns the whole npc dialog that triggers the condition... 

output : 
[Father Mareusis]
 Please visit Father Rubalkabara, and return here, you could find him around North East side of Prontera city 

expected output : Rubalkabara

Nipodemos:
Thats a feature that I also want it :-)

Issue #826, opened 2017-08-02
Wrong npc sell and buy packets in restart


Henrybk:
Summary: The current packet steps for npc selling  and buying are wrong in iRO restart.

I did these tests with wireshark logged in normally in my character.

Test 1:
```
    Instructions:
        Click npc
        Cancel
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 09d4 (sendSellComplete)
```

Test 2:
```
    Instructions:
        Click npc
        Sell
        Cancel empty
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 00c5 (sendNPCBuySellList)
        recv 00c7 (npc_sell_list)
        send 09d4 (sendSellComplete)
        send 09d4 (sendSellComplete)
```

Test 3:
```
    Instructions:
        Click npc
        Sell
        Sell empty
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 00c5 (sendNPCBuySellList)
        recv 00c7 (npc_sell_list)
        send 09d4 (sendSellComplete)
        send 09d4 (sendSellComplete)

```
```
Test 4:
    Instructions:
        Click npc
        Sell
        Add item
        Sell
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 00c5 (sendNPCBuySellList)
        recv 00c7 (npc_sell_list)
        send 00c9 (sendSell or sendSellBulk)
        send 09d4 (sendSellComplete)
        send 09d4 (sendSellComplete)
        recv 00cb (sell_result)
```

Test 5:
```
    Instructions:
        Click npc
        Sell
        Add item
        Cancel
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 00c5 (sendNPCBuySellList)
        recv 00c7 (npc_sell_list)
        send 09d4 (sendSellComplete)
        send 09d4 (sendSellComplete)
```

Test 6:
```
    Instructions:
        Click npc
        Buy
        Cancel empty
```
    
```
    Results:
        send 0090 (npc_talk)
        recv 00c4 (npc_store_begin)
        send 00c5 (sendNPCBuySellList)
        recv 00c6 (npc_store_info)
        send 09d4 (sendSellComplete)
        send 09d4 (sendSellComplete)
```

[...]

PR #829, opened 2017-08-02
Complete rework of npc buy and sell interactions (including buyauto and sellauto)

Henrybk:
These changes adapt openkore to the behaviour seen with wireshark npc interaction in iRO Re:Start showcased in #826 

Packets captured in wireshark: https://i.gyazo.com/1d5e1b8f35eebf329489fafd2713cb8b.png

In the orange boxes it can be seen that the sell and buy packets are sent right before the double 09D4 packets, even in the same second.


control/config.txt
```diff
@@ -692,6 +692,7 @@ buyAuto {
 	price
 	minAmount 2
 	maxAmount 3
+	maxAtTime
```
allanon: `batchSize`?

Henrybk: I guess it is a better name

src/Network/Receive.pm
```diff
@@ -4068,11 +4068,14 @@ sub npc_store_info {
 		my $ID = unpack("v1", substr($msg, $i + 9, 2));
 
 		my $store = $storeList[$storeList] = {};
-		# TODO: use itemName() or itemNameSimple()?
-		my $display = ($items_lut{$ID} ne "")
-			? $items_lut{$ID}
-			: T("Unknown ").$ID;
-		$store->{name} = $display;
+		
+		my $name = itemNameSimple($ID);
```
allanon: Why `itemNameSimple` instead of `itemName`?

src/Commands.pm
```diff
@@ -75,6 +75,7 @@ sub initHandlers {
 	buyer				=> \&cmdBuyer,
 	bs					=> \&cmdBuyShopInfoSelf,
 	c					=> \&cmdChat,
+	cancel				=> \&cmdCancel,
```
allanon: `cancel` is a very generic term, which could mean a lot of things which aren't related to buying/selling. I recommend  `buy cancel` and `sell cancel`, even if they do the same thing, and even though `store` and `sell` are top-level commands. Though only `sell` matches its sub-command name, so `store` is pretty messed up too - it probably should be named `buy list` or something.

src/InventoryList.pm
```diff
@@ -165,6 +165,23 @@ sub getByNameID {
 }
 
 ##
+# Actor::Item $InventoryList->sumByNameID(nameID)
+#
+# Returns the amount of items with a given nameID.
+# If nothing is found, 0 is returned.
+sub sumByNameID {
+	my ($self, $id) = @_;
+	my $sum = 0;
```
Henrybk: I think that having this function may prove useful.

src/Misc.pm
```diff
@@ -200,7 +200,12 @@ our @EXPORT = (
 	closeShop
 	inLockMap
 	parseReload
-	setCharDeleteDate/
+	setCharDeleteDate/,
+	
+	# Npc buy and sell
+	qw/cancelNpcBuySell
+	completeNpcSell
+	completeNpcBuy/
```
Henrybk: Good idea

src/Network/Receive.pm
Henrybk: Because itemName requires an Actor::Item as parameter and npc item lists do not create that, it could be used if npcs properly used inventoryList and Actor::Item

src/Commands.pm
Henrybk: I agree it is not the best command name but the overall npc buy/sell interaction is pretty messed up, I think having a third command is a good idea, but maybe not 'cancel'.

flacrasec:
what is the matter with Receive.pm?

koynro:
🎉 🎉

oleokoong:
Thanks

ya4ept:
we forgot to update the documentation:
https://openkore.com/wiki/BuyAuto

PR #833, opened 2017-08-02
enable eventMacro, fix &inventory and friends, pass params to called macros

allanon:
- [x] Handle TODOs
- Code Review

### General
* Enable eventMacro plugin.
* Fix `&cart()`, `&inventory()`, `&storage()`, `&venderitem()`.

### Add `@array = &split()` function.

One can do this (painfully) with `&eval()`, like so:
```
$list = a,b,c,d
@array = ()
&push(@array, &eval( (split(',', '$list'))[0] ))
&push(@array, &eval( (split(',', '$list'))[1] ))
&push(@array, &eval( (split(',', '$list'))[2] ))
&push(@array, &eval( (split(',', '$list'))[3] ))
```

This pull request adds the ability do this in a simpler way:
```
$list = a,b,c,d
@array = split(',', $list)
```

### Remove repeat option on `call macro_name`, replace with params.

Previously, there were two parameters to `call`: the macro name and a "repeat" parameter, which said how many times to call the macro. Most people probably didn't know the feature existed, and I don't think I've ever seen code using it.

This pull request changes the meaning of extra parameters to `call`, so that they now override the `$.param1`, `$.param2`, ..., `$.paramN` magic variables instead. This provides the ability to pass arbitrary parameters macros, like so:
```
call quest_talk prontera "140 100" "r0 e"
```

### Support calling macros by variable name.

It can (rarely) be useful to implement a dispatch table by passing a macro name around, and then calling the macro with the same name. This requires a sequence of `if` statements, something like this:
```
if ($macro = foo) call foo
if ($macro = bar) call bar
if ($macro = baz) call baz
```

As a side-effect of allowing variables in `call` params (see above), this pull request allows macros to be called by a variable name directly, like so:
```
call $macro
```

### DIFF
--- a/plugins/eventMacro/eventMacro/Runner.pm
+++ b/plugins/eventMacro/eventMacro/Runner.pm
@@ -10,7 +10,7 @@ use Globals;
 use AI;
 use Log qw(message error warning debug);
 use Text::Balanced qw/extract_bracketed/;
-use Utils qw/existsInList/;
+use Utils qw/existsInList parseArgs/;
 use List::Util qw(max min sum);
 
 use eventMacro::Data;
@@ -1080,6 +1080,11 @@ sub next {
 						$eventMacro->set_full_hash($var->{real_name}, \%hash);
 					}
 					
+				} elsif ($var->{type} eq 'array' && $value =~ /^$macro_keywords_character(?:split)\(\s*(.*?)\s*\)$/) {
+					my ( $pattern, $scalar ) = map {/^\s*(.*?)\s*$/} parseArgs( "$1", undef, ',' );
+					my $scalar_var = find_variable( $scalar );
+					$self->error( "Scalar variable not recognized" ), return if !$scalar_var;
+					$eventMacro->set_full_array( $var->{real_name}, [ split $pattern, $eventMacro->get_scalar_var( $scalar_var->{real_name} ) ] );
 				} elsif ($var->{type} eq 'array' && $value =~ /^$macro_keywords_character(keys|values)\(($hash_variable_qr)\)$/) {
 					my $type = $1;
 					my $var2 = find_variable($2);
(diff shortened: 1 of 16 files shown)


plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -1080,6 +1080,11 @@ sub next {
 						$eventMacro->set_full_hash($var->{real_name}, \%hash);
 					}
 					
+				} elsif ($var->{type} eq 'array' && $value =~ /^$macro_keywords_character(?:split)\(\s*(.*?)\s*\)$/) {
+					my ( $pattern, $scalar ) = map {/^\s*(.*?)\s*$/} parseArgs( "$1", undef, ',' );
+					my $scalar_var = find_variable( $scalar );
+					$self->error( "Scalar variable not recognized" ), return if !$scalar_var;
+					$eventMacro->set_full_array( $var->{real_name}, [ split $pattern, $eventMacro->get_scalar_var( $scalar_var->{real_name} ) ] );
```
Henrybk: Maybe add support for any type of var? Like $array[0] and $hash{key} besides just $scalar.

plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -1781,22 +1783,23 @@ sub parse_command {
...
 		} elsif ($keyword eq 'inventory') {
-			$result = getInventoryIDs($parsed);
+			$result = (getInventoryIDs($parsed))[0];
 			
 		} elsif ($keyword eq 'Inventory') {
 			$result = join ',', getInventoryIDs($parsed);
 			
 		} elsif ($keyword eq 'store') {
+		    # TODO: Fix this to work with the new getItemIDs() implementation.
 			$result = getItemIDs($parsed, \@::storeList);
```
Henrybk: Yeah, actually these npc lists should use inventoryList but that is for another day.

Henrybk:
🎉 For inventoryList implementation

PR #884, opened 2017-08-09
Add sub for ai state changes

Henrybk:


src/AI.pm
```diff
@@ -82,8 +82,22 @@ use constant MANUAL => 1;
 # AI is turned on.
 use constant AUTO => 2;
 
+my $AI = AUTO;
```
allanon: This is pretty aggressive in preventing unauthorized state changes. I'd recommend using `our` instead of `my`, in case somebody has a legitimate reason to bypass `AI::state`. Maybe with a comment saying "use AI::state instead of modifying this variable directly".

src/Commands.pm
```diff
@@ -1676,13 +1670,13 @@ sub cmdDebug {
 
 		message center(T(" Debug information "), 56, '-') ."\n".
 			TF("ConState: %s\t\tConnected: %s\n" .
-			"AI enabled: %s\t\tAI_forcedOff: %s\n" .
+			"AI enabled: %s\n" .
```
allanon: Refactors on these multi-line translated messages causes the whole message to lose its translations. What do you think about making it policy to split them up into individual sentences (or maybe lines)? We don't have to do it here, I'm just thinking about it.

allanon:
What about the `--ai` command line option? https://github.com/OpenKore/openkore/blob/master/src/Settings.pm#L243-L245

allanon:
How about plugins?

* https://github.com/OpenKore/openkore/blob/master/plugins/needs-review/automap/automap.pl#L98
* https://github.com/OpenKore/openkore/blob/master/plugins/eventMacro/eventMacro/Core.pm#L1300

Henrybk:
`parseArguments` is called before `Globals` is loaded, i will have to find a solution for that.

Henrybk:
Testet with macros https://pastebin.com/9zfEZyGf on eventMacro and working

PR #907, opened 2017-08-12
talknpc improvements

Henrybk:
This makes openkore able to correctly end a conversation with an npc that auto closes the conversation, like in the knight job change quest.

### DIFF
--- a/src/Task/TalkNPC.pm
+++ b/src/Task/TalkNPC.pm
@@ -84,6 +84,7 @@ sub new {
 	$self->{type} = $args{type};
 	$self->{x} = $args{x};
 	$self->{y} = $args{y};
+	$self->{ID} = $args{ID};
 	$self->{nameID} = $args{nameID};
 	$self->{sequence} = $args{sequence};
 	$self->{sequence} =~ s/^ +| +$//g;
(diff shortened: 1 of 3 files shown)


src/Task/TalkNPC.pm
```diff
@@ -619,51 +640,28 @@ sub iterate {
...
-				debug "Done talking with $self->{target}, but another NPC initiated a talk instantly\n", 'ai_npcTalk';
-				# TODO: maybe better create a new task
-				debug "Continuing the talk within the same task\n", 'ai_npcTalk';
-				my $target = $self->find_and_set_target;
+		# Usual end of a conversation
+		if ($self->noMoreSteps && !%talk) {
+			$self->conversation_end;
+		
+		# There are more steps but no conversation with npc
+		} elsif (!%talk) {
```
allanon: This seems dangerous. What if somebody has an extra conversation step, but the npc refuses to talk after the first time? Or accepts the conversation but doesn't allow the conversation step (eg, an "r0" step for an npc which doesn't offer options)? Will it keep attempting to talk to the NPC over and over?

Henrybk: it won't reinitiate the conversation unless the next step is x, it will just print an error.

PR #932, opened 2017-08-16
Fix eventMacro lock/release 'all'

Henrybk:


Henrybk:
Tested with macros https://pastebin.com/BindF7aS and working

PR #933, opened 2017-08-16
add inTown selfCondition

Nipodemos:
As the title says, it adds inTown selfCondition
Is useful for getAuto and BuyAuto
When you don't want your bot leaving the lockmap (too far from any city) to do just these things.
With this condition the bot will only buy or get when he dies or do autostorage/autosell


lututui:
Why? There's already notInTown. `notInTown 0` and `inTown 1` would behave the exact same way

Nipodemos:
oh.... i didn't know that NotInTown 0 would behave in the same way...
so it's really useless... 😅

lututui:
Oh, actually it doesn't. Sorry :/

notInTown 0 triggers if you are in town. Maybe we should treat this as an issue instead of adding `inTown`?

kaliwanagan:
@lututui is this a bug or a valid addition to the conditions?

PR #956, opened 2017-08-19
Iro restart 2017 08 18

kaliwanagan:
# WARNING: This does not work yet.
## If you're interested in helping out and you know what you're doing, checkout this branch and try it. Then come over and visit us in IRC and say "I tested the branch, and this is what I found:"

This is just initial support for the new packets in iRO restart.
* iRO Restart is now using the crypt keys just like kRO
* it seems they are using brand new packets
* the new packets are just using the same format as server type 0
* there's a cycle of some sort going on with the sync and move packets where the initial 3 or 5 packets (depending on the packet) are different, then they all eventually stabilize to just one packet.
* because of the packet cycle, it's advised to not use this without xKore 1 or 3 as it may lead to the character being flagged and then banned


kaliwanagan:
Ah crap I should have squashed my commit messages :D

hnomkeng:
Missing character on selection.

bebox123:
It actually works however it is stuck only in one cell even with xkore 1.

src/Network/Send/iRO/Restart.pm
```diff
@@ -0,0 +1,48 @@
...
+
+	my %handlers = qw(
+		actor_info_request 48FF
+    actor_look_at 49A3
+		character_move 49B0
+		sync 4AD0
+	);
+	$self->{packet_lut}{$_} = $handlers{$_} for keys %handlers;
+
+	$self->cryptKeys(0x42780CC0, 0x67F86D28, 0x1CEB0ADC);
```
allanon: Should be unnecessary, because of `sendCryptKeys` in servers.txt.

kaliwanagan:
ah yeah, just saw this addition by @lututui :)

BatangKuneho:
i used it but got problem when char. get in the map the automatically disconnected and reconnect 60mins :( anyone got this ?

boom19:
does anyone made it work as of this time? :) still got no update on the other thread. although they made some changes like 10 hours ago.

kaliwanagan:
Locking to prevent spam. Because every complaint here goes to our inbox, and interrupts us, which causes us to spend more time trying to resume work.

See http://www.paulgraham.com/makersschedule.html

kaliwanagan:
closing in favor of #957 957

PR #957, opened 2017-08-19
fix iRO Restart

allanon:
This is enough to get iRO Restart working.

~~However, the server sends a client check packet of some sort periodically, and if it is not responded to correctly, the account will receive a 30 minute ban. This code does not respond correctly, and the ban will happen. Beware.~~ The `sync_ex` issue appears to be fixed, it ran all night without getting disconnected.

There are probably still some packets which were shuffled and aren't being handled correctly. Those will cause disconnects if there are any.

PR #958, opened 2017-08-19
iRO Bypass changes

Henrybk:


Henrybk:
It is working well to connect and play but there may be a few check packets that have not been implemented yep, use of these changes are dangerous, use at your own risk.

sctnightcore:
WORKKK !!

kaliwanagan:
@Henrybk how different is your branch from #957 ? Maybe we should standardize on one branch so we can clean up on that one?

Hurancan:
Is it working for you guys? I tried to but i keep dcing from map server.

ihansouza:
Do u know how to scape from 30 minutes ban?

alisonrag:
download the latest version and use #957

PR #959, opened 2017-08-20
Enhancement/increase timeout for getinfo to 2

kaliwanagan:
this prevents kore from contributing to lag, as kore tries to request for the actor information from every actor in the screen.

this is especially troublesome in highly botted maps like the payon dungeon, or highly populated maps like prontera, where each kore instance will try to request each other's information.

if there are more than 20 actors in the screen, there's 20^2 additional extra packets that is being sent and which the server needs to respond to.

normal client behavior is to send the getinfo packet only when the mouse hovers on the actor.

later we can refactor this so the timeout scales according to the number of actors that are visible, and we can weigh the getinfo sequence according to distance.


alisonrag:
great.
to the future, we can add one option to get only infos of actor types or maybe send the actor info request only to npcs in cities and limit the cells to send actor info request

PR #971, opened 2017-08-22
fix restart portal Xth time

Henrybk:
Just lost 500k on this little bug.


phamtrong204:
Very sorry xD.
My mistake!

alisonrag:
When we use the kafra service, after teleport, we must verify that the destination is correct

PR #976, opened 2017-08-23
beta fix for restart premium character slots

Henrybk:
Should fix http://forums.openkore.com/viewtopic.php?f=16&t=212255 but not in an elegant manner.

Proof: 
94e7e9818ab7c33893f7822a0e76e437

### DIFF
--- a/src/Network/Receive/ServerType0.pm
+++ b/src/Network/Receive/ServerType0.pm
@@ -2903,12 +2903,19 @@ sub received_characters {
 	# it doesn't work...
 	# 30 Dec 2005: it didn't work before because it wasn't sending the accountiD -> fixed (kaliwanagan)
 	$messageSender->sendBanCheck($accountID) if (!$net->clientAlive && $masterServer->{serverType} == 2);
-	if (!$masterServer->{pinCode}) {
-		$charLoginTimeout = {timeout => 2, time => time};
-		
-	} else {
+	
+	if ($masterServer->{pinCode}) {
 		message T("Waiting for PIN code request\n"), "connection";
 		$timeout{'charlogin'}{'time'} = time;
+		
+	} elsif ($masterServer->{pauseCharLogin}) {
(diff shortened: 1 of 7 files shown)


src/Network/Receive/ServerType0.pm
```diff
@@ -2904,11 +2904,8 @@ sub received_characters {
 	# 30 Dec 2005: it didn't work before because it wasn't sending the accountiD -> fixed (kaliwanagan)
 	$messageSender->sendBanCheck($accountID) if (!$net->clientAlive && $masterServer->{serverType} == 2);
 	if (!$masterServer->{pinCode}) {
-		if (charSelectScreen(1) == 1) {
-			$firstLoginMap = 1;
-			$startingzeny = $chars[$config{'char'}]{'zeny'} unless defined $startingzeny;
-			$sentWelcomeMessage = 1;
-		}
+		$charLoginTimeout = {timeout => 2, time => time};
```
alisonrag: Would be nice if we also add a variable in the timeouts.txt

koodpzok:
Hi, Wouldn't just a simple line "pauseCharServer 1" in servers.txt solve this ?

Henrybk:
No it wouldn't that is used to wait before conecting to the character server, this is used to wait before selecting a character in the char server.

kaliwanagan:
The wait is because we're expecting a new packet to arrive, or that the server won't accept it if we're going too fast?

Henrybk:
Thr server is now sending 1 099D packet for each 3 characters.

kaliwanagan:
I see. And I'm guessing we have no way of finding out how many characters there are until the server sends them all (unlike before where we get all of the characters in one go)

Henrybk:
Yep

lututui:
Does this work if the server is going to send more than 2 packets containing the chars?

Henrybk:
Yep @lututui 

5c57785d4caa45def911e305b3b8eb32

PR #988, opened 2017-08-23
handle packet shuffling via a config file

allanon:
Packet shuffling seems to be a preferred technique these days. Support it via a config instead of having to make code changes each time.

This also fixes a few bad changes made previously for iRO Restart:
* `0891` is `map_loaded`, not a new packet
* `0113` (`skill_use`) did not change to a new packet - it worked anyway for some reason
* `0089` (`actor_action`) did not change to a new packet - it worked anyway for some reason
* `092B` is `request_cashitems`, not `map_loaded`

And this should also handle shuffling for a few packets that weren't handled:
* `0178` => `0964` (identify)
* `0202` => `095B` (friend_request)
* `0802` => `087E` (booking_register)
* `0811` => `0896` (buy_bulk_openShop)
* `0815` => `0A69` (buy_bulk_closeShop)
* `0817` => `091E` (buy_bulk_request)
* `0819` => `022D` (buy_bulk_buyer)


src/Network/PacketParser.pm
```diff
@@ -139,6 +139,7 @@ sub create {
 	}
 	
 	my $self = $class->new;
+	$self->shuffle if $self->can( 'shuffle' );
```
kaliwanagan: https://media.giphy.com/media/28DimghjEXmO4/giphy.gif

Issue #991, opened 2017-08-24
ServerType issue


kumulo123:
Hi, i've downloaded the "latest ready-to-use revision" here -> https://github.com/OpenKore/openkore
when i'm trying to connect to fRO i get this error ->

![capture d ecran 2017-08-24 a 14 36 55](https://user-images.githubusercontent.com/31310913/29664826-008db322-88da-11e7-86e6-8a901391cf71.png)

and if i try to change the ServerType from fRO to 0,1,2,3,...,22

![capture d ecran 2017-08-24 a 14 41 07](https://user-images.githubusercontent.com/31310913/29665430-b083d6a6-88dc-11e7-8765-99b0957c2435.png)

**of course i've checked 10 times if my password is good, and it is.**

## here is the fRO settings in servers.txt

![capture d ecran 2017-08-24 a 14 37 06](https://user-images.githubusercontent.com/31310913/29665593-6ab82a5e-88dd-11e7-8f80-f04ddd57e833.png)

## and here is the src/Netwrok/Send/fRO.pm 

![capture d ecran 2017-08-24 a 14 38 22](https://user-images.githubusercontent.com/31310913/29665634-a220e076-88dd-11e7-8fa2-49cbe3c5e223.png)

**i don't know if it's important but i'm using the bot on virtual box...**

# thanks for your help !!

alisonrag:
add:
`use Globals;`

kumulo123:
Thanks but where ?

alisonrag:
after 
`use base` in fRO.pm

PR #992, opened 2017-08-24
fix-missing-import-fRO.pm

alisonrag:
reference: #991


lututui:
Isn't it better to change

```$messageSender->sendToServer(pack 'C*', 0xD4, 0x09);```

to

```$self->sendToServer(pack 'C*', 0xD4, 0x09);```

instead of importing Globals?

alisonrag:
updated

Issue #696, opened 2017-07-15
last update: fatal error


Glikadin:
:(((((((
OpenKore version what-will-become-2.1
@ai_seq = NPC macro
Network state = 5
Network handler = Network::DirectConnection
Revision: ctime:2017_07_11
Loaded plugins:
  plugins/alertsound.pl (alertsound; description: plays sounds on certain events)
  plugins/breakTime.pl (breakTime; description: Automatically disconnect and reconnect at certain times of the day)
  plugins/busCommands.pl (busCommands; description: receive and send commands (and messages too) via BUS system)
  plugins/customCaption.pl (customCaption; description: customize OpenKore's window caption)
  plugins/killcountFix/killcountFix.pl (killcountFix; description: fixes killcount quest goals on some servers)
  plugins/macro/macro.pl (macro; description: allows usage of macros)
  plugins/raiseSkill/raiseSkill.pl (raiseSkill; description: automatically raise character skills)
  plugins/raiseStat/raiseStat.pl (raiseStat; description: automatically raise character stats)
  plugins/spamblocker.pl (spamblocker; description: Spam Blocker Plugin)

Error message:
Can't locate object method "waitingForSteps" via package "Task::TalkNPC" at plugins/macro/Macro/Utilities.pm line 52.

Stack trace:
Can't locate object method "waitingForSteps" via package "Task::TalkNPC" at plugins/macro/Macro/Utilities.pm line 52.
 at src/Interface/Wx.pm line 163
	Interface::Wx::mainLoop('Interface::Wx=HASH(0x45a4994)') called at openkore.pl line 97
	main::__start() called at start.pl line 136

Died at this line:
  	return 1 if (AI::is('macro', 'deal'));
* 	return 1 if (AI::is('NPC') && $char->args->waitingForSteps);
  	return 0;

**This happens when the instruction is executed from macros.txt (do talknpc).**

Glikadin:
timeouts.txt :
ai_npc_talk_wait_to_answer 1.5
ai_npc_talk_wait_after_close_to_cancel 0.5
ai_npc_talk_wait_after_cancel_to_destroy 0.5

kaliwanagan:
@Glikadin if you're able to, can you join us in irc? You can use the webclient at https://webchat.freenode.net/?channels=openkore if you don't have your own IRC client.

Glikadin:
I found a temporary solution:

.../plugins/macro/Macro/Utilities.pm                Line № 52

[...]

PR #995, opened 2017-08-24
update macro utilities

alisonrag:
reference:
#696

### DIFF
--- a/plugins/macro/Macro/Utilities.pm
+++ b/plugins/macro/Macro/Utilities.pm
@@ -49,7 +49,7 @@ sub ai_isIdle {
 		}
 	}
 	return 1 if (AI::is('macro', 'deal'));
-	return 1 if (AI::is('NPC') && $char->args->waitingForSteps);
+	return 1 if (AI::is('NPC') && $char->args->{waitingForSteps});
 	return 0;
 }
 


plugins/macro/Macro/Utilities.pm
```diff
@@ -49,7 +49,7 @@ sub ai_isIdle {
 		}
 	}
 	return 1 if (AI::is('macro', 'deal'));
-	return 1 if (AI::is('NPC') && $char->args->waitingForSteps);
+	return 1 if (AI::is('NPC') && $char->args->{waitingForSteps});
```
allanon: This shouldn't ever happen. `$char->args` is really `AI::args`, and the `AI::args` for an NPC AI sequence should always be a `Task::TalkNPC`, which has `waitingForSteps`. Whatever code is creating the need for this change should be fixed instead.

alisonrag: I agree with your opinion but this code is just to update the macro plugin to work with the latest talk npc task .
we already have eventMacro, if someone wants to update the macro plugin, feel free to this.

allanon: How does this even happen?

* `talknpc` [calls `AI::ai_talkNPC`](https://github.com/OpenKore/openkore/blob/master/src/Commands.pm#L4999)
* `AI::ai_talkNPC` creates an AI sequence with the [args set to a Task::TalkNPC task](https://github.com/OpenKore/openkore/blob/master/src/AI.pm#L648)
* `Task::TalkNPC` [has a `waitingForSteps` method](https://github.com/OpenKore/openkore/blob/master/src/Task/TalkNPC.pm#L797)

So there shouldn't be a bug. If people upgraded only a few openkore files instead of the whole thing, it's okay if it doesn't work, IMO.

flacrasec: I had a problem with an old macro that use do autosell do you think this will fix this problem when talk to the vender npc? @alisonrag

flacrasec: I do like to inform that this PR corrected the issue from macro.pl autosell conversation: before this my macro was not finishing the autosell and now it is: 
image

flacrasec:
unfotunately macro plugin sometime will desappear no one wants to review it anymore =( well... eventMacro here I go, but I'am using this change at utilities.pm and it is working very well. =P

lututui:
How to reproduce this?

PR #999, opened 2017-08-25
remove unconstructive code

alisonrag:
remove useless code


kaliwanagan:
Ooooh. But what does it *do* ? :D

alisonrag:
If the software name are XR-Kore or KSMode are set to 1 in config spam hook with "die" in content

lututui:
I find it amusing this is been there for at least 12 years (I couldn't find the commit, that's as far as I went) and no one noticed it/questioned it

alisonrag:
xr-kore are dated in sourceforge from 2006-01-17
the code must been more old.

PR #1019, opened 2017-08-28
bus: use json for data transport, new plugins

allanon:
* Change the `bus` transport mechanism from packed binary data to JSON.
* Add a `verbose` option to `Bus::Client`.
* Add a new built-in `CLIENT_LIST2` packet type, which returns data as an array of hashes instead of a flat hash.
* Add a new `bus_hook` plugin which converts bus messages into hook calls, and vice versa.
** Hook `bus/send` - allow plugins other than `bus_hook` to send bus messages.
** Hook `bus/sent` - called whenever `bus_hook` sends a bus message.
** Hook `bus/recv` - called whenever `bus_hook` receives a bus message.
** Hook `bus/recv/$MID` - where `$MID` is the actual message ID (eg, `JOIN` or `LEAVE`) - called whenever `bus_hook` receives a bus message with message ID equal to `$MID`
* Add a new `bus_party` plugin which sends party update packets every 0.2 seconds, and updates the party with data from the bus.

### DIFF
--- a/control/sys.txt
+++ b/control/sys.txt
@@ -49,7 +49,7 @@ loadPlugins 2
 # loadPlugins_list <list>
 #   if loadPlugins is set to 2, this comma-separated list of plugin names (filename without the extension)
 #   specifies which plugin files to load at startup or when the "plugin load all" command is used.
-loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill,map,reconnect,eventMacro
+loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill,map,reconnect,eventMacro,bus_hook,bus_party
 
 # skipPlugins_list <list>
 #   if loadPlugins is set to 3, this comma-separated list of plugin names (filename without the extension)
--- a/src/Bus/Messages.pm
+++ b/src/Bus/Messages.pm
@@ -90,9 +90,10 @@ use Exporter;
 use base qw(Exporter);
 use Encode;
 use Utils::Exceptions;
+use JSON;
 
 our @EXPORT_OK = qw(serialize unserialize);
-
+our $json = JSON->new->utf8->allow_nonref;
 
 ##
 # Bytes Bus::Messages::serialize(String ID, arguments)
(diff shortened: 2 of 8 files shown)


Nipodemos:
it would be good to be able to send a message via bus,instead of just commands, just like the old one

moneyisgod:
W
O
W

lututui:
Should bus_party be enabled by default?

Nipodemos:
having this error while trying to open **wxstart.exe**
image


when using openkore.pl works normally

also i have another thing to ask... how to use bus_commands??? i've notice that there is no command to use

and the bus screen its not separeted anymore, its mixes with the first openkore opened...

here's my feedback so far

lututui:
@allanon can't we use JSON::Tiny? It's already in src/deps

control/sys.txt
```diff
@@ -49,7 +49,7 @@ loadPlugins 2
 # loadPlugins_list <list>
 #   if loadPlugins is set to 2, this comma-separated list of plugin names (filename without the extension)
 #   specifies which plugin files to load at startup or when the "plugin load all" command is used.
-loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill,map,reconnect,eventMacro,item_weight_recorder
+loadPlugins_list macro,profiles,breakTime,raiseStat,raiseSkill,map,reconnect,eventMacro,item_weight_recorder,bus_hook,bus_party,bus_command
```
lututui: I don't think bus_party should be on by default

src/Bus/Messages.pm
```diff
@@ -90,9 +90,10 @@ use Exporter;
 use base qw(Exporter);
 use Encode;
 use Utils::Exceptions;
+use JSON;
```
lututui: JSON::Tiny is already part of src/deps, can we use it?

src/Bus/bus-command.pl
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env perl
+
+use strict;
+use warnings;
+
+use FindBin qw($RealBin);
+use lib "$RealBin/..";
+use lib "$RealBin/../deps";
```
lututui: `use lib "$RealBin/../.."`, since XSTools.dll is there by default.

The same goes for src/Bus/bus-server.pl

lututui:
@Nipodemos 
> and the bus screen its not separeted anymore, its mixes with the first openkore opened...

src/Bus/bus-server.pl is still working

> also i have another thing to ask... how to use bus_commands??? i've notice that there is no command to use

Use src/Bus/bus-command.pl

I don't know if it's the best way to do this (probably not), but I got it working this way
image

PR #1024, opened 2017-08-29
Enhancement/update pro server details

black1e04:
is there update for pRo thor/loki?


voltemir:
pRO's IP is 139.99.113.156 right?

Issue #1027, opened 2017-08-30
[bug] command "storage get" causing crash


Guiim:
###### Summary: 
After starting the dialog with kafra, and using the command "storage get" without indicating the rest of the command, it crashes openkore. Both in the wx interface and in openkore.pl
###### Affected configuration(s)/ file(s):
The storage get command
###### Impact:
crash and close interface wx or console cmd 
###### Expected Behavior:
Should return the correct way to use the command
storage get
ex: Syntax Error in function 'storage get' (storage)
Usage: storage get <storage index item#> <(amount)>
###### Actual Behavior:
When the command is used without indicating the index of the item and without the quantity the error is generated
###### Steps to Reproduce:
Use the storage get command without indicating the item index in the inventory After starting the dialog with kafra and opening the storage.
storage get
###### Additional Details:

error.txt
OpenKore version what-will-become-2.1
@ai_seq = storageAuto items_take route
Network state = 5
Network handler = Network::DirectConnection
Revision: git:ba20aeb_2017-08-30 ctime:2017_08_02
Loaded plugins:
  plugins/breakTime/breakTime.pl (breakTime; description: Automatically disconnect and reconnect at certain times of the day)
  plugins/macro/macro.pl (macro; description: allows usage of macros)
  plugins/map/map.pl (map; description: map)
  plugins/raiseSkill/raiseSkill.pl (raiseSkill; description: automatically raise character skills)
  plugins/raiseStat/raiseStat.pl (raiseStat; description: automatically raise character stats)
  plugins/reconnect/reconnect.pl (reconnect; description: v1.0)

Error message:
Assertion failed!
 at C:/Users/xxxxx/Desktop/openkore/src/deps/Carp/Assert.pm line 271
	Carp::Assert::assert('') called at src/InventoryList.pm line 246
	InventoryList::getMultiple('InventoryList::Storage=HASH(0x9cf12cc)', undef) called at src/Commands.pm line 4719
	Commands::cmdStorage_get(undef) called at src/Commands.pm line 4636
	Commands::cmdStorage('storage', 'get') called at src/Commands.pm line 303
	Commands::run('storage get') called at src/functions.pl line 998
	main::parseInput('storage get') called at src/functions.pl line 63
	main::mainLoop() called at src/Interface.pm line 75
	Interface::mainLoop('Interface::Console::Win32=HASH(0x7bbad4c)') called at C:\Users\xxxxx\Desktop\openkore\openkore.pl line 97
[...]

PR #1028, opened 2017-08-30
Fix due to #1027

Mortimal:
https://github.com/OpenKore/openkore/issues/1027

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -4631,14 +4631,12 @@ sub cmdStorage {
 			cmdStorage_log();
 		} elsif ($switch eq 'desc') {
 			cmdStorage_desc($items);
-		} elsif ($switch eq 'add' || $switch eq 'addfromcart' || $switch eq 'get' || $switch eq 'gettocart' || $switch eq 'close') {
+		}elsif(($switch =~ /^(add|addfromcart|get|gettocart)$/ && $items)
+				|| $switch eq 'close'
+			  )
+		{
 			if ($char->storage->isReady()) {
-			if (!$items
-				&& ($switch ne 'close')
-			   )
-			{
 				error T("Cannot get/add items because no items provided\n");
-			}else{
 				if ($switch eq 'add') {
 					cmdStorage_add($items);
 				} elsif ($switch eq 'addfromcart') {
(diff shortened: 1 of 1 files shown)


src/Commands.pm
```diff
@@ -4633,6 +4633,12 @@ sub cmdStorage {
 			cmdStorage_desc($items);
 		} elsif ($switch eq 'add' || $switch eq 'addfromcart' || $switch eq 'get' || $switch eq 'gettocart' || $switch eq 'close') {
 			if ($char->storage->isReady()) {
+			if (!$items
```
allanon: This works, but it's ugly. The `if` is spread over multiple lines, has unnecessary parentheses (some people swear by unnecessary parentheses (and I tend to use them when writing sentences too) so maybe that's not bad), and the indentation is wrong.

Also, it doesn't tell people why nothing happened. It's better to fall through to the usage if the syntax is wrong.

How about this instead?
```
diff --git a/src/Commands.pm b/src/Commands.pm
index ba7e6042f..e248e9492 100644
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -4631,7 +4631,7 @@ sub cmdStorage {
                        cmdStorage_log();
                } elsif ($switch eq 'desc') {
                        cmdStorage_desc($items);
-               } elsif ($switch eq 'add' || $switch eq 'addfromcart' || $switch eq 'get' || $switch eq 'gettocart' || $switch eq 'close') {
+               } elsif (($switch =~ /^(add|addfromcart|get|gettocart)$/ && $items) || $switch eq 'close') {
                        if ($char->storage->isReady()) {
                                if ($switch eq 'add') {
                                        cmdStorage_add($items);
```

Mortimal: k i understood il rewrite that

Mortimal:
thx @allanon all fixed

src/Commands.pm
```diff
@@ -4631,7 +4631,10 @@ sub cmdStorage {
 			cmdStorage_log();
 		} elsif ($switch eq 'desc') {
 			cmdStorage_desc($items);
-		} elsif ($switch eq 'add' || $switch eq 'addfromcart' || $switch eq 'get' || $switch eq 'gettocart' || $switch eq 'close') {
+		}elsif(($switch =~ /^(add|addfromcart|get|gettocart)$/ && $items)
```
allanon: Sigh. Why make the whitespace ugly?

windhamwong: lol I agree they are ugly, but at least it is working.

PR #1029, opened 2017-08-30
make call on an automacro accept params

Nipodemos:
Before, the only way to make call accept params is the call from inside a macro.
So you had to do something like this to get what you want:
```
automacro goSave {
    ConfigKeyNot saveMap prontera 
    call {
        call saveIn "prontera" "100 100"
    }
}
```
but with this update, the call of an automacro accepts directly the params, like this:
```
automacro goSave {
    ConfigKeyNot saveMap prontera 
    call saveIn "prontera" "100 100"
}
```

Obs: **macros cannot have spaces in name!!!!**
there shouldn't be even possible to add space in name, but its a feature not implemented yet.

i would like to thank **@Henrybk** for giving me tips to do this  and **@allanon** for made the call on macros to accept params in the first place, that piece of code was very useful to me.

Tested with this macro and working great:
```
automacro test_with_params {
	BaseLevel > 0
	exclusive 1
	run-once 1
	priority 1
	call test_macro "aldebaran" "170 155"
}

automacro test_without_params {
	BaseLevel > 0
	exclusive 1
	run-once 1
	priority -5
	call test_macro
}

macro test_macro {
	log param1 is : $.param1
	log param2 is : $.param2
	log $.caller running the macro without erros
}
```

### DIFF
--- a/plugins/eventMacro/eventMacro/Core.pm
+++ b/plugins/eventMacro/eventMacro/Core.pm
@@ -195,22 +195,29 @@ sub create_automacro_list {
 		}
 		
 		PARAMETER: foreach my $parameter (@{$value->{'parameters'}}) {
-		
 			###Check Duplicate Parameter
 			if (exists $currentParameters{$parameter->{'key'}}) {
 				warning "[eventMacro] Ignoring automacro '$name' (parameter ".$parameter->{'key'}." duplicate)\n";
 				next AUTOMACRO;
 			}
-			###Parameter: call
-			if ($parameter->{'key'} eq "call" && !$self->{Macro_List}->getByName($parameter->{'value'})) {
-				warning "[eventMacro] Ignoring automacro '$name' (call '".$parameter->{'value'}."' is not a valid macro name)\n";
-				next AUTOMACRO;
-			
+			###Parameter: call with or without param
+			if ($parameter->{'key'} eq "call" && $parameter->{'value'} =~ /(\S+)\s+(.*)?/) {
+				if (!$self->{Macro_List}->getByName($1) ) {
+					warning "[eventMacro] Ignoring automacro '$name' (call '".$1."' is not a valid macro name)\n";
+					next AUTOMACRO;
+				} else {
+					if (defined $2) {
+						#put both values together in call name, split later in sub call_macro
+						$parameter->{'value'} = join (" ",$1,$2);
+					} else {
+						$parameter->{'value'} = $1;
+					}
+					$currentParameters{$parameter->{'key'}} = $parameter->{'value'};
+				}
 			###Parameter: delay
 			} elsif ($parameter->{'key'} eq "delay" && $parameter->{'value'} !~ /\d+/) {
 				error "[eventMacro] Ignoring automacro '$name' (delay parameter should be a number)\n";
 				next AUTOMACRO;
-			
 			###Parameter: run-once
 			} elsif ($parameter->{'key'} eq "run-once" && $parameter->{'value'} !~ /[01]/) {
 				error "[eventMacro] Ignoring automacro '$name' (run-once parameter should be '0' or '1')\n";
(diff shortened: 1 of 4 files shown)


plugins/eventMacro/eventMacro/Core.pm
```diff
@@ -195,22 +195,29 @@ sub create_automacro_list {
...
 				next AUTOMACRO;
 			}
-			###Parameter: call
-			if ($parameter->{'key'} eq "call" && !$self->{Macro_List}->getByName($parameter->{'value'})) {
-				warning "[eventMacro] Ignoring automacro '$name' (call '".$parameter->{'value'}."' is not a valid macro name)\n";
-				next AUTOMACRO;
-			
+			###Parameter: call with or without param
+			if ($parameter->{'key'} eq "call" && $parameter->{'value'} =~ /(\S+)\s+(.*)?/) {
+				if (!$self->{Macro_List}->getByName($1) ) {
```
allanon: It's not safe to use regex capture variables (eg, `$1`, `$2`) after calling another function. If that function does a regex anywhere, it'll overwrite the capture variables. So this call is safe, but additional uses of `$1` in this code are not.

While this might not be a problem right now, if somebody changes `getByName` to include a regex, it'll break this code totally unexpectedly. Please store the capture variables in temporary variables if you need to use them more after calling a function.
```
my ($name, $variables) = ($1, $2);

File: plugins/eventMacro/eventMacro/Core.pm [hunk 2]
```diff
@@ -195,22 +195,29 @@ sub create_automacro_list {
...
+					warning "[eventMacro] Ignoring automacro '$name' (call '".$1."' is not a valid macro name)\n";
+					next AUTOMACRO;
+				} else {
+					if (defined $2) {
```
allanon: Do we need to do this? It just transforms the sequence of spaces between the name and variables into a single space.

plugins/eventMacro/eventMacro/FileParser.pm
```diff
@@ -113,6 +111,19 @@ sub parseMacroFile {
 				} else {
 					undef %block
 				}
+				
+				
+			} elsif ($_ =~ /call [^{]/) {
```
allanon: Perl syntax: `$_ =~` is unnecessary here, this can be written as `} elsif (/call [^{]/) {`.

I don't understand why this is here. It matches two cases: `call ^` and `call {`, and in either case, the text before and after is ignored. So this will match `notacall ^wat`, which probably isn't intentional?

The `call {` case completely prevents the following branch (`} elsif ($_ eq "call {") {`) from ever triggering. If that is intentional, we can remove the following branch.

plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -1407,13 +1407,12 @@ sub parse_call {
...
 	my $macro_name   = $call_command;
 	my $repeat_times = 1;
-	if ( $call_command =~ /\s/ ) {
+	if ( $call_command =~ /\s+/ ) {
 	    my @params;
 		( $macro_name, @params ) = parseArgs( $call_command );
 
 		# Update $.paramN with the values from the call.
 		$eventMacro->set_scalar_var( ".param$_", $params[ $_ - 1 ], 0 ) foreach 1 .. @params;
-		$eventMacro->set_scalar_var( ".param$_", undef,             0 ) foreach ( @params + 1 ) .. 100;
```
allanon: This line should not be removed. It undefines params which weren't passed. If we remove this, then the following will do unexpected things.

```
macro foo {
  call bar 1 2
  call bar a
}
macro bar {
  log 1: $.param1
  log 2: $.param2
}
```

After this change, the code above will unexpectedly log `2` on the second call to `bar`:
```
1: 1
2: 2
1: a
2: 2
```

plugins/eventMacro/eventMacro/Core.pm
Nipodemos: i didn't know that, i'll update the code, thanks for the tip

plugins/eventMacro/eventMacro/Core.pm
Nipodemos: now that you said, i've noticed, it's useless, thanks 😄

plugins/eventMacro/eventMacro/Runner.pm
Nipodemos: oh, you also right about that, i'll return this line of code to there.

PR #1030, opened 2017-08-30
fixed defined array depriciated

tlinkin:
After perl 5.14.4 defined @array was depriciated as it was a bug in perl. This should work on 5.14.4+

### DIFF
--- a/plugins/needs-review/webMonitor/trunk/webMonitorServer.pm
+++ b/plugins/needs-review/webMonitor/trunk/webMonitorServer.pm
@@ -380,7 +380,7 @@ sub request {
 	my ($i, $name, $class, $lvl, $title, $online, $ID, $charID);
 	my (@listMemberIndex, @listMemberName, @listMemberClass, @listMemberLvl, @listMemberTitle, @listMemberOnline, @listMemberID, @listMemberCharID);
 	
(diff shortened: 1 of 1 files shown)


Mortimal:
Maybe use $] to check version and use another syntax? if u rly need that....

plugins/needs-review/webMonitor/trunk/webMonitorServer.pm
```diff
@@ -380,7 +380,7 @@ sub request {
 	my ($i, $name, $class, $lvl, $title, $online, $ID, $charID);
 	my (@listMemberIndex, @listMemberName, @listMemberClass, @listMemberLvl, @listMemberTitle, @listMemberOnline, @listMemberID, @listMemberCharID);
 	
-	if (defined @{$guild{member}}) {
+	if (defined $guild{member}) {
```
allanon: `defined @{$guild{member}}` is equivalent to `$guild{member} && @{$guild{member}}`

tlinkin:
It was just a syntax error, it should work on older versions as well, I see no need to change.

I am not that well versed in perl, but from my research, everything looks normal.

This plugin required a lot of work, to bring it to a stable release, I am planning to fix some of things I understand.

PR #1034, opened 2017-08-30
repair sex for iRO renewal & odin; add sex & mapname on charSelectScreen

c4c1n6kr3m1:
- repair sex for iRO renewal & odin ( different sex on same account fixed )
- add sex & map on charSelectScreen
http://forums.openkore.com/viewtopic.php?f=36&t=212126

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -1185,7 +1185,7 @@ sub charSelectScreen {
 			}
 		}
 		
-		push @charNames, TF("Slot %d: %s (%s %s, level %d/%d, %s)%s",
+		push @charNames, TF("Slot %d: %s (%s, %s, level %d/%d, %s)%s",
 			$num,
 			$chars[$num]{name},
 			$jobs_lut{$chars[$num]{'jobID'}},
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -1185,12 +1185,14 @@ sub charSelectScreen {
 			}
 		}
 		
-		push @charNames, TF("Slot %d: %s (%s, level %d/%d)%s",
+		push @charNames, TF("Slot %d: %s (%s %s, level %d/%d, %s)%s",
 			$num,
 			$chars[$num]{name},
 			$jobs_lut{$chars[$num]{'jobID'}},
+			$sex_lut{$chars[$num]{sex}},
 			$chars[$num]{lv},
 			$chars[$num]{lv_job},
+			$chars[$num]{map_name},
 			$messageDeleteDate);
 		push @charNameIndices, $num;
 	}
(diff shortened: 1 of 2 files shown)


src/Misc.pm
```diff
@@ -1185,12 +1185,14 @@ sub charSelectScreen {
 			}
 		}
 		
-		push @charNames, TF("Slot %d: %s (%s, level %d/%d)%s",
+		push @charNames, TF("Slot %d: %s (%s %s, level %d/%d, %s)%s",
```
lututui: How does this affect other servers, if at all?

c4c1n6kr3m1: my english is not good

but i don't see any problem, as far as i remember,  if server doesn't send mapname, it will show empty string
can you test on your server?

lututui: image

Works with bRO. There should be a comma between Job and Sex,
```push @charNames, TF("Slot %d: %s (%s, %s, level %d/%d, %s)%s",```

Comment (c4c1n6kr3m1) [reply src/Misc.pm]: This is my first time , so how can i do that ? edit my commit ? do i need to make pull again ?

**Comment by Nipodemos** [comment]
You just make a commit in the branch you made the pull request , push it 
and it will update here 
git commit
git push

**Comment by alisonrag** [comment]
are working in others servers?

**Comment by c4c1n6kr3m1** [comment]
iRO ( with charBlockSize 145 ) allows an account to have different sex for a char, this code fixed it , 
for other servers , it does not make big difference , only show sex and map name ( if any ) on charSelectScreen , as @lututui shows on his screenshot

File: src/Network/Receive/ServerType0.pm [hunk 2]
```diff
@@ -2875,11 +2875,13 @@ sub received_characters {
 		$chars[$slot]{int} = $int;
 		$chars[$slot]{dex} = $dex;
 		$chars[$slot]{luk} = $luk;
-		$chars[$slot]{sex} = $accountSex2;
+		$chars[$slot]{sex} = ($masterServer->{charBlockSize} == 145 && $masterServer->{serverType} =~ /^iRO/) && (unpack( 'C', substr($args->{RAW_MSG}, $i + $blockSize -1)) =~ /^0|1$/)? unpack( 'C', substr($args->{RAW_MSG}, $i + $blockSize -1)) : $accountSex2;
```
lututui: Maybe we should make this a servers.txt option, so other servers with the same behavior can make use of it.

c4c1n6kr3m1: If we know other servers send the char sex at last byte ... but i don't.
I don't play other servers , and i have very limited bandwith to try other servers , so i limit to iRO only

PR #1037, opened 2017-08-31
Chose control folder automatically if no other profile is found

lututui:
We don't need to ask which profile to use if none but the control folder are available

### DIFF
--- a/plugins/profiles/profiles.pl
+++ b/plugins/profiles/profiles.pl
@@ -63,6 +63,11 @@ sub onStart {
       next if ($_ =~ /^\.|^#/);
       push @profiles, $_;
    }
+   
+   if (!@profiles) {
+		message "No profiles found, using standard control folder\n";
+		return;
+	}
 
    @profiles = sort { $a cmp $b } @profiles;
    push @profiles, 'Use standard control folder';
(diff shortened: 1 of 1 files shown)


Nipodemos:
THANK YOU! that box when you don't use it is annoying, although it's a great plugin. You just made this plugin better

alisonrag:
I really love you, the popup window bothered me for many time

plugins/profiles/profiles.pl
```diff
@@ -72,18 +72,23 @@ sub onStart {
...
+			);
+
+			my $num = @profiles;
+			return 0 if $choice == $num - 1;
+			return $quit = 1 if $choice == -1;
+
+			$profile = $profiles[$choice];
+		} else {
+			message "No profiles found, using standard control folder\n";
+			$profile = $profiles[0];
```
allanon: We shouldn't do this. `@profiles` will be `("Use standard control folder")` here, and we end up adding that to the list of control directories. It won't break anything, because that folder can't exist if we got here, but it won't help either. Just remove this line, and it should all work fine. Better yet, move this logic up and make it an early return instead:
```
if (@profiles == 1) {
    message "No profiles found, using standard control folder\n";
    return;
}
```

PR #1018, opened 2017-08-28
servertybe 2015-2016 

sctnightcore:
Wait to update the Receive and more.
NEED TEST
ty
20131223 by @moneyisgod 

20151104:**WORK**
RagexeRE_2016_02_03 : fix TEST in : http://zirvro.com/

PR #1055, opened 2017-09-02
update new sakexe for client 2016+

sctnightcore:
Sakexe_0 has not updated for a long time.
**send** and **rev**  for sakexe 
port servertybe0 to Sakexe_0

@alisonrag  
I think we should fix the private server to read servertybe0.


sctnightcore:
#1018

sctnightcore:
@alisonrag  @moneyisgod  
I think we should fix the private server to read qw servertybe0 ? 
Is it better?

sctnightcore:
@lututui  ok sir

PR #1057, opened 2017-09-02
negative exp

c4c1n6kr3m1:
show negative exp when die and show exp gained when level up


allanon:
If I'm reading this correctly, you're trying to change the "you killed a monster and gained 0 exp" message that happens when killing a monster causes a level up?

c4c1n6kr3m1:
@allanon
yes, 
0 exp when level up and when die

PR #1064, opened 2017-09-04
add item_weight_recorder plugin

allanon:
Adds a plugin which saves item weights to a file, and provides those weights via `Actor::Item->new->weight`. Item weights are very useful for predicting how many items we can get from storage, how many items until we'll be overweight, etc.


allanon:
I extracted `item_weights.txt` from iRO using this plugin (for a small number of items), and from divinepride for most of the remainder. The idea is that the plugin should self-heal if any of the weights are wrong.

PR #1071, opened 2017-09-04
fix call bug, moved $.param to array

Nipodemos:
when i've last implemented the parameters on an automacro call, a very serious bug appeared making every automacro that has a submacro that call another, to be ignored...

```
automacro test {
    BaseLevel > 0
    run-once 1
    exclusive 1
    call {
        call anotherMacro
    }
}
```

this automacro above do not run, it points out the error "parameter call duplicate"
this pull request fix this error 

move from $paramN to an array, which is much more natural
now instead of using:
`$.param1` or `$.param2`
you use `$.param[0]` or `$.param[1]`

### DIFF
--- a/plugins/eventMacro/eventMacro/Automacro.pm
+++ b/plugins/eventMacro/eventMacro/Automacro.pm
@@ -90,6 +90,7 @@ sub set_call {
 	my ($self, $parameters, $macro_name) = @_;
 	$self->{parameters}{'call'} = $macro_name;
 }
+
 sub set_parameters {
 	my ($self, $parameters) = @_;
 	foreach (keys %{$parameters}) {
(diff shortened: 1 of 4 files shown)


phamtrong204:
I think it is better when kept ```$.param1```,```$.param2``` for backward-compatibility

Henrybk:
I really don't think we should trouble ourselves with backwards compatibility when it clearly is a use of bad coding decisions. Param scalar was only used because there were no arrays in macro plugin, eventMacro is not meant to be backwards compatible with macro plugin so that should not be a problem, and we have stabilished the parameters functionality only about 2 weeks ago, so not so much trouble for the people who used them in these past two weeks.

I give it A+ in coding design personally.

PR #1077, opened 2017-09-05
Using arr for console param in eventMacro plugin

phamtrong204:
ref: https://github.com/OpenKore/openkore/pull/1071


plugins/eventMacro/eventMacro.pl
```diff
@@ -585,7 +585,7 @@ sub commandHandler {
 			next if $variable_name !~ /^\.param\d+$/o;
```
Henrybk: should also remove all this `if` block.

Henrybk:
Oh sorry, I meant that whole foreach block

PR #1093, opened 2017-09-07
Fix if block inside while block eventMacro

Henrybk:


Henrybk:
Have you confirmed it is working @phamtrong204 ?

phamtrong204:
Yes, work like a champion!

allanon:
This works for the `test_wi` macro below, but fails for the `test_ww` macro:
```
macro test_wi {
  $i = 0
  while ($i < 3) {
    log $i
    $j = 0
    if ($j < 3) {
      log $i.$j +
    }
    if ($j > 3) {
      log $i.$j -
    }
    $i++
    pause 0.1
  }
}

macro test_ww {
  $i = 0
  while ($i < 3) {
    log $i
    $j = 0
    while ($j < 3) {
      log $i.$j
      $j++
      pause 0.1
    }
    $i++
    pause 0.1
  }
}
```

Henrybk:
Thanks for pointing that out @allanon 
It should be working after the last commit.

allanon:
Competing implementation, which handles while-in-while as well as if-in-while cases: https://github.com/OpenKore/openkore/pull/1095

PR #1095, opened 2017-09-07
eventmacro: handle nested blocks

allanon:
Support nested blocks in eventMacro. This pull request does not attempt to clean up `if` handling (and maybe `switch` handling too?), but it seems likely that some of if/elsif/else logic could be handled in a similar way.

This was tested with the following macros, which work correctly:
```
macro test_wi {
  $i = 0
  while ($i < 3) {
    log $i
    $j = 0
    if ($j < 3) {
      log $i.$j +
    }
    if ($j > 3) {
      log $i.$j -
    }
    $i++
    pause 0.1
  }
}

macro test_ww {
  $i = 0
  while ($i < 3) {
    log $i
    $j = 0
    while ($j < 3) {
      log $i.$j
      $j++
      pause 0.1
    }
    $i++
    pause 0.1
  }
}
```

### DIFF
--- a/plugins/eventMacro/eventMacro/Runner.pm
+++ b/plugins/eventMacro/eventMacro/Runner.pm
@@ -41,7 +41,7 @@ sub new {
 	$self->{line_index} = 0;
 	
 	$self->{label} = {scanLabels($self->{lines_array})};
-	$self->{loops} = {scanLoops($self->{lines_array})};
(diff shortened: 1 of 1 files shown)


Henrybk:
Cool, this implementation could also be expanded to be used in `sub define_next_valid_command` to skip blocks with false conditions without having to check each line.

PR #1122, opened 2017-09-13
deal hooks

madalilng:
add deal hooks

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -3025,9 +3025,10 @@ sub deal_begin {
 
 	if ($args->{type} == 0) {
 		error T("That person is too far from you to trade.\n"), "deal";
+		Plugins::callHook("error_deal", { type =>$args->{type}} );
 	} elsif ($args->{type} == 2) {
 		error T("That person is in another deal.\n"), "deal";
-		Plugins::callHook("error_deal", { msg => "That person is in another deal."});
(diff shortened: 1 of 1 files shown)


src/Network/Receive.pm
```diff
@@ -3027,6 +3027,7 @@ sub deal_begin {
 		error T("That person is too far from you to trade.\n"), "deal";
 	} elsif ($args->{type} == 2) {
 		error T("That person is in another deal.\n"), "deal";
+		Plugins::callHook("error_deal", { msg => "That person is in another deal."});
```
lututui: Since this is the only place error_deal is called, do we need a msg argument?

alisonrag: i think the message is not necessary or we need to add a message in others case like player is too far or using the storage

lututui: Passing $args{type} is better than passing a string as argument

alisonrag: so the better way is:
Plugins::callHook("error_deal", { type =>$args->{type}});

madalilng: will i remove the engaged_deal hook and add 
hooks at the bottom with { type =>$args->{type}} ??

lututui: You can keep the other hooks the way they are. Just change it for the error hooks

lututui: @Mortimal no, there are other codes different from 3 that are not errors

Issue #1123, opened 2017-09-13
[EventMacro.pl] Problem with ( ) in conditions


Mortimal:
###### Summary:
When there is  ( ) inside any condition as part of a string if statement works incorrect

###### Affected configuration(s)/ file(s):
Runner.pm -> sub parse_and_check_condition_text (and subs it is calling) -> line 1212
###### Impact:
Getting error in comparison statements
###### Expected Behavior:.
Condition must be met
###### Actual Behavior:
Condition not met
###### Steps to Reproduce:
```
macro testme{
  $va1 = isatext
  log $va1
  $va2 = (isatext)
  log $va2
  if ($va1 == "isatext") {
   log correct
  }
  if ($va2 == "(isatext)") {
   log error here
  }
}
```

###### Summary:
Need to be rewritten all logic of conditions parsing
###### Affected configuration(s)/ file(s):
Runner.pm -> sub parse_and_check_condition_text (and subs it is calling) -> line 1212
###### Benefits:
1 You fix this...
2 Bug disappear...
3 .........
4 Profit!

Nipodemos:
#2368 fixed this issue.

PR #1124, opened 2017-09-13
[eventMacro.pl] Added &strip()

Mortimal:
Added &strip() as temporary fix for #1123.

### DIFF
--- a/control/timeouts.txt
+++ b/control/timeouts.txt
@@ -172,3 +172,6 @@ injectSync 5
 injectKeepAlive 12
 welcomeText 4
 patchserver 120
+
+# Time to wait before character connection when pauseCharLogin is active
+char_login_pause 2
--- a/plugins/eventMacro/eventMacro/Core.pm
+++ b/plugins/eventMacro/eventMacro/Core.pm
@@ -785,28 +785,34 @@ sub get_scalar_var {
 	if ( substr( $variable_name, 0, 1 ) eq '.' ) {
 
 		# Time-related variables.
-		if    ( $variable_name eq '.time' )     { return time; }
-		elsif ( $variable_name eq '.datetime' ) { return scalar localtime; }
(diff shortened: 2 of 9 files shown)


lututui:
This takes away the () from the match?

Mortimal:
this ads &strip() function which can be used to remove ( ) from $. vars like

```
automacro monster {
 SimpleHookEvent target_died
 call monster2 
 } 
macro monster2 {
	if (&strip($.SimpleHookEventLastMonster) =~ /Poring/) {
	 log yes
	 stop
	}
	log not
	stop 
}
```

Because normal fix is just impossible... now

lututui:
Doesn't this have the same behavior
`if (&strip($.SimpleHookEventLastMonster) =~ /Poring/) {`
as this
`if ($.SimpleHookEventLastMonster =~ /Poring/) {`
?

Mortimal:
`if ($.SimpleHookEventLastMonster =~ /Poring/) {`

this will be error because of #1123 

**$.SimpleHookEventLastMonster** contains **Poring (\d)** and if \d not 0 or 1 condition parser gives error...

Nipodemos:
i've noticed something similar to what happened to you
i've tried this:
```
               $name = name: Nipo lastname: demos
		log $name
		if ($name =~ /name: (\w+) lastname: (\w+)/) {
			log it worked 
		}
```
the intereresting part is , the code removes the parentesis: 

image

so `/name: (\w+) lastname: (\w+)/` is changed to `/name: \w+ lastname: \w+/`
and i don't know why, do you know what sub is doing this thing?
I think is whenever the `sub parse_command` is called
which calls `parse_keywords`
which calls `extract_brackted` which is meant to remove `()`

Nipodemos:
@Mortimal do you have irc or discord? i would like to discuss some improvements on eventMacro, and two heads think better than one 😄

Nipodemos:
just tested and is working fine

PR #1132, opened 2017-09-14
[eventMacro.pl] if parsing fix

Mortimal:
if (<condition>){ <- error
if (<condition>) { <- not error
if (<condition>)call <- error
if (<condition>) call <- not error
if (<condition>)goto <- error
if (<condition>) goto <- not error
but why?
....
fixed

### DIFF
--- a/plugins/eventMacro/eventMacro/Runner.pm
+++ b/plugins/eventMacro/eventMacro/Runner.pm
@@ -660,7 +660,7 @@ sub define_next_valid_command {
 		# Initial 'if'
 		######################################
 		} elsif ($self->{current_line} =~ /^if\s/) {
-			my ($condition_text, $post_if) = $self->{current_line} =~ /^if\s+\(\s*(.*)\s*\)\s+(goto\s+.*|call\s+.*|stop|{|)\s*/;
+			my ($condition_text, $post_if) = $self->{current_line} =~ /^if\s+\(\s*(.*)\s*\)\s*(goto\s+.*|call\s+.*|stop|{|)\s*/;
 
 			debug "[eventMacro] Script is a 'if' condition.\n", "eventMacro", 3;
 			


Mortimal:
i really fking love openkore cause of it i can get gob of my dream,,, but u must have option to not put space after "if" condition....

plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -660,7 +660,7 @@ sub define_next_valid_command {
 		# Initial 'if'
 		######################################
 		} elsif ($self->{current_line} =~ /^if\s/) {
-			my ($condition_text, $post_if) = $self->{current_line} =~ /^if\s+\(\s*(.*)\s*\)\s+(goto\s+.*|call\s+.*|stop|{|)\s*/;
+			my ($condition_text, $post_if) = $self->{current_line} =~ /^if\s+\(\s*(.*)\s*\)\s*(goto\s+.*|call\s+.*|stop|{|)\s*/;
```
lututui: If were going to allow

if (x){
shouldn't we allow
if(x){
?

I mean like
`/^if\s*\(\s*(.*)\s*\)\s*(goto\s+.*|call\s+.*|stop|{|)\s*/`

Mortimal: no we need that for future to separate possible sub if.... from if block

lututui: But `/^if\s*\(\s*(.*)\s*\)\s*(goto\s+.*|call\s+.*|stop|{|)\s*/` wouldn't match sub ifsomething (note the `^`)

Mortimal: i mean if() <- function if () <- condition

Nipodemos:
my thought is: who that is sane enough to make code, will make a sub called `if`  ???
i think it will never have, i hope.
so i agree with @lututui

Nipodemos:
i'm closing this pull request because it does not fix the error completly, and it is even arguable if this is an error anyway

PR #1149, opened 2017-09-18
EventMacro if/elsif/while improved

Nipodemos:
hello,

this PR introduce two new functions:

## You can use just one argument
Now it will check if this argument is true of false
like:
```perl
$isTrue = 1
if ( $isTrue ) {
    #do something
}
```
whitespaces are ignored and doesn't gives any error.
Note that when you use only one argument, it will test if the value exists, or its value, just like in perl

## You can use `!` or `not` to invert the result
Just like in perl
Ex:
```perl
$isTrue = 1
if (! $isTrue ) { #you can use 'not' instead of '!' and work the same way
    #do something
} else {
    #it is negated, so it will be false and the else will be triggered
}

if ( not 1 > 10) {
    #it is negated, so it will be true
}  
```

Also, the error messages were improved, now it shows where you are wrong
for example, if you forget the last statment, it will print an error telling that you forgot the last
if you forget the condition, it will print an error tellign that you forget the condition
and so on.
this image demonstrate an error:
image

**Tested with this macro and worked perfectly:**
```perl
automacro testStatment {
	exclusive 1
	run-once 1
	BaseLevel > 0
	call {
		while (not $test) {
			log entered on while correctly
			$test = 1
		}
		
		if ( &defined($test) && 1 && ! 10 < 1 ) {
			log expected result1	
		} else {
			log this shoudn't appear1
		}

		if (! text = text ) {
			log this shoudn't appear2
		} else {
			log expected result2
		}

		if (!  1 <= 1 ) {
			log this shoudn't appear3
		} else {
			log expected result3
		}
		if ( $test || you != you || 1 ) {
			log expected result4
		} else {
			log this shoudn't appear4
		}
		switch ($test) {
			case (= 0) {
				log this shoudn't appear5
			}
			case (= 1) {
				log expected result5
			}
		}
	}
}
```

expected output:
```
[eventmacro log] entered on while correctly
[eventmacro log] expected result1
[eventmacro log] expected result2
[eventmacro log] expected result3
[eventmacro log] expected result4
[eventmacro log] expected result5
```

### DIFF
--- a/plugins/eventMacro/eventMacro/Runner.pm
+++ b/plugins/eventMacro/eventMacro/Runner.pm
@@ -1450,9 +1450,13 @@ sub parse_call {
 # use of [!/not] is optional, and will invert the result
 sub statement {
 	my ($self, $temp_multi) = @_;
-	my ($negate, $first) = $temp_multi =~ /^\s*(\!|not)?\s*"?(\S+)"?\s*/; #checks the first argument
-	my ($cond) = $temp_multi =~ /"?\S+"?\s*(==|=|<=|<|>=|>|!=|!|=~|~)/; #checks the condition
-	my ($last) = $temp_multi =~ /"?\S+"?\s*(?:==|=|<=|<|>=|>|!=|!|=~|~)\s*"?([^=~\s"]+)"?\s*$/; #checks the last argument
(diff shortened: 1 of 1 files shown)


Nipodemos:
i am closing my PR temporaly until henryBk have free time to evaluate them ( dont know when )

Issue #1190, opened 2017-09-27
[iRO Chaos] Error "Did not received the buy result from server after buy packet was sent."


userix:
Getting the following error: "Did not received the buy result from server after buy packet was sent."

everytime the bot goes to buy stuff from the NPC, but the purchase still goes through successfully.

Using the latest version of openkore.

sctnightcore:
Have an image?

userix:
![image](https://user-images.githubusercontent.com/20689010/30902090-e56f2a9e-a31e-11e7-89d6-5bf9bfc5bbd1.png)

userix:
![image](https://user-images.githubusercontent.com/20689010/30902102-ebd42d6c-a31e-11e7-8963-ffbe8ce36b5d.png)
This is a screenshot from the log file.  It already scrolled past the console history in openkore.

sctnightcore:
I think it should be normal because trading is complete.

userix:
Cool Thanks!  side question: is there a way to block the logging of certain NPC chat?  

Sep 26 22:45:00 2017][C] [prontera 116, 71] I am looking for the hero who returned from the expedition...

Everytime I walk by this NPC, it says this.

Glikadin:
![openstuck](https://user-images.githubusercontent.com/20753532/30915381-52110a52-a39f-11e7-8d6e-b9bd3feb803a.jpg)
Autobuy works very bad. I cant buy Nepeta Cataria from a cat paw seller and use macros:
```
...
	if (@invamount(Nepeta Cataria) < 6000) {
		do move 92 53 aldeba_in
		pause 1
		do talk 0
		pause 2
		do talk resp 0
		pause 1
		do buy 2 10000
		pause 2
		do talk no
		pause 2
	}
...
```

Other tool dealers give an error , but buing items possibly.

lordzpipo:
Using latest version from git, still got this problem, or i missing something?

@sctnightcore

bunnyy00:
![dc](https://user-images.githubusercontent.com/25070736/32448363-a7874fea-c349-11e7-8c9e-d0a60bddd855.png)

This is my auto-buy config:

######## Autostorage/autosell ########

buyAuto arrow {
	npc geffen_in 77 167 Tool Dealer
	standpoint
	distance 5
	price
	minAmount 80
	maxAmount 5000
	batchSize
	disabled 0
}

sellAuto 0
sellAuto_npc
sellAuto_standpoint
[...]

PR #1191, opened 2017-09-27
Fix talknpc seller npc interactions

Henrybk:
This probably fixes a few bugged npc interactions, I am pr it now so I will not forget do do it later since I made it in Aug 24.


sctnightcore:
#1190

Issue #1178, opened 2017-09-24
iRO Loki Crypt


suhery:
It seems they also apply renewal anti bot to classic. Timeout on map server. I have try to use renewal crypt key, copy shuffle.txt, sync.txt, and recvpacket.txt. It doesn't work.

c4c1n6kr3m1:
no , it isn't, yet
the ip is 128.241.92.203
try to separate classic table folder from iro table

kimer19:
how about in chaos ?

suhery:
Thanks for pointing it out c4c1n6kr3m1. I think we should move shuffle.txt and sync.txt to separated renewal folder.

kimer19:
Thanks for the reply , my bot is okay now i download the latest openkore

tftgigo:
@kimer19  Hi Sir, can you share what bot are you using?

sctnightcore:
download the latest openkore ! @tftgigo

kimer19:
​
 openkore-master (1).zip
<https://drive.google.com/file/d/0By4LyVpgcEYtVkRoTU1DY3lOTmM/view?usp=drive_web>
​

sctnightcore:
@kimer19  del dude !!!!
@tftgigo 
![image](https://user-images.githubusercontent.com/23263315/30803313-594f3b64-a213-11e7-9dc3-28a3c1923260.png)

suhery:
I solved this issue by creating blank shuffle.txt and sync.txt in Classic folder. This way renewal, classic, and restart can work correctly.

c4c1n6kr3m1:
@suhery Yes that would work too
I try different  approach on pull #1193

So with table like this
```addTableFolders iRO/Classic;iRO```
It will only search shuffle.txt on iRO/Classic folder, but not on iRO folder

PR #1193, opened 2017-09-27
IRO classic , shuffle.txt effect

c4c1n6kr3m1:
issue #1178; 
repair shuffle.txt effect


lututui:
Maybe a better solution would be to override shuffle in iRO Classic servertype? Or maybe a servers.txt key? I have no idea how reliable -e is.

c4c1n6kr3m1:
Make it  a key on servers.txt sounds good option
But i leave it to you

PR #1222, opened 2017-10-01
UP= sT20161207e / fix rev for all client 2015 

sctnightcore:
I'm doing If not lazy
If there is an unpack client, I will do it.
- [x] ST20161207e
- [x] FIX REV CLIENT 2015


alisonrag:
@allanon, @lututui  can some of you do the final approve?

sctnightcore:
Receive
I'm thinking of fixing qw of sakexe_0 so it reads from servertype0.

PR #1223, opened 2017-10-02
Items control.txt Substitue name for ID

Nipodemos:
done this change using this little amateur script: https://gist.github.com/Nipodemos/483a8f57e491a7ba59c668b855c0c890
(which gave me a lot more difficult to make than i though :roll_eyes: )

unfortunaly, it doesn't get all the ids, some i know the reason, others i dont
and i do not want to change manually the last two hundred items

### DIFF
--- a/control/items_control.txt
+++ b/control/items_control.txt
@@ -44,18 +44,18 @@ all 0 1 0
 ##### HEAL/SUPPORT ITEMS #####
 # Adjust these to your needs
 
-Butterfly Wing 10 1 0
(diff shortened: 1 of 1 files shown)


lututui:
It doesn't get everything because things like [2], [3] and [4] are added to the names during runtime

Nipodemos:
@lututui so my script would never work, and it is "simpler" to change manually instead of making another script

because it is missing "only" 150 items more or less

flacrasec:
Who will merge this? ^^

PR #1241, opened 2017-10-06
fix network fRO !

sctnightcore:
fRO !  bot work ! 
I'm wrong
```
Edit : Make everything smaller. For lazy people like me
FIX : npc talk on xkore3 
```
`my discord : sctnightcore#1063 `

xKORE 3 WORK!

### DIFF
--- a/src/Network/Receive/fRO.pm
+++ b/src/Network/Receive/fRO.pm
@@ -17,22 +17,20 @@ package Network::Receive::fRO;
 use strict;
 use base 'Network::Receive::ServerType0';
 
-use Globals qw( $messageSender );
-
 sub new {
 	my ($class) = @_;
 	my $self = $class->SUPER::new(@_);
-	my %packets = (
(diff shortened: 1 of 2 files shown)


src/Network/Receive/fRO.pm
```diff
@@ -15,7 +15,7 @@
 package Network::Receive::fRO;
 
 use strict;
-use base 'Network::Receive::ServerType0';
+use base qw(Network::Receive::ServerType0);
```
lututui: They're equivalent.

(But qw looks better, though)

src/Network/Send/fRO.pm
```diff
@@ -15,7 +15,7 @@
 package Network::Send::fRO;
 use strict;
 
-use base 'Network::Send::ServerType0';
+use base qw(Network::Send::ServerType0);
```
lututui: They're equivalent

(but qw looks better)

src/Network/Receive/fRO.pm
sctnightcore: i use with use base 'Network::Receive::ServerType0'; not work for me !

PR #1259, opened 2017-10-11
Add vending hook

Naozumi2k:
This will allow plugins to modify prices of selling items.


Naozumi2k:
This is an example of plugins - [ragial](https://github.com/Naozumi2k/openkore_ragial)
It currently hard coded for re:start server and needs improvement

PR #1263, opened 2017-10-12
IsNotInCoordinate Condition

Nipodemos:
created this Condition by simply copying the entire IsInCoordinate code and inverting the ValidateCondition 😄 

### IsNotInCoordinate
Description: Checks if the current coordinates **does not match** one of the given ones.

Syntax:
`IsNotInCoordinate <comma separated list of pairs of x and y coordinates>`

It also accepts ranges in place of coordinates. (like 100..200 for x and 50..100 for y)
Example:
```
IsNotInCoordinate 100 150
IsNotInCoordinate 120 100, 115 289, 158 236
IsNotInCoordinate 20..80 200, 50 60, 80 90..150, 100..200 150..200
IsNotInCoordinate 30..35 40..45
```
Set variables:
```
$.IsNotInCoordinateLast => Saves the last coordinate that made IsNotInCoordinate become true
$.IsNotInCoordinateLastMemberIndex => Saves the list index in condition that made IsNotInCoordinate become true
```

### DIFF
--- a/plugins/eventMacro/eventMacro/Condition/IsNotInCoordinate.pm
+++ b/plugins/eventMacro/eventMacro/Condition/IsNotInCoordinate.pm
@@ -1,0 +1,113 @@
+package eventMacro::Condition::IsNotInCoordinate;
+
+use strict;
+use Globals;
+use Utils;
+
+#Use: x1 y1, x2 y2, x3min..x3max y3, x4 y4min..y4max, x5min..x5max y5min..y5max
+
+use base 'eventMacro::Conditiontypes::MultipleValidatorState';
+
+sub _hooks {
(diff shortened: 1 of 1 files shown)


moneyisgod:
Amazing.
Retreat plugin done to macro.

I would think this is amazing.

Henrybk:
Shouldn't `and` assertions be made using more than one line? Like

```
IsNotInCoordinate 100 150
IsNotInCoordinate 120 100
```

Nipodemos:
you mean not to accept more than one pair of coordinates per condition??? @Henrybk

Henrybk:
Yep

Nipodemos:
this makes sense, i will change this today at night when i'm home

Nipodemos:
i turns out that this is much more difficult than i though, so i will close this until i learn how to do

PR #1276, opened 2017-10-16
plugin: eventMacro: fix issue in getQuestStatus

okeyb:
getQuestStatus reported complete on newly added quests.
Since there's no goal property in a newly added quest, getQuestStatus subroutine always returns complete on such quests until relogging. This patch addresses this bug.


fadreus:
After a day running this sure save me lot of time in 24 hours without having to relog.
Thanks a lot^^

fadreus:
Just realized after applying this fix non-existed quest is now return to incomplete rather than inactive (before this fix it return to complete).

At least it does fix the incomplete status.

fadreus:
#1420

fadreus:
Merge this?

fadreus:
Won't this get to merged?
Some players need to add this manually and some get confused with eventMacro status ain't working.

Henrybk:
Can you confirm this is working?

fadreus:
Yup, this fixed the addressed bug.

PR #1286, opened 2017-10-19
fRO tables update ! ♥

sctnightcore:
OK! ♥♥
fRO has too many new maps. LOL 
I think it would make the file openkore too big.

new map : 
0de9da0


lututui:
> fRO has too many new maps. LOL

You can compress them like #1292

PR #1287, opened 2017-10-19
xkore 2 fix male character in iRO

alisonrag:
Fix for male char and unknow map in iRO/bRO

Character Server:
xkore 2 iro char server

Map Server:
xkore 2 iro map server

### DIFF
--- a/src/Base/Ragnarok/CharServer.pm
+++ b/src/Base/Ragnarok/CharServer.pm
@@ -230,8 +230,14 @@ sub sendCharInfo {
 		my $index = -1;
 		foreach my $char ($self->getCharacters($session)) {
 			$index++;
+			
+			my $sex = "";
+			my $map = $field->baseName.".gat";
+			if($masterServer->{charBlockSize} == 145 && $masterServer->{serverType} =~ /^iRO/) {			
+				$sex = $char->{sex};
+			}
+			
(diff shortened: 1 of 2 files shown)


alisonrag:
@allanon 
to openkore work with xkore 2 in iro we need to add this in iRO.pm
```
my %handlers = qw(
		received_characters 099D
		received_characters_info 082D
		sync_received_characters 09A0
	);

	$self->{packet_lut}{$_} = $handlers{$_} for keys %handlers;
```
because of this:
src\base\ragnarok\CharServer.pm line 97
`if ($self->{recvPacketParser}{packet_lut}{received_characters} eq '099D') {`

lututui:
Also fixes "unknown map" in bRO

src/Base/Ragnarok/CharServer.pm
```diff
@@ -230,8 +230,14 @@ sub sendCharInfo {
 		my $index = -1;
 		foreach my $char ($self->getCharacters($session)) {
 			$index++;
+			
+			my $sex = "";
+			my $map = $field->baseName.".gat";
```
lututui: Why use a new var when using it only once?

lututui: = "" is redundant

PR #1315, opened 2017-10-26
[eventMacro][bugfix] whitespace support for hashes

Nipodemos:
found a little bug while i was wrinting a new macro:
### the macro:
```perl
automacro defineHash {
	BaseLevel > 0
	run-once 1
	exclusive 1
	disabled 0
	call {
		log creating hashes values and keys
		$hash{noSpace} = 1
		log created noSpace
		$hash{ has Space } = 1
		log created has Space
		$hash{ noSpaceAgain } = 1
		log created noSpaceAgain
		$hash{has More Space} = 1
		log created has More Space
	}
}
```
### the bug:
bug fix before

with these new changes it works normally:
### working pic:
bug fix after

### DIFF
--- a/plugins/eventMacro/eventMacro/Utilities.pm
+++ b/plugins/eventMacro/eventMacro/Utilities.pm
@@ -529,6 +529,8 @@ sub find_accessed_variable {
 		my $name = $1;
 		my $open_bracket = $2;
 		my $complement = $3;
+		$complement =~ s/^\s+//;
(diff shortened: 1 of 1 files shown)


Nipodemos:
turns out that was not a bug, is just that perl doesn't accept whitespaces between words, so eventMacro also should not accept

and i've found another way to use without whitespaces

so, closing this

PR #1343, opened 2017-11-01
[items_control.txt] substitue more names by its ID's

Nipodemos:
converted the items to ID that were missing on #1223 

now is quite complete 😃 

i think it would be good to add a `s/#.+//;` to fileParser of items_control, but so far seems to have no problem

### DIFF
--- a/control/items_control.txt
+++ b/control/items_control.txt
@@ -54,7 +54,6 @@ all 0 1 0
 506 0 1 0 # Green Potion
 501 25 1 0 # Red Potion
 502 25 1 0 # Orange Potion
-504 25 1 0 # White Potion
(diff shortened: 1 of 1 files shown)


control/items_control.txt
```diff
@@ -54,7 +54,6 @@ all 0 1 0
 506 0 1 0 # Green Potion
 501 25 1 0 # Red Potion
 502 25 1 0 # Orange Potion
-504 25 1 0 # White Potion
```
alisonrag: whit potion is missing

Nipodemos: oh dear for some random reason it was duplicated, and i don't know how i've deleted both of them

today at night i will change this

PR #1345, opened 2017-11-01
Pet

c4c1n6kr3m1:
$pet{ID} always undef
it's reverse on iRO
that's why Wx pet doesn't work
need test for other servers


alisonrag:
will test soon

alisonrag:
@c4c1n6kr3m1 
can we merge this?

send us some screenshots

src/Network/Receive/ServerType0.pm
```diff
@@ -2020,6 +2020,7 @@ sub item_skill {
 sub map_changed {
 	my ($self, $args) = @_;
 	$net->setState(4);
+	%pet = ();
```
Nipodemos: hey, i might saying nonsense...

but why this needs to stay here? i mean, is already declared at Globals.pm isn't?

Nipodemos: oh now i understand, is to clear old values right? sorry for misunderstanding

lututui: Also this should be in functions.pl, not here

c4c1n6kr3m1: @lututui , the pet ID changed every map server , iRO

my current bot date years ago, work with addition only on functions.pl
but not with our latest bot, so i need to put it on map_changed,looks like something changed

Nipodemos: But when map changes the method on functions.pl is always called... Or am I wrong?

PR #1346, opened 2017-11-01
iRO Renewal packet update

allanon:


brenybear:
hi Allanon. every patch update i also see you making updates here, wanna take this chance to tell you Thank you so much, appreciate your works sir. more power to you!

PR #1364, opened 2017-11-06
Update Misc.pm

Arkelsa:
Insertion of the Wildcard code to allow pickupitems.txt to use Registry Extensions

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -2306,9 +2306,21 @@ sub mon_control {
 # Returns the pickupitems.txt settings for item name $name.
 # If $name has no specific settings, use 'all'.
 sub pickupitems {
-	my ($name) = @_;
+   my $name = lc shift;
 
-	return ($pickupitems{lc($name)} ne '') ? $pickupitems{lc($name)} : $pickupitems{all};
+   return $pickupitems{$name} if exists $pickupitems{$name} and $pickupitems{$name} ne '';
+   
+   my $result = $pickupitems{all} || {};
+   
+   foreach my $key (keys %pickupitems) {
+      next unless $key =~ m!^/(.*)/$!;
+      my $regex = $1;
+      if ($name =~ /$regex/) {
+         $result = $pickupitems{$key};
+      }
+   }
+   
+   return $pickupitems{$name} = $result;
 }
 
 sub positionNearPlayer {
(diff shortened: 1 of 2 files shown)


src/Misc.pm
```diff
@@ -2306,9 +2306,21 @@ sub mon_control {
...
-	my ($name) = @_;
-
-	return ($pickupitems{lc($name)} ne '') ? $pickupitems{lc($name)} : $pickupitems{all};
+   my $name = lc shift;
+
+   return $pickupitems{$name} if exists $pickupitems{$name} and $pickupitems{$name} ne '';
+   
+   my $result = $pickupitems{all} || {};
+   
+   foreach my $key (keys %pickupitems) {
```
allanon: This will iterate over every item in `%pickupitems`, every time an item is considered for picking up. Compare with the hash lookup we were doing before (and are still doing for non-regex lookups), which has constant-time cost.

We should separate regexes from non-regexes in the parser, and only look at the regexes in this logic.

For comparison, here's my implementation:

```Perl
sub pickupitems {
    my $name = lc shift;
    my $pui  = $pickupitems{$name};
    $pui = ( map { $pickupitems_r{$_} } grep { $name =~ /$_/ } keys %pickupitems_r )[0] if !defined $pui;
    $pui = $pickupitems{all} if !defined $pui;
    $pui;
}
```

It uses a separate hash (`%pickupitems_r`) which only contains the regex-based rules.

allanon:
Good idea! I'm sure a lot of people would love to have this feature.

A few suggestions:

I recommend separating the regexes from the literal matches. We can do this by surrounding the rule in `pickupitems.txt` with slashes, which are commonly used to denote a regex. For example:

```
# Regex-based rule: Pick up all cards fast.
/.* Card/ 2

# Literal match rule: Pick up good armor.
Chain [1] 1
```

When this gets merged, the only metadata from this pull request which goes into the master branch is the commit message. Please try to make commit messages which are meaningful. For example: `regex-based wildcard support in pickupitems.txt`.

tables/portals.txt
```diff
@@ -10,7 +10,7 @@ alberta 245 93 amatsu 197 83 10000 c r1 c c c r0 n
 amatsu 194 79 alberta 243 91 0 c r0 n
 alberta 245 69 gon_fild01 258 82 0 c r1 c c r0 n
 gon_fild01 255 79 alberta 243 67 0 c r0 n
-izlude 201 181 izlu2dun 107 50 150 c r0
+izlude 197 205 izlu2dun 107 50 150 c r0
```
sctnightcore: what server you test ?

lututui:
Implementation with an awful performance impact

PR #1375, opened 2017-11-10
[bRO] Fix getting stuck in moc_fild20

lututui:
As reported by RafaelO

There's like ~50 bots stuck in moc_fild20 it's really funny


RafaelOds:
Bug's arrumados com sucesso, esta rodando lisinho agora!!

PR #1376, opened 2017-11-10
xconf on Commands.pm

Nipodemos:
make

**iconf
mconf
pconf
sconf**

integrated on openkore

i did not made that code, although i've edited a little bit, to work perfectly

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -1439,7 +1439,7 @@ sub cmdConf {
 
 sub cmdxConf {
 	my ($cmd, $args) = @_;
-	my ($file,$file2,$found,$key,$oldval,$shopname,$type,$value, $name, %inf_hash, %ctrl_hash);
+	my ($file,$file2,$found,$key,$oldval,$shopname,$type,$value, $name, $inf_hash, $ctrl_hash);
 	if ($cmd eq 'sconf') {
 		($key, $value) = $args =~ /(name)(?:\s)(.*)/;
 		$shopname = 1 if $key eq 'name';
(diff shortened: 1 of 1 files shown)


oleokoong:
this mean dont need the plugins anymore? merge into src?

src/Commands.pm
```diff
@@ -1431,6 +1437,255 @@ sub cmdConf {
...
+	debug "extracted from args: KEY: $key, VALUE: $value\n";
+	if (!$key) {
+		error "Syntax Error in function '$cmd'. Not found <key>\n".
+				"Usage: $cmd <key> [<value>]\n";
+		return;
+	}
+	if ($cmd eq 'iconf') {
+		%inf_hash  = %items_lut;
+		%ctrl_hash = %items_control;
+		$file = 'items_control.txt';
```
lututui: You're not respecting command line arguments, if they're passed. This is very bad.

src/Commands.pm
```diff
@@ -1431,6 +1437,255 @@ sub cmdConf {
...
+	$key = $args if !$key;
+	$key =~ s/^\s+|\s+$//g;
+	debug "extracted from args: KEY: $key, VALUE: $value\n";
+	if (!$key) {
+		error "Syntax Error in function '$cmd'. Not found <key>\n".
+				"Usage: $cmd <key> [<value>]\n";
+		return;
+	}
+	if ($cmd eq 'iconf') {
+		%inf_hash  = %items_lut;
```
lututui: Instead of copying the hash, maybe you should instead use a reference.

$inf_hash_ref = \%items_lut

src/Commands.pm
```diff
@@ -1431,6 +1437,255 @@ sub cmdConf {
...
+		return;
+	} elsif (($key eq "setall")) { ## If $key is "setall" setting all keys in file to $key
+		filesetall($file, $value);
+		return;
+	}
+	## Check $key in tables\monsters.txt & tables\items.txt
+	if ($key ne "all") {
+
+		#key is an ID, have to find the name of the item/monster
+		if ($inf_hash{$key}) {
```
lututui: Continuing the argument of using references, you would access this as `${$inf_hash}{$key}`

src/Commands.pm
```diff
@@ -1431,6 +1437,255 @@ sub cmdConf {
...
+			$found = 1;
+			if ($cmd eq 'iconf' && $itemSlotCount_lut{$key}) {
+				$name = $inf_hash{$key}." [".$itemSlotCount_lut{$key}."]";
+			} else {
+				$name = $inf_hash{$key};
+			}
+
+		#key is a name, have to find ID of the item/monster
+		} else {
+			foreach (values %inf_hash) {
```
lututui: Also, this would be accessed as `%{$inf_hash}`

Nipodemos:
i'll modify these things as soon as possible , thanks for feedback :smile:

Mortimal:
xconf should be left as plugin +1 (add updates there, if u have good one). Duno if it needed to be loaded by default (i think not).
Reason: xconf is advance plugin used mainly in macroses

Nipodemos:
@Mortimal i think is valid, it will not make kore slower , and it will add a very handy function to it.

in the same way that is very good **conf** command to modify config.txt , **iconf** is equally good (and yes, it is awesome for eventMacros 😃 )

Nipodemos:
as i've speaked earlier with lututui, xconf already get the filename of command line arguments, so i just changed the hashes to references
tested here and worked fine

alisonrag:
we have a conf command, why not iconf, mconf, etc?

this command will not affect the openkore performance

Mortimal:
ok...

Nipodemos:
just closing this, and waiting for the other one to be merged 😄

PR #1380, opened 2017-11-11
iRO Conflict fix in recvpackets/shuffle/sync/portals

### LINKED ISSUE #184: iRO classic moc_fild10
iRO classic uses an old version of morroc maps, when on lockMap moc_fild10 ai can't route to lockMap whenever it's outside the map.

### LINKED ISSUE #1379: iro classic not connecting
timeout on map and account server.

old version working fine version ctime:2017_11_08 but not working with latest verison

alisonrag:
fix conflict in load file from other iRO server

kore is loading the renewal server files when you select classic server.

This closes #1379, closes #184


tables/servers.txt
```diff
@@ -94,7 +94,7 @@ serverType iRO
 serverEncoding Western
 patchserver ropatch1.gravityus.com
 patchpath /patch02
-addTableFolders iRO
+addTableFolders iRO/Renewal;iRO
```
sctnightcore: iRO RE and PRE 
Use with the same client?

alisonrag: @sctnightcore 
i dont know, maybe not, because classic doesn't use shuffle or sync

sctnightcore: @allanon  @alisonrag 
I'm sure it will use the same client.

lututui:
Is this still relevant after #1404?

alisonrag:
In my point of view every single server need a new tree.
why classic and restart have and renewal not?

alisonrag:
Now we are mixing files without know how it will affect other server

hougochia:
perhaps we should have a master file for every servers , plugin separation and config files that use on different server

PR #1391, opened 2017-11-14
fix crash when test hp of monster

### LINKED ISSUE #1300: Illegal Division By Zero when calculating target HP before using skill
###### Affected configuration(s)/ file(s): Misc.pm :: sub checkMonsterCondition() (Line 4355)
```
	if ($config{$prefix . "_hp"}) {
		return 0 if (!$monster->{hp});
		if ($config{$prefix . "_hp"} =~ /^(.*)\%$/) {
			return 0 unless (inRange(($monster->{hp} * 100 / $monster->{hp_max}), $1)); 
		} else {
			return 0 unless (inRange($monster->{hp}, $config{$prefix . "_hp"})); 
		}
	}
```
###### Steps to Reproduce:
attackSkillSlot <skill_name> { target_hp >25% }

I used Sonic Blow with target_hp >25%
Occasionally, $monster->{hp_max} is zero for some reason, causing the crash. A simple fix would be to add zero-division check, where a failure will result to skill not used instead of crashing the bot.

alisonrag:
fix crash when do Illegal Division By Zero in calculate target HP before use skill

this closes #1300


src/Misc.pm
```diff
@@ -4353,14 +4353,28 @@ sub checkMonsterCondition {
...
+					return 0;
+				}
+			} elsif($config{$prefix . "_hp"} =~ /(\d+)$/) {
+				if($monster->{hp} && $monster->{hp_max}) {
+					return 0 unless (inRange($monster->{hp}, $config{$prefix . "_hp"}));
+				} else {
+					return 0;
+				}
+			} else {
+				warning TF("%s is wrong configured.\n", $prefix . "_hp");
```
lututui: We've never warned about invalid config blocks, I don't think we should start now, at least not in broken English

PR #1399, opened 2017-11-16
xconf plugin bugfix

Nipodemos:
while it is decided if xconf will be merged to `Commands.pm` or not

fix this little bug, that when you use iconf or mconf  or pconf, and the `key` had utf8 characters, it crashes the openkore

by the way, this plugin needs update, it is not prepared to work with IDs on items_control.txt 
(in #1376 i've already updated)


Nipodemos:
thank you :smile:

Issue #1401, opened 2017-11-16
XKore 1, 2, and 3 DC on Map Server- iRO- Renewal Thor


vhonn:
I just came back to play after a year of hibernation, Just need to know if there is any way I can Run Xkore 1 on iRO Thor. As of the moment I can run on XKore 0 smoothly but when I go Xkore 1 or 3 it wont connect on Map server or being disconnected right after. Any Help. TYIA.

Plus NPC locations are messed up on iRO Renewal Thor.

sctnightcore:
**Please use the issue template, please post infornations about the server and your configuration.**

Wait @allanon

fadreus:
@vhonn Have you try #1374 ?
This PR hasn't been merged to master, yet.

alisonrag:
use #1374 
and if the renewal update wait for the update.

one more thing:
XKore is less tolerant to incorrect server settings. If everything works without XKore and there is some problem with XKore, it may still be a problem with server settings like serverType, charBlockSize etc.

vhonn:
so basically this is just copy/paste overwrite files I assume because there is 5 parts of this update [#1374](https://github.com/OpenKore/openkore/pull/1374) 
1. **_fix xkore 1 and 3_** - contains  .gitignore, reconnect.pl, Receive.pm ,ServerType0.pm, Xkore.pm and 
    XKoreProxy.pm
2. **_remove not used variable_** - contains update for reconnect.pl file
3.  **_missing import_** - update to XKore.pm file (overwrites XKore.pm file on the 1st part of the update).
4. **_remove portal LOS_**  - not applicable
5. **_xkore 1 reset key when chang mapserver_** - update to XKore.pm file (overwrites XKore.pm file on the 1st and 3rd part of the update).

alisonrag:
@vhonn 
just access this link:
https://github.com/OpenKore/openkore/pull/1374/files
this files is the latest version in pull request

vhonn:
XKore 1 works perfectly for iRO -Thor Server. One issue I bump in is being DC on map server while or after doing autostorage and so disabled autostorage for the time being.

alisonrag:
so this is with servertype file =p

Issue #1397, opened 2017-11-15
openkore does not move.


gr4ndmasterz:
apparently after downloading the latest openkore.

the bot wont move and just stays in 1 place.

Bombahee:
mine, getting stuck in town, stuck during route

Gulxx:
stuck during route openkore  i have

Nipodemos:
Type a command when this happen:
**reload portals **

Gulxx:
not works reload portals

davinci1986:
Me too. Keep like this after re-download the new Openkopre

Server: iRO-Restart: Odin

> Calculating random route to: Mushroom Farm (job_thief1): 66, 178
> Stuck at job_thief1 (61,52), while walking from (61,52) to (66,178).
> Stuck during route.
 Any idea?

oleokoong:
the packet for move is broke, and the devs who did the shuffler will look for it in next 9 hour..because he busy on real life things

akositek:
Stuck at payon (258,241), while walking from (258,241) to (233,304).
Stuck during route.
Calculating lockMap route to: Payon Cave(pay_dun03)
Stuck at payon (258,241), while walking from (258,241) to (228,329).
Stuck during route.
Calculating lockMap route to: Payon Cave(pay_dun03)

I got this issue guys. please help ..

c4c1n6kr3m1:
be patient

null32sys:
Openkore Log

Sent sendMapLogin
Received packet: 0283 Handler: account_id

And can't load map login

I've been got this on map server tcp port 5000 through wireshark (C Array)

Send
0x72, 0x00, 0xa7, 0x28, 0x03, 0x00, 0xb7, 0xc5, 
0x07, 0x00, 0xa3, 0x2e, 0x00, 0x00, 0x54, 0x60, 
0x06, 0x00, 0x01 };

Recv
char peer1_1[] = { /* Packet 49 */
char peer1_2[] = { /* Packet 50 */
0x83, 0x02, 0xa7, 0x28, 0x03, 0x00 };
char peer1_3[] = { /* Packet 51 */
0xeb, 0x02, 0x4b, 0xb4, 0x07, 0x1f, 0x38, 0x4d, 
0x60, 0x05, 0x05, 0x00, 0x00 };
char peer1_4[] = { /* Packet 53 */
0xb0, 0x00, 0x31, 0x00, 0x2b, 0x01, 0x00, 0x00, 
[...]

PR #1404, opened 2017-11-17
fix 1397: ignore shuffles of packets which are not the version the cu…

allanon:
…rrent server uses

- [ ] QA (does it work?)


sctnightcore:
@allanon  wait me !

sctnightcore:
#1401
#1397

PR #1410, opened 2017-11-17
iRO ReStart fields extract from GRF.

BottlerPB:
ReStart use some differente maps. So i compare tem openkore fields with the ReStart GRF, and create this list.


alisonrag:
@sctnightcore 
compared with sha-256, and compacted

PR #1414, opened 2017-11-18
Fix segfault on initialization of Wx interface

zstyblik:


zstyblik:
I have tested this on Debian 9.2 Stretch. Without this patch, % perl openkore.pl --interface=Wx; segfaults immediately. Sorry, but I cannot test other platforms as I own none of them.

zstyblik:
Thanks guys :)

PR #1418, opened 2017-11-19
Prevent party info loss

lututui:
Sometimes we receive (or process?) party_users_info before party_join when logging in. This doesn't happen always, but when it does, it breaks Slave bots. This was noticed in bRO Thor

This is to prevent clearing info already in $char->{party}


lututui:
Fixes this, which happened in iRO Restart 
http://openkorebrasil.org/index.php?/topic/4181-sem-conseguir-buffar-e-curar-mais-slave/#entry30967

PR #1440, opened 2017-11-25
Fix sT_private server 

sctnightcore:
```
1. Fix sT miss and New sT
2. Fix sT qw sT0
```


src/Network/Send/kRO/Sakexe_2004_07_05a.pm
```diff
@@ -18,7 +18,7 @@
 package Network::Send::kRO::Sakexe_2004_07_05a;
 
 use strict;
-use base qw(Network::Send::kRO::Sakexe_0);
+use base qw(Network::Send::ServerType0);
```
lututui: I don't think you should be doing this

lututui:
I don't think we should be ditching Sakexe0

src/Network/Send/kRO/Sakexe_2004_07_05a.pm
sctnightcore: @lututui  I'm lazy to update the sakexe_0 part. It's missing too many packets.
Update just servertybe0 it would be better, right?

sctnightcore:
@lututui  I have moved to kRO> backup> sakexe_0

src/Network/Send/kRO/Sakexe_2004_07_05a.pm
lututui: You don't know what this implies. All RagexeRE and Sakexe would become ST0 children.

sctnightcore: I think that's good too.

sctnightcore: Does recv use qw sT0 and send using Sakexe_0? @lututui

PR #1456, opened 2017-11-29
Fields extracted from iRO Re:Start GRF, compared with sha-256 and compacted.

BottlerPB:
Same files from: https://github.com/OpenKore/openkore/pull/1410 

For consistency purpose, movede from fields\iRO Restart to fields\iRO\Restart.


lututui:
:)

Issue #1484, opened 2017-12-07
RE:Start - After server maintenance - Timeout on Map Server, connecting to Account Server...


0Noro:
Please refer to the following resources for help using OpenKore:
 faqs - http://openkore.com/index.php/Frequently_Asked_Questions (such as "Packet Tokenizer")
 forum - http://forums.openkore.com
 documentation wiki - http://openkore.com/index.php/Main_Page
 download new openkore - https://github.com/OpenKore/openkore/archive/master.zip
* Before submitting an issue, ensure you are on the latest release
* Check if an issue already exists for your request
* Please post in English if possible
* Please use the issue template, please post infornations about the server and your configuration

###### Affected configuration(s)/ file(s):
```Include relevant code snippets and source code references```
###### Affected configuration(s)/ file(s):
```Include relevant code snippets and source code references```
###### Additional Details:
[Ragexe.zip](https://github.com/OpenKore/openkore/files/1537551/Ragexe.zip)

Pachu1985:
you will have to wait for someone to update the packets after the maintenance.

is there any noob friendly guide to do this? i have free time and i want to try doing it

xaerus88:
i am new to openkore, can someone please help me, i have the same problem as well/.

gr4ndmasterz:
dev are aware, give them time, they are contributing without expecting anythin in return , so just wait, everytime after maintenance this will happen. so relax

xaerus88:
thanks. =)

johnleydelgado:
They will closed this issue if they fixed it

EnzoMachiavelli:
Can someone link me allanon's packet generator?

sctnightcore:
https://gist.github.com/allanon/d8181ceaf6380da6d8fa4fddf1885706

Glikadin:
Okay . How get and install disassemble x86.pm ?  I am noob
![125125ssssss](https://user-images.githubusercontent.com/20753532/33706633-c2f91cb8-db45-11e7-9f5b-050b260f06e4.jpg)
xD

xaerus88:
Is it fix now? thanks. =)

EnzoMachiavelli:
@sctnightcore these are the crypt keys for restart. now I'm clueless how to get the rest

sctnightcore:
@EnzoMachiavelli  key is true ! but problem is recvpacket and shuffle

EnzoMachiavelli:
[...]

PR #1486, opened 2017-12-07
iRO Restart packet update

allanon:


lututui:
Fixes #1484

xaerus88:
do I just redownload the whole thing and it would work? thanks

PR #1510, opened 2017-12-15
Removing teleport that doesn't exist on bRO

Nipodemos:


lututui:
WHERE DID THAT COME FROM GEEZ

PR #1555, opened 2017-12-27
use xConf as one of default's plugins

Nipodemos:
and also updated it to work properly when it find ID's
instead of inserting into Commands.pm like the other PR ( #1376 ) now it keeps as a plugin, but it is activated by default on sys.txt


lututui:
Move xconf.pl from plugins/ to plugins/xconf/ (give the plugin its own folder)

Nipodemos:
okay

Nipodemos:
done

PR #1601, opened 2018-01-05
Fix RMS test server

lututui:
Fix connection with servers RMS Renewal Test Server and RMS Pre-Re Test Server


Nipodemos:
image
working !!! thank you @lututui

PR #1605, opened 2018-01-07
Add player condition target_spirit

lututui:


Mortimal:
Need not to forget and add it to wiki...

lututui:
http://openkore.com/index.php/Target_spirit

PR #1613, opened 2018-01-08
Add storageAuto_useItem

lututui:
Allows opening storage with a consumable item, like [Kafra Card](http://ratemyserver.net/index.php?page=item_db&item_id=12211)

Usage:
```
storageAuto_useItem 1
storageAuto_useItem_item Kafra Card
```

* If you run out of Kafra Cards and storageAuto_npc is set, kore will do execute the normal storageAuto sequence. If storageAuto_npc is not set, storageAuto will be disabled.
* storageAuto_useChatCommand has priority over storageAuto_useItem.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -713,6 +713,8 @@ storageAuto_password
 storageEncryptKey
 storageAuto_keepOpen 0
 storageAuto_useChatCommand
+storageAuto_useItem 0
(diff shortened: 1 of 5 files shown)


dcyxxt:
can add a talk step by "storageAuto_useItem_item"
because some private service has vip card to open storage,but it need a talk step

PR #1615, opened 2018-01-08
fix pServer 

sctnightcore:
item_list_res to item_list_window_selected


lututui:
Forgot to rename it in %handlers

PR #1622, opened 2018-01-10
Fix loop on buyAuto

### LINKED ISSUE #1587: Buying Yellow Potions make kore disconnect
###### Summary: 
yellow

###### Affected configuration(s)/ file(s):
Disconnect if it was like this
`
buyAuto Yellow Potion {
	maxAmount 100
	minAmount 0
	npc 
	distance 8
}
`

Works if it was like this
`
buyAuto Yellow potion {
	maxAmount 100
	minAmount 0
	npc 
	distance 8
}
`
###### Steps to Reproduce:
autobuy

lututui:
buyAuto is case insensitive, but if we try to buy, for instance, 'Yellow potion' and the store sells 'Yellow Potion' (mind the uppercase and lowercase 'P'), the item is bought -- but we enter a loop, because we still don't have enough of 'Yellow potion', even though we have enough of 'Yellow Potion'. This commit fixes that behaviour.

Also fixes #1587 

Expected behaviour (with this commit applied):
2018-01-10_11h32_06

What happens without this commit. Notice we don't actually buy the stuff, but we keep trying.
2018-01-10_11h40_01

### DIFF
--- a/src/InventoryList.pm
+++ b/src/InventoryList.pm
@@ -395,7 +395,7 @@ sub sumByName {
 	assert(defined $name) if DEBUG;
 	my $sum = 0;
 	for my $item (@$self) {
-		if ($item->{name} eq $name) {
+		if (lc($item->{name}) eq lc($name)) {
 			$sum = $sum + $item->{amount};
 		}
 	}


iamtonysoft:
confirmed it works!

lututui:
Fixes #1587

PR #1628, opened 2018-01-11
add config settings for missing portals

Henrybk:
Title
Now users will be able to activate and de-activate these features using config keys


Nipodemos:
don't forget to put these config keys on control-br

Nipodemos:
wow
we can close like, a bunch of issues now hahaha

lututui:
@Nipodemos we can't assume it's all fixed

PR #1630, opened 2018-01-11
Add [NONE] to equipAuto

lututui:
Using <slot> [NONE] in equipAuto now unequips that slot.

For example:
```
equipAuto {
	midHead [NONE]
	onAction sitAuto
}
```

Unequips whatever is equipped in midHead when sitAuto.


Nipodemos:
that's beautiful, thank you for update

PR #1637, opened 2018-01-12
Update ISSUE_TEMPLATE.md

sctnightcore:
If it's short, I think people will use it.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,35 +1,9 @@
-Please refer to the following resources for help using OpenKore:
- faqs - http://openkore.com/index.php/Frequently_Asked_Questions (such as "Packet Tokenizer")
- forum - http://forums.openkore.com
- documentation wiki - http://openkore.com/index.php/Main_Page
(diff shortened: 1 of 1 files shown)


Nipodemos:
this seems very good, shorter, simpler and equally efficient as the old one
maybe better

alisonrag:
i think we should keep:
´´´
-* Before submitting an issue, ensure you are on the latest release
 -* Check if an issue already exists for your request
 -* Please post in English if possible
´´´

Fadreus:
Agree with alison.

I think this is unnecessary
`* **Topic**:` 
Topic = Issue Title

How about put a little note/warning? Such as
-DON'T DELETE THIS. FILL THIS FORM.
-Use this template or your issue will be closed.

wei:
Alternatively, you can use [wei/new-issue](https://github.com/wei/new-issue) : D

Nipodemos:
i must agree that this we/new-issue seems to be a awnsome thing :smile:

Nipodemos:
what do you guys think now? @Fadreus @alisonrag @sctnightcore ??
i've updated

alisonrag:
ok to me

PR #1657, opened 2018-01-16
xConf changes

lututui:
Fix "set to" message
Prioritize already existing config
Use parseReload instead of running reload command
Support item ID (pconf)
priconf: respect !include


Nipodemos:
nice :smile:

ReedemBR:
**Excelent !**

iamtonysoft:
thank you for this...
i love you <3

PR #1662, opened 2018-01-17
Add whenPartyMembersNear and whenPartyMembersNearDist

lututui:
whenPartyMembersNear : range of party members nearby necessary to trigger this block
whenPartyMembersNearDist : range of distance those party members have to be to trigger this block. If unset, defaults to less than removeActorWithDistance or clientSight or 15 (in this order)

Both are self coditions, therefore can be used in partySkill, useSelfSkill, useSelf_item, buyAuto, etc.

Allows users to do things like this:

Use Clementia if there's more than two party members nearby
Use Blessing otherwise

```
partySkill Clementia {
    target_whenStatusInactive Blessing
    whenPartyMembersNear > 2
    whenPartyMembersNearDist <= 7
}

partySkill Blessing {
    target_whenStatusInactive Blessing
    whenPartyMembersNear 2
}
```

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -4243,8 +4243,36 @@ sub checkSelfCondition {
 		return 0 if ($field->getBlock($pos->{x}, $pos->{y}) != Field::WALKABLE_WATER);
 	}
 	
-	if (defined $config{$prefix.'_devotees'}) {
+	if ($config{$prefix.'_devotees'}) {
 		return 0 unless inRange(scalar keys %{$devotionList->{$accountID}{targetIDs}}, $config{$prefix.'_devotees'});
+	}
+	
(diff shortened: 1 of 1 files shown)


fadreus:
Doesn't need to update config.txt _(Wiki too?)_?
Other user might not know this _(unless they search for it)_.

lututui:
Documentation:

* whenPartyMembersNear
http://openkore.com/index.php/WhenPartyMembersNear
https://openkorebrasilwiki.miraheze.org/wiki/WhenPartyMembersNear

* whenPartyMembersNearDist
http://openkore.com/index.php/WhenPartyMembersNearDist
https://openkorebrasilwiki.miraheze.org/wiki/WhenPartyMembersNearDist

alisonrag:
these keys are missing in config.txt

PR #1670, opened 2018-01-18
Fix pickupitems

lututui:


lututui:
@Nipodemos no, I just screwed up the pickupitems when adding nameID support xD

PR #1696, opened 2018-01-24
Add 'disabled 0' to default getAuto block

lututui:


Nipodemos:
hey @lututui there is something that i also miss:
the `zeny` inside the defalut buyAuto block.

could you add this too? or i can make a PR

PR #1717, opened 2018-01-30
checking weight when get item via console commands

Nipodemos:
This PR do:

* if the target is cart or inventory, will check weight of each item (and only if has information about weight)
* calc how much of the current item the bot can get, if the amount is bigger that he can carry, will get just enough to stay with 99% weight (or 100%)

```
-- Não utilizável --
13   Sangue de Lobo x 45
14   Gelo x 3507 #this is the item will get
16   Pedra x 1
17   Ombreira x 2

Capacidade: 27/600
--------------------------------------------------
storage get 14 #command to get the item
Amount of item Gelo (14) is more that you can carry, getting the maximum possible (2099)
Item adicionado ao inventário: Gelo (23) x 2099 - Event
Você está agora: Owg 90%
Item retirado do armazém: Gelo (14) x 2099
```

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1041,6 +1041,25 @@ sub processTransferItems {
 			redo;
 		}
 
+		#verify if we can carry the amount of item
+		#if you're sending to storage, doesn't need to check weight
+		#if there is no information about weight, there's no point in check
+		if ($row->{target} ne 'storage' && $item->weight()) {
+            my $freeWeight;
+            if ($row->{target} eq 'inventory') {
+                $freeWeight = int ($char->{weight_max} - $char->{weight});
+            } else { #if target not inventory, then is cart!
+                $freeWeight = $char->cart->{weight_max} - $char->cart->{weight};
+            }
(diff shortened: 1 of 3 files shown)


Nipodemos:
tested with other items, such as arrows that have 0.1 weight and is working well
changed the string as suggested by @lututui

c4c1n6kr3m1:
server sent item's weight?

src/AI/CoreLogic.pm
```diff
@@ -1041,6 +1041,25 @@ sub processTransferItems {
 			redo;
 		}
 
+		#verify if we can carry the amount of item
+		#if you're sending to storage, doesn't need to check weight
+		#if there is no information about weight, there's no point in check
```
sctnightcore: It is compatible with item_weights it?

Nipodemos: actually, is only possible because of this implementation of allanon
i am just make good use of item_weights.txt :smile: 
i am using item_weights.txt, and if don't have the value, it will simply skip this piece of code and keep going

PR #1719, opened 2018-01-30
add 0A43 , 0A44

sctnightcore:
It can only be used temporarily.

Need back to fix allsT2017

### DIFF
--- a/src/Network/Receive/kRO/Sakexe_0.pm
+++ b/src/Network/Receive/kRO/Sakexe_0.pm
@@ -10,7 +10,7 @@
 #  See http://www.gnu.org/licenses/gpl.html for the full license.
 #
 #  $Revision: 6687 $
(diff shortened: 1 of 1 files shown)


lututui:
/// Adds new member to a party.
/// 0a43 <account id>.L <role>.L <class>.W <base level>.W <x>.W <y>.W <state>.B <party name>.24B <char name>.24B <map name>.16B <item pickup rule>.B <item share rule>.B (ZC_ADD_MEMBER_TO_GROUP3)

0A44 and 0A43 are very wrong 


/// Sends party information (ZC_GROUP_LIST).
/// 0a44 <packet len>.W <party name>.24B { <account id>.L <nick>.24B <map name>.16B <role>.B <state>.B <class>.W <base level>.W }* <item pickup rule>.B <item share rule>.B <unknown>.L

sctnightcore:
@lututui  It can only be used temporarily.
I have tested 0a43 it works.

PR #1735, opened 2018-02-02
[kRO Zero] fix freeze char when login/change map 

alisonrag:
kRO Zero freeze character everytime when login in a map-server.

this pull remove this restriction

especial thanks to jun and asheraf


Topongoro:
its not working to Idro

Issue #1720, opened 2018-01-30
All TODO : pServe 2017+


sctnightcore:
###### Additional Details:
TODO 

 Adds new member to a party.
0a43
0a44

implemented in #1884

sctnightcore:
```
    '0069': 
    [
        (
            ("accept login", (0, 100000000)),
            [
                ('auth code', generic, 4, 1),
                ('account id', generic, 4, 1),
                ('level', generic, 4, 1),
                ('last ip', generic, 4, 1, getInt, printIp),
                ('last time', string, 26, 1),
                ('sex', generic, 1, 1, getInt, printSex),
                (' num', counter, 0, 1),
                (' ip', generic, 4, 1, getInt, printIp),
                (' port', generic, 2, 1),
                (' name', string, 20, 1),
                (' user count', generic, 2, 1),
                (' state', generic, 2, 1, getInt, printCharServerState),
                (' server type', generic, 2, 1, getInt, printCharServerType),
            ]
        ),
    ],

	
  '0ac4': 
    [
        (
            ("accept login", (0, 100000000)),
            [
                ('auth code', generic, 4, 1),
                ('account id', generic, 4, 1),
                ('level', generic, 4, 1),
                ('last ip', generic, 4, 1, getInt, printIp),
                ('last time', string, 26, 1),
                ('sex', generic, 1, 1, getInt, printSex),
                ('unknown', unknown, 17, 1),
                (' num', counter, 0, 1),
                (' ip', generic, 4, 1, getInt, printIp),
                (' port', generic, 2, 1),
                (' name', string, 20, 1),
                (' user count', generic, 2, 1),
                (' state', generic, 2, 1, getInt, printCharServerState),
                (' server type', generic, 2, 1, getInt, printCharServerType),
                (' unknown', unknown, 128, -7),
            ]
        ),
    ],
```
new account_server_info for pServer

sctnightcore:
New charBlockSize for client 201708+ 
New recv in client 201708+
```
// 20170830
packet(SMSG_PLAYER_STAT_UPDATE_7,0x0acb,  12, &PlayerRecv::processPlayerStatUpdate7,20170830);
packet(SMSG_PLAYER_GET_EXP2,0x0acc,  18, &PlayerRecv::processPlayerGetExp2,20170830);


// 20171207
if (packetVersion >= 20171207)
{
[...]

PR #1736, opened 2018-02-02
new account_server_info for sT20170305+

sctnightcore:
Issues : #1720
from kRO:zero

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -266,18 +266,32 @@ sub parse_account_server_info {
 	} else {
 		delete $args->{lastLoginIP};
 	}
-
-	@{$args->{servers}} = map {
-		my %server;
-		@server{qw(ip port name users display)} = unpack 'a4 v Z20 v2 x2', $_;
-		if ($masterServer && $masterServer->{private}) {
+	if ($args->{switch} eq '0069') {
+		@{$args->{servers}} = map {
+			my %server;
+			@server{qw(ip port name users display)} = unpack 'a4 v Z20 v2 x2', $_;
+			if ($masterServer && $masterServer->{private}) {
 			$server{ip} = $masterServer->{ip};
-		} else {
+			} else {
 			$server{ip} = inet_ntoa($server{ip});
-		}
-		$server{name} = bytesToString($server{name});
-		\%server
+			}
+			$server{name} = bytesToString($server{name});
+			\%server
 	} unpack '(a32)*', $args->{serverInfo};
+}	
+	if ($args->{switch} eq '0AC4') {
+			@{$args->{servers}} = map {
+			my %server;
+			@server{qw(ip port name users display)} = unpack 'a4 v Z20 v3 a128', $_;
+			if ($masterServer && $masterServer->{private}) {
+				$server{ip} = $masterServer->{ip};
+			} else {
+				$server{ip} = inet_ntoa($server{ip});
+			}
+			$server{name} = bytesToString($server{name});
+			\%server
+		} unpack '(a160)*', $args->{serverInfo};
+	}
 }
 
 sub reconstruct_account_server_info {
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -279,7 +279,7 @@ sub parse_account_server_info {
 			\%server
 	} unpack '(a32)*', $args->{serverInfo};
 }	
-	if ($args->{switch} eq '0AC4') {
+	elsif ($args->{switch} eq '0AC4') {
 			@{$args->{servers}} = map {
 			my %server;
 			@server{qw(ip port name users display)} = unpack 'a4 v Z20 v3 a128', $_;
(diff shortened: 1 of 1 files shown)


src/Network/Receive.pm
```diff
@@ -266,18 +266,32 @@ sub parse_account_server_info {
 	} else {
 		delete $args->{lastLoginIP};
 	}
-
-	@{$args->{servers}} = map {
+	if ($args->{switch} eq '0069') {
+		@{$args->{servers}} = map {
```
allanon: Looks like this would work, but there's a lot of code duplication. The only thing that's different is the unpack string, right?

Maybe we could do this instead:
```
my $unpack_str = 'a4 v Z20 v2 C2';
my @unpack_key = qw(ip port name users display);
if ($args->{switch} eq '0AC4') {
	$unpack_str = 'a4 v Z20 v3 a128';
	@unpack_key = qw(ip port name users state property unknown);
}
my $unpack_len = length pack $unpack_str;
@{$args->{servers}} = map {
	my %server;
	@server{@unpack_key} = unpack $unpack_str, $_;
	if ($masterServer && $masterServer->{private}) {
 		$server{ip} = $masterServer->{ip};
 	} else {
 		$server{ip} = inet_ntoa($server{ip});
 	}
 	$server{name} = bytesToString($server{name});
 	\%server
} unpack "(a$unpack_len)*", $args->{serverInfo};
```

What do you think?

sctnightcore:
```
sub parse_account_server_info {
	my ($self, $args) = @_;

	if (length $args->{lastLoginIP} == 4 && $args->{lastLoginIP} ne "\0"x4) {
		$args->{lastLoginIP} = inet_ntoa($args->{lastLoginIP});
	} else {
		delete $args->{lastLoginIP};
	}
	if ($args->{switch} eq '0069') {
		my $unpack_str = 'a4 v Z20 v2 C2';
		my @unpack_key = qw(ip port name users display);
		$unpack_len = a32
	if ($args->{switch} eq '0AC4') {
		$unpack_str = 'a4 v Z20 v3 a128';
		@unpack_key = qw(ip port name users state property unknown);
		$unpack_len = a160
		}
		my $unpack_len = length pack $unpack_str;
		@{$args->{servers}} = map {
		my %server;
		@server{@unpack_key} = unpack $unpack_str, $_;
			if ($masterServer && $masterServer->{private}) {
			$server{ip} = $masterServer->{ip};
			} else {
			$server{ip} = inet_ntoa($server{ip});
			}
			$server{name} = bytesToString($server{name});
			\%server
		} unpack "(a$unpack_len)*", $args->{serverInfo};
	}
}	
```
my $unpack_len = length pack $unpack_str;
idk xD @allanon

allanon: The idea is that we want to split the packet up into pieces which are exactly the number of bytes which the `$unpack_str` will consume. That's 32 bytes and 160 bytes respectively. We _could_ hardcode those two numbers... or we could calculate them. We can calculate them by giving `$unpack_str` to `pack`, which will create an output filled with null bytes, which is exactly the right number of bytes. We then use `length` to figure out how long that string is.

Your copy of the code won't work for a few reasons. The indentation is wrong, and if you fix it, it'll probably be more obvious why the code won't work.

Try this instead:

```Perl
sub parse_account_server_info {
	my ($self, $args) = @_;

	if (length $args->{lastLoginIP} == 4 && $args->{lastLoginIP} ne "\0"x4) {
		$args->{lastLoginIP} = inet_ntoa($args->{lastLoginIP});
	} else {
		delete $args->{lastLoginIP};
	}
	my $unpack_str = 'a4 v Z20 v2 C2';
	my @unpack_key = qw(ip port name users display);
	if ($args->{switch} eq '0AC4') {
		$unpack_str = 'a4 v Z20 v3 a128';
		@unpack_key = qw(ip port name users state property unknown);
	}
	my $unpack_len = length pack $unpack_str;
	@{$args->{servers}} = map {
		my %server;
		@server{@unpack_key} = unpack $unpack_str, $_;
		if ($masterServer && $masterServer->{private}) {
			$server{ip} = $masterServer->{ip};
		} else {
			$server{ip} = inet_ntoa($server{ip});
		}
		$server{name} = bytesToString($server{name});
		\%server
	} unpack "(a$unpack_len)*", $args->{serverInfo};
}
```

sctnightcore: so good but I'm not ok with `my $unpack_len = length pack $unpack_str;`
Maybe
```
	my $unpack_str = 'a4 v Z20 v2 C2';
	my @unpack_key = qw(ip port name users display);
	my $unpack_len = a32
	if ($args->{switch} eq '0AC4') {
		$unpack_str = 'a4 v Z20 v3 a128';
		@unpack_key = qw(ip port name users state property unknown);
        $unpack_len = a160
	}
```

alisonrag: you can use something like this?
https://github.com/OpenKore/openkore/pull/789/files?diff=split

sctnightcore: so good , next day try it .

sctnightcore:
New idea copy `parse_account_server_info` to sakexe_0.pm

PR #1741, opened 2018-02-04
make xConf respect command line arguments

Nipodemos:
The old way, if someone make a items_control.txt with different name, it wouldn't work 
(ex.: if it was "items_control2.txt")


Nipodemos:
this pull request still doesn't solve the problem, i will search more

PR #1746, opened 2018-02-05
moved sub shop_sold to Receive.pm, removed sub from sakexe_0.pm

Nipodemos:


alisonrag:
we need to test this in official and private server, if work in this 2 we can merge it

Nipodemos:
i tested in RMS and worked, so need to test on others servers

PR #1754, opened 2018-02-07
[2018-02-07] kRO Zero update 

alisonrag:
update:
* recvpackets
* sync
* shuffle
* items

its possible to play, but is necessary bypass new **motp** system


majjng:
How to bypass motp system?

PR #1772, opened 2018-02-14
[TalkNPC] fix getting stuck when disconnected

lututui:


Nipodemos:
i tested right now, and worked perfectly

before that openkore got stuck, or disconnects, now it is smooth 😄

PR #1783, opened 2018-02-18
[Base] AccountServer

alisonrag:
update in Account Server to allow login based in token [kRO Zero]

startRO param:
`Ragexe.exe 1rag1 -t:OpenkoreToken accountName G RagnarokZero /account:openkore.xml`

where 
OpenkoreToken need to be equal to config adminPassword

and
accountName need to be equal to config username

### DIFF
--- a/src/Base/Ragnarok/CharServer.pm
+++ b/src/Base/Ragnarok/CharServer.pm
@@ -233,7 +233,7 @@ sub sendCharInfo {
 
 			my $sex;
 
-			if($masterServer->{charBlockSize} == 145 && $masterServer->{serverType} =~ /^iRO/) {
+			if($masterServer->{charBlockSize} >= 145) {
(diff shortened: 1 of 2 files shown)


src/Base/Ragnarok/CharServer.pm
```diff
@@ -233,7 +233,7 @@ sub sendCharInfo {
 
 			my $sex;
 
-			if($masterServer->{charBlockSize} == 145 && $masterServer->{serverType} =~ /^iRO/) {
+			if($masterServer->{charBlockSize} >= 145) {
```
sctnightcore: well ?

alisonrag: all char block size >= 145 have sex in struct, we dont need to filter by server

PR #1796, opened 2018-02-21
Make coords optional in sl command

lututui:
Unless provided, the target coordinates used will be our standing position

Valid examples (with Storm Gust - ID 89):

* Using max lvl available on our current position (behaves like `ss 89`)
`sl 89`

* Using lvl 3 on our current position (behaves like `ss 89 3`)
`sl 89 3`

* Using max lvl available on coords 30 75
`sl 89 30 75`

* Using lvl 6 on coords 30 75
`sl 89 30 75 6`

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -5295,11 +5295,21 @@ sub cmdUseSkill {
 	my @args = parseArgs($args_string);
 
 	if ($cmd eq 'sl') {
-		my $x = $args[1];
-		my $y = $args[2];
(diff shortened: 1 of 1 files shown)


lututui:
Documented: 
http://openkore.com/index.php?title=Skill_Use_Command&diff=prev&oldid=3955
https://openkorebrasilwiki.miraheze.org/w/index.php?title=Comando_de_Utiliza%C3%A7%C3%A3o_de_Habilidade&diff=prev&oldid=1943

PR #1814, opened 2018-02-27
Fix eventmacro crash on reload with unclosed blocks

lututui:
Reported by Kotsu

# How to reproduce:

```
automacro crash {
	BaseLevel > 0
	call {
		if ($var == 0) {
			log I won't close this if
	}
}
```

Run openkore and use command ```reload eventMacros.txt```

# Crash message

image

# After applying this patch:
image


Nipodemos:
i truly have no ideia what this piece of code do and how it changed so much the error message

i must study more

PR #1815, opened 2018-02-27
[Receive.pm] move some identical  actor functions

alisonrag:
move some identical functions related to actor from st0/sakexe0 to Receive.pm

actor_muted is already implemented in stat_info in st0 and sakexe_0

### DIFF
--- a/src/Network/Receive/kRO/Sakexe_0.pm
+++ b/src/Network/Receive/kRO/Sakexe_0.pm
@@ -272,7 +272,7 @@ sub new {
 		'01A4' => ['pet_info2', 'C a4 V', [qw(type ID value)]], # 11
 		'01A6' => ['egg_list'], # -1
 		'01AA' => ['pet_emotion', 'a4 V', [qw(ID type)]], # 10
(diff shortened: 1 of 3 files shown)


src/Network/Receive/kRO/Sakexe_0.pm
```diff
@@ -272,7 +272,7 @@ sub new {
 		'01A4' => ['pet_info2', 'C a4 V', [qw(type ID value)]], # 11
 		'01A6' => ['egg_list'], # -1
 		'01AA' => ['pet_emotion', 'a4 V', [qw(ID type)]], # 10
-		'01AB' => ['actor_muted', 'a4 v V', [qw(ID duration)]], # 12
+		'01AB' => ['stat_info', 'a4 v V', [qw(ID type val)]], # was "actor_muted"; is struct/handler correct at all?
```
sctnightcore: 01ab <account id>.L <var id>.W <value>.L

sctnightcore: 01ab <account id>.L <var id>.W <value>.L

alisonrag:
This should work the same way

PR #1817, opened 2018-03-01
[kRO:zero]Fix cannot login to char

sctnightcore:


lututui:
Can someone playing kRO Zero confirm this works

alisonrag:
as we can see in here:
https://github.com/OpenKore/openkore/blob/cb5c7b7773d1307f6fb03f64e9683fa67f1f8f5c/src/Network/DirectConnection.pm#L534

this captcha option is related to char server, not map server as in kRO Zero, so we need to disable this

PR #1844, opened 2018-03-11
copy packet from sT0 to sakexe0 

sctnightcore:


sctnightcore:
@lututui 
It have in rA and Herc

alisonrag:
@sctnightcore
Move the sub to receive.pm then add only the packets to sakexe
We dont need to duplicate the sub

PR #1884, opened 2018-03-28
merge/move party packets

alisonrag:
merge/move party packets from sakexe_0 and ServerType0 to Receive.pm

now the party support is equal to official and private servers.

tested in bRO:
bro

and in private rAthena Based:
rathena


sctnightcore:
Pserver is work . 
image

PR #1904, opened 2018-04-02
Implementing:setMonControlFilename,setShopFilename

OvoKore:
Implementing: setMonControlFilename and setShopFilename

### DIFF
--- a/src/Settings.pm
+++ b/src/Settings.pm
@@ -709,6 +709,18 @@ sub getMonControlFilename {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


c4c1n6kr3m1:
i do think it will easier if u have additional switchShopFile, switchMon_controlFile, similar with switchConfigFile

Nipodemos:
this is for a specific use, and don't have command line options, but it will not make the openkore slower ou heavier, so why not?

PR #1911, opened 2018-04-06
[ Send / Receive ] Move Rodex subs

alisonrag:
move subs and add support to rodex in sakexe send


src/Network/Send/kRO/Sakexe_0.pm
```diff
@@ -1367,434 +1381,4 @@ sub sendFriendRemove {
 # 0x0208,11,friendslistreply,2:6:10
 # sendFriendReject:0/sendFriendAccept:1
 
-1;
```
c4c1n6kr3m1: when i fast searching a packet meaning using notepad++ , many times the results is these lines you want to delete
So i vote to not to delete , as a document

alisonrag: what about update openkore\tables\packetdescriptions.txt?

in my point of view this piece of code is unnecessary in this file, some of this packets dont have description or are receive packets

c4c1n6kr3m1: if it's already in packetdescriptions.txt then it's fine
i remember asked these lines on forum , why we have these line almost in every ST and a moderator in forum said as a document (or something similar)

alisonrag:
image
tested in iRO

src/Network/Send/kRO/Sakexe_0.pm
alisonrag: the better is keep all in one file

PR #1914, opened 2018-04-07
packets send update

alisonrag:
```
'00AB' => ['send_unequip_item', 'v', [qw(Index)]],
'00BB' => ['send_add_status_point', 'v2', [qw(statusID Amount)]],
'0112' => ['send_add_skill_point', 'v', [qw(skillID)]],
'0130' => ['send_entering_vending', 'V', [qw(accountID)]],
'02BA' => ['send_shortcut_key_change', 'v c V v', [qw(index isSkill ID count)]],
```

tested in iRO and bRO
image

image

### DIFF
--- a/src/Network/Send.pm
+++ b/src/Network/Send.pm
@@ -1441,4 +1441,48 @@ sub rodex_close_mailbox {
 	undef $rodexList;
 }
 
+sub sendEnteringVender {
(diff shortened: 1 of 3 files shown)


lututui:
About 02BA:
* isSkill is actually called type in rA
* count is the skill level

Maybe rename those accordingly?

PR #1959, opened 2018-04-22
Shop log changes

Nipodemos:
something i've noticed, always when you sell an item, this is how is printed on log:
`[Apr 02 11:22:57 2018] [Apr 02 11:23:11 2018] Sold: Churrasco de Selvagem x 200 - 1300000z (Buyer charID: 90 6C CF 00)`
the `datetime` information is printed twice, and the **totally** irrelevant charID of the buyer is also printed...

this pull request fix this
Also, now you have the option of print or not the log of items sold, it is enabled by default.

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -1879,9 +1879,9 @@ sub shop_sold_long {
 	$shopEarned += $earned;
 	$articles[$number]{quantity} -= $amount;
 	
-	my $msg = TF("Sold: %s x %s - %sz \n", $articles[$number]{name}, $amount, $earned);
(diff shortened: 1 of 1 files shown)


lututui:
1- Keep the charID
2- Keep the removed sold time in the console message, but not in the shop log

```
my $msg = TF("Sold: %s x %s - %sz (Buyer charID: %s)\n", $articles[$number]{name}, $amount, $earned, $charID);
shopLog($msg) if $config{logShop};
message(getFormattedDate($when) . " " . $msg, "sold");
```

Nipodemos:
thanks for the tips :smiley:

PR #1967, opened 2018-04-27
๊Update . . LegacyChangelog.md

sctnightcore:


Nipodemos:
always good to not erase anything, it can be useful someday

alisonrag:
its ok to me, but @itsrachelfish, @allanon need to be ok with this

PR #1988, opened 2018-05-08
Clear Something in pServer

sctnightcore:


alisonrag:
packets will be not added to packet_list, so openkore will not parse, 
the better world is move these packets to sakexe_0 and keep only the handler in servertype/regexeRE file

PR #2024, opened 2018-05-31
Star Emperor/Soul Reaper

sctnightcore:
https://github.com/HerculesWS/Hercules/pull/2049
ty @Asheraf


Nipodemos:
:heart: i want so much to play with these classes

PR #2030, opened 2018-06-03
[items_control.txt] equipments must have literal name instead of ID

Nipodemos:
this is to prevent that something like this happens:
image

for those who can't read: the bot sold a **VVS Damascus with Fire Element**
because by default items_control.txt was selling ALL Damascus


control/items_control.txt
```diff
@@ -72,242 +72,242 @@ Arrow 3000 1 0
...
-2405 0 0 1 # Boots
-2103 0 0 1 # Buckler
-2227 0 0 1 # Cap
-2314 0 0 1 # Chain Mail
-2233 0 0 1 # Circlet
-2309 0 0 1 # Coat
-2310 0 0 1 # Coat [1]
-2302 0 0 1 # Cotton Shirt
-Cotton Shirt [1] 0 0 1
+Adventurer's Suit 0 0 1 # 2305
```
sctnightcore: I think add name and itemID ? 

```
Adventurer's Suit 0 0 1 # 2305
2305 0 0 1 # Adventurer's Suit
```

Nipodemos: it is not good, because that way, if the armor has refinement level, or has a card, it will sell the same way, because the ID is still the same

lututui:
Maybe it'd be a better idea to match the nameID (item ID) only if said item is 'clean' (that is, has no enchant, card, renaming (forged))? I have no idea how hard that would be.

alisonrag:
I think we need merge this and in the future add a check if item have card/refine/options

Nipodemos:
fair enough

PR #2062, opened 2018-06-16
forgot to define var on sub

Nipodemos:
Sorry about that, when i pasted i forgot this little part :worried:


Nipodemos:
thanks, withou this openkore doesn't even open hahaha

PR #2076, opened 2018-06-21
[cRO] fix freeze char

alisonrag:
fix freeze char when change map.

merge map_change and map_changed


src/Network/Receive/ServerType0.pm
```diff
@@ -623,6 +623,7 @@ sub new {
 		'0AA2' => ['refineui_info', 'v v C a*' ,[qw(len index bless materials)]],
 		'0AC4' => ['account_server_info', 'v a4 a4 a4 a4 a26 C x17 a*', [qw(len sessionID accountID sessionID2 lastLoginIP lastLoginTime accountSex serverInfo)]],
 		'0AC5' => ['received_character_ID_and_Map', 'a4 Z16 a4 v a128', [qw(charID mapName mapIP mapPort mapUrl)]],
+
```
sctnightcore: well :thinking:

alisonrag: 0ac5 and 0ac7 is already implemented in st0/sakexe_0

sctnightcore: It has a gap

sctnightcore: It has a gap

PR #2093, opened 2018-06-27
storageAuto_npc_type for iRO

sctnightcore:
Tested : iRO: Chaos
IT work .
EX. 
```
storageAuto 1
storageAuto_npc payon 181 104
storageAuto_standpoint
storageAuto_distance 5
storageAuto_npc_type iRO_novip
storageAuto_npc_steps
storageAuto_password
storageEncryptKey
storageAuto_keepOpen 0
storageAuto_useChatCommand
storageAuto_useItem 0
storageAuto_useItem_item
storageAuto_notAfterDeath
relogAfterStorage 0
minStorageZeny 50
```


Nipodemos:
someone said to me someday that making an option only for 1 server is not a good thing

lututui:
I don't think this is necessary.

There are too many servers, we would end up with several npc types

PR #2096, opened 2018-06-28
[eventMacro] Changes requested by henry bk

Nipodemos:
@Henrybk said to me:
you are using eval 3 times ( you could do once and copy ) and you create two variables `@testArray ` and `%testHash` but only uses one.

Additionally, added support to when a person return a reference to a scalar variable (difficult but possible)


Nipodemos:
i tested and it still working, so i'll merge this

PR #2109, opened 2018-07-02
support regex groups on eventMacro if-elsif-while blocks

Nipodemos:
automacro for testing
```perl
automacro test {
	exclusive 1
	run-once 1
	BaseLevel > 0
	priority -5555
	macro_delay .5
	call {
		$phrase = Name: Nipodemos, Age: 21
		if ($phrase =~ /Name: (\w+)\, Age: (\d+)/i ) {
			log name is: $.lastMatch1
			log age is: $.lastMatch2
		}
	}
}
```
output:
```
[eventMacro] Conditions met for automacro 'test', calling macro 'tempMacro4'
[eventmacro log] name is: Nipodemos
[eventmacro log] age is: 21

```


lututui:
I believe this needs to be a little more heavily tested.

For example, does this patch break somehow your macros for leveling up to lvl 99, or do they work as expected?

Nipodemos:
i can test this now
although i don't use regex a lot on the macro

Nipodemos:
now that the #2368 was merged, we don't need this pull request anymore

PR #2116, opened 2018-07-05
Rewrite eventmacro multivalidator conditions

Henrybk:
MultipleValidatorState and MultipleValidatorEvent are a horrible mess, we should get rid of them.

This is (initially) a rewrite of the condition IsInCoordinate so it won't use MultipleValidatorState anymore.

It now uses instead the base eventMacro::Condition, this is much safer and cleaner.
This condition now also supports variables.
Requires further testing.


Nipodemos:
i'll test it today at night

Nipodemos:
tested like that:
`IsInCoordinate $x $y`
`IsInCoordinate $x..$y 150`
`IsInCoordinate $x..$y $x..$y`
`IsInCoordinate 150 $x..$y`
and `IsInCoordinate 150 150`

and worked perfectly

PR #2119, opened 2018-07-07
[Feature] Buying Store

alisonrag:
https://irowiki.org/wiki/Buying_Store

config.txt keys:
```
buyerShopTitleOversize 0
buyerShopAuto_open 0
buyerShop_random 0
```

commands:
```
openbuyershop
closebuyershop
```
file:
buyer_shop.txt

tests:
Char that have fakelicense:
image

Char that have license and buying skill:
image

Check if have skill/item
image

auto open buyer shop:
image


Nipodemos:
i think the two commands should be `openbuyshop` and `closebuyshop`
since the normal shop command is `openshop`

src/Misc.pm
```diff
@@ -4631,6 +4634,116 @@ sub closeShop {
...
+
+	my $max_items = 2;
+
+	if (!$char->{skills}{ALL_BUYING_STORE}{lv}) { # don't have skill but have the necessary item
+		my $item = $char->inventory->getByNameID(12548);
+		if(!$item) {
+			error T("You don't have the Buying Store skill or Black Market Bulk Buyer Shop License.\n");
+			return;
+		} else {
+			$item->use;
```
Nipodemos: another important thing: the code should look if there are items to sell before use the item
because if has no items to sell, char get stuck, because the skill were already casted

this also happens with normal openshop. Something i want to change

src/Misc.pm
```diff
@@ -4631,6 +4634,116 @@ sub closeShop {
...
+
+	if($limitZeny > $char->{zeny}) {
+		$limitZeny = $char->{zeny};
+	}
+
+	Plugins::callHook ('buyer_open_shop', {title => $buyer_shop{title}, limitZeny=> $limitZeny, items => \@items});
+	$messageSender->sendBuyBulkOpenShop($limitZeny, 1, $buyer_shop{title}, \@items);
+	
+	message T("Trying to set up buyer shop...\n"), "vending";
+	$buyershopstarted = 1;
```
Nipodemos: just to be sure, when it gets to this line, is because it have 100% of chance of open buy shop right? it's impossible to fail when get to this line right?

src/Network/Receive.pm
```diff
@@ -6176,4 +6176,18 @@ sub refine_result {
...
+	my $result = $args->{result};
+	if($result == 1){
+		message TF("Failed to open Purchasing Store.\n"),"info";
+	} elsif ($result == 2){
+		message TF("The total weight of the item exceeds your weight limit. Please reconfigure.\n"), "info";
+	} elsif ($result == 8){
+		message TF("Shop information is incorrect and cannot be opened.\n"), "info";
+	} else {
+		message TF("Failed opening your buying store.\n");
+	}
```
Nipodemos: shoud put `$buyershopstarted = 0;` here too, shoudn't?

src/Settings.pm
```diff
@@ -189,6 +191,7 @@ sub parseArguments {
 		'mon_control=s',      \$mon_control_file,
 		'items_control=s',    \$items_control_file,
 		'shop=s',             \$shop_file,
+		'buyer_shop=s',		  \$buyer_shop_file,
```
Nipodemos: don't forget to put a default `buy_shop.txt` file on control folder

src/Misc.pm
c4c1n6kr3m1: as far as i remember,
it will fail if char zeny less than all items price * amount

src/Commands.pm
```diff
@@ -88,6 +88,7 @@ sub initHandlers {
 	clearlog			=> \&cmdChatLogClear,
 	closeshop			=> \&cmdCloseShop,
 	closebuyshop		=> \&cmdCloseBuyShop,
+	closebuyershop		=> \&cmdCloseBuyerShop,
```
sctnightcore: cmdCloseBuyShop , cmdCloseBuyerShop 

I think can be used together?

src/Misc.pm
```diff
@@ -4631,6 +4634,116 @@ sub closeShop {
...
+	$buyer_shop{title} = $buyershopnames[int rand($#buyershopnames + 1)];
+	$buyer_shop{title} = ($config{shopTitleOversize}) ? $buyer_shop{title} : substr($buyer_shop{title},0,36);
+
+	foreach my $item (@items) {
+		$limitZeny += ($item->{amount} * $item->{price});
+	}
+
+	if($limitZeny > $char->{zeny}) {
+		$limitZeny = $char->{zeny};
+	}
```
c4c1n6kr3m1: did u ever test it with char zeny lower than $limitZeny ? will it succed or fail ?

src/Network/Receive.pm
alisonrag: fixed

src/Misc.pm
alisonrag: image
fixed

src/Misc.pm
alisonrag: image

will be ok

src/Commands.pm
alisonrag: merged

src/Settings.pm
alisonrag: its already there
```
https://github.com/OpenKore/openkore/pull/2119/files#diff-c02ec3dcf568b2ade692634056827dacR1
```

src/Misc.pm
alisonrag: yep, theres no way to get here without the items or skill

lehotf:
Before everything, FANTASTIC JOB!!!!

Just a small point. A merchant class character is opening the store with 2 products.

The item 6377 is Bulk Buyer Shop License. So I should give you access to 5 items, I think.

So I think it should look like this:

my $ max_items = 2;    # was 5
my @items = ();

if ($ char-> inventory-> getByNameID (6377)) {
$max_items = 5; # was 2
}


---------

In Portuguese

Fala, cara. Vi que você é brasileiro. Ótimo trabalho, cara. Quando fui testar vi que a loja abriu com apenas dois itens, mesmo sendo mercador com Bulk Buyer Shop License. Tive que fazer a alteração acima pra funcionar. Faz sentido pra mim, já que o mercador que tem a habilidade e a licença 6377 deveria abrir loja com 5 itens.

legokidd:
i would like to ask if we can put something like al function for buyvend. so that we can see how many is sold to me or being sold

alisonrag:
> i would like to ask if we can put something like al function for buyvend. so that we can see how many is sold to me or being sold

I did not understand the question.

if u are talking about a command to enter in another buying it already exist
`bl` to show buying list
`buyer <index>` to enter in buying 
`buyer <index> <item index> <amount>` to sell item to buyer

if u are talking about self vending:
`bs `to show self buying list

PR #2124, opened 2018-07-07
Removed everything about MultipleValidator

Henrybk:
The following conditions will now use Condition.pm as base
```
BaseActorNearDist.pm
BaseMsgDist.pm
BaseMsgName.pm
BaseMsgNameDist.pm
IsInMapAndCoordinate.pm
```

The following files have been removed:
```
MultipleValidatorEvent.pm
MultipleValidatorState.pm
```

Needs testing


Nipodemos:
i remember that before this PR , PlayerNearDist wasn't updating correctly. and it was not working properly because of that

let's see if with PR this will fix, i hope it does <3

Nipodemos:
Just tested PlayerNearDist Condition, and everything worked perfectly, with variable and without
always updating perfectly, no errors

so i thought, what if user make a typo?
so i wrote `PlayerNearDist /muita treta vish/ ` without the numeric comparison
and give this error:
image
although is understandable, i believe this is not a desired crash, should only point out the mistake

plugins/eventMacro/eventMacro/Condition/IsInMapAndCoordinate.pm
```diff
@@ -17,62 +19,150 @@ my $map_qr = qr/\w+\d*_?(?:\d|\w)*/;
...
+	$self->{x_validators} = {};
+	$self->{y_validators} = {};
+	$self->{map_validators} = {};
+	
+	my $var_exists_hash = {};
+	
+	my $member_index = 0;
+	foreach my $member (split(/\s*,\s*/, $condition_code)) {
 		my ($map, $coord_x, $coord_y);
 		if ($member =~ /^(?:($map_qr)|(?:(\S+)\s+(\S+))|($map_qr)\s+(?:(\S+)\s+(\S+)))$/) {
```
Nipodemos: map doesnot support variables because of this regex i think.
```
[eventMacro] Ignoring automacro 'IsInMapAndCoordinate'
[eventMacro] Error in condition 'IsInMapAndCoordinate'
[eventMacro] Error type: Wrong condition syntax ('$map 97 62, 96 62, $map 96 61')
[eventMacro] Error code: 'List member '$map 97 62' must have a map and/or a coordinate'.
```

Nipodemos: but x and y are supporting variables normally

Henrybk:
@Nipodemos I believe I fixed both those errors

Nipodemos:
i'll test now

plugins/eventMacro/eventMacro/Condition/IsInMapAndCoordinate.pm
```diff
@@ -37,34 +35,30 @@ sub _parse_syntax {
 	my $member_index = 0;
 	foreach my $member (split(/\s*,\s*/, $condition_code)) {
 		my ($map, $coord_x, $coord_y);
-		if ($member =~ /^(?:($map_qr)|(?:(\S+)\s+(\S+))|($map_qr)\s+(?:(\S+)\s+(\S+)))$/) {
-			$map = $1 || $4;
-			$coord_x = $2 || $5;
-			$coord_y = $3 || $6;
-		}
+		my @parts = split(/\s+/, $condition_code);
```
Nipodemos: little typo: you have to split `$member`

PR #2134, opened 2018-07-09
[iRO] Unfreeze char after map loaded

sctnightcore:
- [x] iRO:RE
- [x] iRO:ReStart

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -4861,8 +4861,8 @@ sub map_change {
 	} else {
 		$messageSender->sendMapLoaded();
 		# $messageSender->sendSync(1);
-#TODO : Need check classic iRO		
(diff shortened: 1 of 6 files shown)


c4c1n6kr3m1:
just tested, renewal and restart still use 00A7 for item_use

sctnightcore:
ah = = Test with 00A7 it work :p

c4c1n6kr3m1:
restart does not send 0447 , 
yet?

sctnightcore:
@c4c1n6kr3m1  Can u check it? 

I dont have iRO_Restart client .

c4c1n6kr3m1:
yes. i tested it , restart don't send 0447
it's weird isn'it , may be they put it letter

sctnightcore:
@c4c1n6kr3m1 
Maybe problem is unfreeze char not use item.

c4c1n6kr3m1:
as i said, restart client doesn't send 0447 right now
you shouldn't rush

lututui:
@Fadreus Does this still work (after my change)?

PR #2136, opened 2018-07-09
Revert "[eventMacro]improve statments of eventMacro"

Nipodemos:
Unfortunaly i have found a serius bug on this Pull Request
The problem is the `&config()` return an empty string if it's false
so, if the statment is something like `&config(skillsAddAuto) != 1` and the config is false, then the code will accuse of a non-existence of first argument
image

so i must first solve this...


Nipodemos:
just deleting branch

PR #95, opened 2016-05-11
Completed received messages by `skill_msg 07E6`

cydh:
- [x] Code review
- [ ] QA testing on idRO
- [ ] QA testing on iRO

Received message will be dispalyed as: 

> `[<Skill Name>] <Message from msgstringtable.txt>`

Signed-off-by: Cydh Ramdh cydh@pservero.com

PR #2149, opened 2018-07-13
[Receive] Skill message rework

alisonrag:
Skill message rework.

updated to work with private server

Based in #95


lututui:
Fixed conflict

PR #2165, opened 2018-07-18
Add ruRO :p

sctnightcore:


psichopath:
I have access to the new russianRO server.
I would love to help with adding his data (but dont know how).

PR #2186, opened 2018-07-24
[Network] idRO update

alisonrag:
I'm happy to say that after months of trying, we finally got to connect to the idRO Gravindo without using xkore mode 1 or the client.

Because of FF we were able to implement the same encryption used on the client to login in openkore

Login:
image

In-Game:
image


Nipodemos:
@c4c1n6kr3m1 this way is good or not?

AncientEgyptSoul:
my suggestion : 
title : update packet login
@alisonrag avoid using real server name & ip (or make it censored)
note : btw, congrats for breakthrough

PR #2192, opened 2018-07-26
change configKeys from 'disconnect' to 'quit'

Nipodemos:
## Things changed:
all configs that start with 'disconnect' were changed to `quit`, since the config is in fact quiting

#### dcOnDeath
this has additional changes, it was splitted into two different configs: `quitOnDeath` and `noRespawnAfterDeath`
the `noRespawnAfterDeath` config are meant for people who don't want to get back to city when the char dies.

#### dcOnDualLogin
this has also additional changes, it was splitted into two different configs too: `quitOnDualLogin`and `reconnectTimeAfterDualLogin`
the `reconnectTimeAfterDualLogin` defines how many seconds it will wait before recconect after a dual login detected

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -144,21 +144,24 @@ cachePlayerNames_maxSize 100
 
 clientSight 20
 
-dcPause 1
-dcOnDeath 0
-dcOnDualLogin 0
-dcOnDisconnect 0
-dcOnEmptyArrow 0
-dcOnMaxReconnections 0
-dcOnMute 0
-dcOnPM 0
-dcOnZeny 0
-dcOnStorageFull 1
-dcOnPlayer 0
-dcOnServerShutDown 0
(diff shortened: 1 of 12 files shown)


Fadreus:
My only concern is the wiki.
Some still missing in wiki _(or already been updated? I haven't check yet for recent changes >w<)_

Nipodemos:
forgot about this.... i'll change tonight

lututui:
I really think changing names in this case is a very bad idea.

Maybe you could implement different behaviors? Such as

```
dcOnStorageFull 0 -> do nothing
dcOnStorageFull 1 -> quit
dcOnStorageFull 2 -> offlineMode
```

?

Nipodemos:
yeah, thinking about it, this will be problematic to change

but this pull request it is not intended to implement different behaviors, so i'll close this

PR #2209, opened 2018-07-31
Added git clone command for guidance

cheezeburger:
Let users know more on how to use the method mentioned.


Nipodemos:
@Fadreus this is just a quick start, if people want more information (and they will) they can google it and find duzens of tutorials

Nipodemos:
thank you for your contribution :smiley:

PR #2249, opened 2018-08-14
[eventMacro] New condition: IsNotInMapAndCoordinate

Nipodemos:
Resuming, this is basiclly the code from IsInMapAndCoordinate,with the inverse result on the end

and you can't put more than one coordinate/map
`IsNotInMapAndCoordinate prontera 150 150, payon, geffen 130 121` is wrong
can have only 1 member , like `IsNotInMapAndCoordinate prontera 150 150`

### DIFF
--- a/plugins/eventMacro/eventMacro/Condition/IsNotInMapAndCoordinate.pm
+++ b/plugins/eventMacro/eventMacro/Condition/IsNotInMapAndCoordinate.pm
@@ -18,7 +18,7 @@ sub _parse_syntax {
 	my ( $self, $condition_code ) = @_;
 	
 	if ($condition_code =~ /,/) {
-		$self->{error} = "comma separeted values are not accepted on this condition";
+		$self->{error} = "Comma separated values are not accepted on this condition";
 		return 0;
 	}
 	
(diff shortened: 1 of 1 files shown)


Fadreus:
Not sure which PR but semicolon no longer used in eventMacro?

lututui:
This is why I'm not happy with eventmacro

Nipodemos:
@Fadreus yeah, semicolon no longer exists

@lututui why?

Fadreus:
Crap.. Now this seems to be slow for me, have to wait line by line >w<
PR 2226 removed this, no? :thinking:

Fadreus:
Hmm.. Now this seems to be slow for me, have to wait line by line >w<
PR 2226 removed this, no? :thinking:

Nipodemos:
is because by default takes 1 second between each line
when were used sub lines, they didn't respect this, that's why ther are faster

but you can simply put on any automacro:
`macro-delay 0`
and it will have no delay between lines

lututui:
Also you can nest with [] to ignore macro_delay

```
macro example {
[
do something
do something else
do yet another thing
]
do something respecting macro_delay
}```

All those actions inside [] will be executed with macro_delay 0

PR #2254, opened 2018-08-15
Network cleaning

lututui:


src/Network/Send.pm
```diff
@@ -1750,4 +1750,65 @@ sub sendAdoptReply {
...
 }
 
+sub sendPrivateAirshipRequest {
+	my ($self, $args) = @_;
+	
+	$self->sendToServer($self->reconstruct({
+		switch => 'private_airship_request',
+		map_name => stringToBytes($args->{map_name}),
+		nameID => $args->{nameID},
+	}));
```
sctnightcore: I think this is better?
```
		map_name => stringToBytes($MapName),
		itemID => $itemID,


```

lututui:
Ok, but it makes no difference at all

PR #2256, opened 2018-08-15
Fix to KVM NPC bug

grodli:
While talking to KVM NPC's to join the waiting room, sometimes it "randomly" deletes the quest log and doesnt initialize the talk properly, leading to an infinite loop until you make a MANUAL move/action using the openkore command.

adding this "else" will deal with empty and wrong initialized talks, fixing the most known issue on KVM NPC's.


Nipodemos:
i like it!

this has no problem in add? because putting an else there can make unwanted results...

you tested a lot?

grodli:
I've been running it for some hours, so far no problems were found.

Would be cool if someone else could also run it for more time, doing different actions...

Nipodemos:
i'll test here i little bit

Nipodemos:
i tested i didn't saw any problems

i hope that this solve the problem on rogue quest too :smile:

grodli:
I sent you a PM on openkorebrasil.
Anyway, I will explain here: Openkore fails to handle "unsuccessful initialized talks to NPCs".

**What is happening?**
In this case, when you try to talk the KVM NPC, instead of initializing a talk, sometimes it will close the "KVM Faction Quest" on your quest log and you need to click on the NPC again to start the talk.

So, Openkore will iniate the talk (debugging on your console $self->{target}: Initiating the talk), but since it wont receive any response or chat box, it will reach the timeout. 
When it reaches the timeout, it will keep an infinite loop on $self->cancelTalk, because $ai_v{'npc_talk'}{'talk'} is empty, and the sub cancelTalk can't handle it properly, not triggering the conversation_end and not changing the status to close the talk.

grodli:
@c4c1n6kr3m1 quest being deleted after clicking to talk to the NPC and not initiating the conversation also happens when you are playing manually, so it is a server issue.

It is totally related to Talk, since it fails under an infinite loop debugging "Trying to auto close the conversation due to error.". So, you can't really do anything when it is stuck inside a loop, not even closing the macro or w/e...

I gave more details on my explanation above.

Nipodemos:
why my commit is mixed here?

lututui:
@grodli I need you to redo this, I kinda fucked up your branch by accident.

Oops

PR #2259, opened 2018-08-16
Update Re:Start files 16-08-2018

PipeDeveloper:
My first pull request! :) Already tested and worked


marielyn:
tested it and works

ZaskarAlpha:
tested... working!

PR #2263, opened 2018-08-17
Update packet for iRO Renewal

zPhotons:
Hello, I can't login on renewal after the update.


Nipodemos:
hahahahah
i wonder how did he do this :thinking:

PR #2295, opened 2018-08-24
Cleanup GM packets and commands

lututui:


src/Commands.pm
```diff
@@ -2603,7 +2594,7 @@ sub cmdGmrecall {
 	}
 
 	my $packet = pack("C*", 0xBC, 0x01).pack("a24", $args);
```
Nipodemos: if i understand well, this line now do nothing.
Shoun't be removed?

lututui: missed that, fixed

PR #2301, opened 2018-08-26
[Network / Receive] Clean 2008 ragexeRE files

alisonrag:
RagexeRE 2008 files

### DIFF
--- a/src/Network/Receive/kRO/RagexeRE_2008_09_10a.pm
+++ b/src/Network/Receive/kRO/RagexeRE_2008_09_10a.pm
@@ -22,28 +22,7 @@ use base qw(Network::Receive::kRO::RagexeRE_2008_08_27a);
 
 sub new {
 	my ($class) = @_;
-	my $self = $class->SUPER::new(@_);
-	my %packets = (
-		# 0x0436 is sent packet
-		# 0x0437 is sent packet
-		# 0x0438 is sent packet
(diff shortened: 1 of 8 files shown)


src/Network/Receive/kRO/RagexeRE_2008_09_10a.pm
```diff
@@ -22,28 +22,7 @@ use base qw(Network::Receive::kRO::RagexeRE_2008_08_27a);
...
 
-1;
-
-=pod
-//2008-09-10aRagexeRE
-packet_ver: 24
-0x0436,19,wanttoconnection,2:6:10:14:18
-0x0437,7,actionrequest,2:6
-0x0438,10,useskilltoid,2:4:6
-0x0439,8,useitem,2:4
```
sctnightcore: dont del it !

lututui: Until we have a better way to document when a packet was added, it's better to keep it there

lututui: Is that up to date?
If so, I'm OK with deleting it from the Network src. Also, maybe we should leave a note to lookup that file?

Nipodemos:
since the packetdescription is up to date, i believe that this pull request can be merged

steadybear69:
do you have updated packetlist?

PR #2304, opened 2018-08-26
[Network] update received_characters_info

alisonrag:
update received_characters_info to work with iRO Renewall
tested in bRO, iRO (Re:Start, Renewall, Classic)


Nipodemos:
now is just wait for @c4c1n6kr3m1

c4c1n6kr3m1:
login is fine , the sex is correct for old char and new char
but , can not login the 4th char ... 5th.... so on

PR #2348, opened 2018-09-05
Add 'revive' command

lututui:
This command allows the use of the Token Of Siegfried to self-revive
Also supports custom servers that have other items like the Token
Allows for force-trying to auto-revive

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -171,6 +171,7 @@ sub initHandlers {
 	relog				=> \&cmdRelog,
 	repair				=> \&cmdRepair,
 	respawn				=> \&cmdRespawn,
+	revive				=> \&cmdRevive,
 	rodex				=> \&cmdRodex,
 	s					=> \&cmdStatus,
 	sell				=> \&cmdSell,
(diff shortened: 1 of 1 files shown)


src/Commands.pm
```diff
@@ -7267,4 +7268,48 @@ sub cmdSearchStore {
...
+	} elsif (scalar @args == 0) {
+		# Try to find Token Of Siegfried
+		$item = $char->inventory->getByNameID(7621);
+	} else {
+		error T("Error in 'revive' command (incorrect syntax)\n".
+				"revive [force|<item name>|<item ID>]\n");
+		return;
+	}
+	
+	if (!$item && $args[0] ne "force") {
```
Nipodemos: the way it is made
is impossible to declare a different item to revive and the "force" arg

if user type something like `revive 1234 force`, and item does not exist, "force" arg will not be respected

lututui: The point of force is NOT specifying an item.
This item check exists solely to avoid trying when it's impossible - something the client would never do.
The item is not sent with the packet

Nipodemos: oh i understand. 
well, then for me the code are perfect :smile:

Issue #2345, opened 2018-09-04
iRO: Get rare critical error , "Assertion Failed!"


Glikadin:
Server: iRO Chaos Rnewal

I get rare error , using eventMacro.txt:
```
                                    ...
		if (&invamount(Pet Food) < 10) {
			do move &rand(185,195) &rand(110,115) yuno
			do move &rand(185,195) &rand(110,115) yuno
			do move &rand(185,195) &rand(110,115) yuno
			do talknpc 197 114 b
			pause &rand(1,3)
			do buy 0 20
			pause &rand(1,3)
			do talk no
			pause &rand(1,3)
		}
                                ...
```
With other NPCs this macro works correctly. But if try buy items from this NPC i get crash.

Error.txt
```
OpenKore version what-will-become-2.1
@ai_seq = NPC eventMacro
Network state = 5
Network handler = Network::DirectConnection
Revision: ctime:2018_08_04
Loaded plugins:
  plugins/breakTime.pl (breakTime; description: Automatically disconnect and reconnect at certain times of the day)
  plugins/busCommands.pl (busCommands; description: receive and send commands (and messages too) via BUS system)
  plugins/customCaption.pl (customCaption; description: customize OpenKore's window caption)
  plugins/eventMacro/eventMacro.pl (eventMacro; description: allows usage of eventMacros)
  plugins/ignore.pl (ignore; description: ignore messages)
  plugins/macro/macro.pl (macro; description: allows usage of macros)
  plugins/map.pl (map; description: map)
  plugins/raiseSkill/raiseSkill.pl (raiseSkill; description: automatically raise character skills)
  plugins/raiseStat/raiseStat.pl (raiseStat; description: automatically raise character stats)
  plugins/reconnect.pl (reconnect; description: v1.0)
  plugins/xConf.pl (xConf; description: commands for change items_control, mon_control, pickupitems, priority)

Error message:
Assertion failed!
 at C:/Work/iRO/q21q30_levelingMechanic/src/deps/Carp/Assert.pm line 271
	Carp::Assert::assert('') called at src/ActorList.pm line 98
	ActorList::add('InventoryList=HASH(0xa115744)', 'Actor::Item=HASH(0x9cd6a6c)') called at src/InventoryList.pm line 109
	InventoryList::add('InventoryList=HASH(0xa115744)', 'Actor::Item=HASH(0x9cd6a6c)') called at src/Network/Receive.pm line 5043
[...]

PR #2350, opened 2018-09-05
Fix assertion fail in some NPC stores

lututui:
Workaround some npcs that have items appearing more than once in their store list, for example the Trader at `moc_ruins 90 149` sells only bananas, but 6 times

Usually, `$Actor::Item->{ID}` is equal to `$Actor::Item->{nameID}` - that WILL crash kore in the event described above (as seen in #2345)

This workaround causes `$Actor::Item->{ID}` to be equal to `$Actor::Item->{binID}` and, therefore, never overlap


lututui:
Fixes #2345

PR #2368, opened 2018-09-12
[eventMacro] rework of statements

Nipodemos:
Before, it was used simple regex to parse statements in eventMacro (too simple)

Now, instead of a regex, code will check letter by letter, to find values correctly and find typos more effectvely
New:

- Using quotes is now to make ignore special characters, so if a string is like `xX/!>~=nipodemosXx=~!!>>` you can just surround it with quotes that all this special characters will be ignored
- Variables, macro_keywords and subs are parsed only on the final step, which makes the code more error-proof, because before things were parsed too early , causing sometimes that statements would be empty just because the result of a macro_keyword was empty too
- Regex were also fixed, before it was getting '(' and ')' inside regexs and parsing, which is wrong. now it doesn't parse theses parenthsis as part of syntax
- Better error handling: Now is more easy to know what is the reason of multiple errors that might appear
and now will show an informative message of what's wrong
- Support for statements with only one argument inside like: 
```perl
if ($name) log var \$name defined
```
- Support to save groups from regex on a variable, each group will be save on `$.lastMatchx` where 'x' is the number of the group, starting by one ($.lastMatch1, $.lastMatch2 and so on)
- Now it is possible to use `!` to revert the result of an statement, just like you do with perl Ex: `if (! 1 != 1)` should return false, but with `!` in front of it, will make return opposite result
- Additionally, while i was coding, i found a bug on switch case statement, and i fixed it

### Test 1: ignore special characters inside quoted string
```
    if ("xX/!>~=nipodemosXx=~!!>>" =~ /nipodemos/) {
        log special characters ignored inside quotes: sucess
    }
```
image

### Test 2: ignore parenthesis inside regex, and saving groups
```
    $name = &random("nipodemos", "henrybk", "lututui", "alisonRag")
    if ($name =~ /(\w+)/ ) {
        log it worked, choosen name: '$.lastMatch1'
    }
```
image

### Test 3: return opposite result using `!`
```
    if (! simple text =~ /foo/) {
        log opposite result: sucess
    }
```
image

### Test 4: quotes inside regex being ignored
```
    $name = ni"po"de"mos
    if ($name =~ /ni"po"de"mos/) {
        log quote inside regex being ignored: sucess
    }
```
image

### Test 5: statement with only one arg
    if (&config(troll_face)) {
        #never true
    } else {
        log since this config return an empty string, is false
    }
image

```
    do c \@storage
    do i nu
    do conf debug 3
    $i = 1
    if ($.storageopen && $i && &invamount(Savage Meat) > 200) {
        log storage is open, \$i is defined, and we have more than 200 hundred Savage Meat
    }
```
image

### Test 6: Complex if statements
    if (($i = 1 || $i < 5 || $i ~ 0 .. 5) && (1 = 1 || 2 = 2) && (&invamount(Savage Meat) > 1 || &invamount(Savage Meat) < 1) && &eval(&eval($i - 1) - &eval($i - 0)) = -1) {
        log oi
    }
image


lututui:
switch/case is not default perl syntax and Switch module is neither in our dependencies nor included in start.exe and wxstart.exe

```Can't locate Switch.pm in @INC (@INC contains: plugins/eventMacro src src/deps
F:/Users/lututui/Documents/GitHub/openkore/src/deps
F:/Users/lututui/Documents/GitHub/openkore/src F:/Users/lututui/Documents/GitHub/openkore .)
at plugins/eventMacro/eventMacro/Runner.pm line 1213.
BEGIN failed--compilation aborted at plugins/eventMacro/eventMacro/Runner.pm line 1213.
Compilation failed in require at plugins/eventMacro/eventMacro/Core.pm line 15.
BEGIN failed--compilation aborted at plugins/eventMacro/eventMacro/Core.pm line 15.
Compilation failed in require at
F:/Users/lututui/Documents/GitHub/openkore/plugins/eventMacro/eventMacro.pl line 17.
BEGIN failed--compilation aborted at
F:/Users/lututui/Documents/GitHub/openkore/plugins/eventMacro/eventMacro.pl line 17.
```

Nipodemos:
image

@lututui  :+1:

PR #2424, opened 2018-10-19
iRO Renewal packet update

Fadreus:
In case @allanon running late (again).


itsrachelfish:
@Fadreus Thanks for this update! It looks like allanon has been busy lately. In the future, would it be possible for you to update all of these files in a single commit? In the mean time, I'll squash and merge.

Fadreus:
Aah, right. Will do that.
image

Hopefully daddy allanon be back soon.

PR #2427, opened 2018-10-20
Modified server details for ROPH/pRO

junners:
Credits to lorwyn2007 stated in [OpenKore Forums](http://forums.openkore.com/viewtopic.php?p=281101#p281101)

### DIFF
--- a/tables/servers.txt
+++ b/tables/servers.txt
@@ -245,11 +245,11 @@ gameGuard 0
 
 # https://ragnarokonline.com.ph/main
 [Philippines - pRO: Valkyrie]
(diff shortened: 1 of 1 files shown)


Nipodemos:
but in the same post he said that these configurations are not working
image

junners:
AFAIK the culprit is CDClient.dll; I wanted to keep everyone up with the correct server details untill someone figures out how to bypass CDClient.dll. 

I'm also trying to hex the current ragexe client and unload the dlls using NEMO and test whether this works on xkore 1. However I don't have the  proper resources to do this since i'm running on Linux.

PR #2429, opened 2018-10-20
Wx , the orange dot

c4c1n6kr3m1:
update Wx.pm for the orange dot , the portals
the orange dot nailed on tables\portals.txt 
but not anymore , i think start3 hook is a better place


c4c1n6kr3m1:
the orange dot bind to tables\portals.txt
try delete or rename your tables\portals.txt and see what happen to the orange dot

c4c1n6kr3m1:
remove your tables\portals.txt and the orange dot gone

Fadreus:
@c4c1n6kr3m1 Any screenshot ?

c4c1n6kr3m1:
i'm too lazy to create ss :D

i'll make it clear
you have localize tables\yourserver\portals.txt
an example , you play bRO , you have tables\bRO\portals.txt
yes your bot will use this one
but...
wx, the orang dot always come from tables\portals.txt
you will realize the buggy orange dot if you play renewal and classic from the same openkore

Nipodemos:
just tested, it is working
image

Nipodemos:
for a possible future, it would be nice do differentiate npcs that can teleport (kafras and so on) from actual portals, 
because on this map, there is no difference

PR #2449, opened 2018-10-25
add warning log and error log to eventMacro

Nipodemos:
Before, if you want to show a log with red colors you would have to do this:
```perl
do eval Log::error "This is an error message\n"
```

now you can simply do this
`error This is an error message`

automacro that it can be used to test:
```
automacro test {
    BaseLevel > 0
    exclusive 1
    run-once 1
    call {
        log this is a normal log message
        warning this is a warning message
        error this is a error message
    }
}
```
Output:
image


Nipodemos:
retested and working

PR #2455, opened 2018-10-26
iRO Renewal Packets Update

Fadreus:
Poke @allanon


Fadreus:
Daddy allanon has forsaken us. :sob:

bunnyy00:
Thanks! It works

pelthor:
Iro not working still,

Nipodemos:
you problably did not update the openkore the right way

PR #2460, opened 2018-10-28
[eventMacro] new condition: SystemMsg

Nipodemos:
In some cases, npcs send broadcasts on the map, like on quests or challenges

Some times it is used the local_broadcast, but sometimes are used the system broadcast (yelllow sentence on top of game)

This condition is to when system broadcast is used
[](url)
Tested on [izlude Arena](https://browiki.org/wiki/Arena_de_Batalha)

### DIFF
--- a/plugins/eventMacro/eventMacro/Condition/SystemMsg.pm
+++ b/plugins/eventMacro/eventMacro/Condition/SystemMsg.pm
@@ -1,0 +1,32 @@
+package eventMacro::Condition::SystemMsg;
+
+use strict;
+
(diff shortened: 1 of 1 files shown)


Nipodemos:
yeah, is just a copy of LocalMsg but with the different hook :smile:

ya4ept:
please add a description to the documentation

Nipodemos:
@ya4ept i don't have access to international documentation, only brazilian. Can you give me access?

ya4ept:
please refer to @lututui

PR #2462, opened 2018-10-29
iRO Re:Start Packets Update

Fadreus:
- [x] Tested


Nipodemos:
this new avatar of yours @Fadreus is you when reading issues like "i can't connect to iRO please help"

Fadreus:
More like-;
image

PR #2464, opened 2018-10-29
[eventMacro]Improve reuse of code

Nipodemos:
### Changes:
**IsInCoordinate** and **IsNotInMapAndCoordinate** now are childs of **IsInMapAndCoordinate**

**IsInCoordinate** now have only the `sub parse_syntax` , the change is to accept only coordinates
**IsNotInMapAndCoordinate** now have only the `sub check_location` the change is to be true only if you not in any of the specified maps/coodinates. So, if the condition is like this: `IsNotInMapAndCoordinate prontera, morocc, geffen, aldebaran, yuno` , the condition will only be true if you not in any of these maps
(but i wonder if it is correct to keep it like this)

**NotInMap** Same of last condition

### DIFF
--- a/plugins/eventMacro/eventMacro/Condition/IsInCoordinate.pm
+++ b/plugins/eventMacro/eventMacro/Condition/IsInCoordinate.pm
@@ -1,7 +1,6 @@
 package eventMacro::Condition::IsInCoordinate;
 
 use strict;
-use Globals qw( $char $field );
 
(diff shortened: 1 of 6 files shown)


Nipodemos:
i'll just wait for HenryBk opinion about this, then i'll merge

Nipodemos:
with this new commit, it now follows the things he said, so i'll merge it

PR #2475, opened 2018-11-02
Make a check for when have more than 30000 of same item

Nipodemos:
### Things to discuss:
* i think that variable `$target_item` have a very bad name (i've didn't think too much) and i'm looking for some tip to give a better name.
that variable store the item object on the target inventory

* Also, if i use `$row->{item}`, shows the name of item along with it's index. If i use `$row->{item}->name` the name show without the index of it
Which one you guys think is better?

### Before: 
image

### After:
image

### When already have 30,000 of the item and you try to add more
image

#### Additional thing:
I removed a little bit of repetitive code on Commands.pm

**Thanks to @lututui for provided me a better sentence to use on error message**

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1029,7 +1029,7 @@ sub processTransferItems {
 			my ( $name, $list ) = @$_;
 			next if $list->isReady;
 			#name can be storage, inventory or cart
-			error T( "Your $name is not available. Unable to transfer item '%s'.\n", $row->{item} );
+			error TF( "Your %s is not available. Unable to transfer item '%s'.\n", $name, $row->{item} );
 			redo;
 		}
 
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1037,7 +1037,7 @@ sub processTransferItems {
 		my $item = $source->get( $row->{item}->{binID} );
 		
 		if ( !$item || $item->nameString ne $row->{item}->nameString ) {
-			error TF( "$row->{source} item '%s' disappeared!\n", $row->{item} );
+			error TF( "%s item '%s' disappeared!\n", $row->{source}, $row->{item} );
 			redo;
 		}
 
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1069,11 +1069,11 @@ sub processTransferItems {
 		my $stack_limit = $target->item_max_stack( $row->{item}->{nameID} );
 		
 		my $amount = min($stack_limit, min( $item->{amount}, $row->{amount} || $item->{amount} ));
-		if ( ( $stack_limit - $amount ) < 0 || ($target_item->{amount} - $stack_limit) == 0) {
+		if ( $stack_limit - $amount < 0 || $target_item->{amount} - $stack_limit == 0) {
 			error TF("Unable to add %s to %s. You can't stack over %s of this item\n", $item->name, $row->{target}, $stack_limit);
 			redo;
 			
-		} elsif ( ($target_item->{amount} + $amount) > $stack_limit ) {
+		} elsif ( $target_item->{amount} + $amount > $stack_limit ) {
 			warning TF("Amount of %s will surpass the maximum %s capacity (%d), transfering maximum possible (%d)\n",
 			$row->{item}->name, $row->{target}, $stack_limit, $stack_limit - $target_item->{amount} );
 			
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1068,19 +1068,17 @@ sub processTransferItems {
 		my $target_item = $target->getByName( $row->{item}->{name} );
 		my $stack_limit = $target->item_max_stack( $row->{item}->{nameID} );
 		
-		my $amount;
-		if (( $stack_limit - $target_item->{amount} ) <= 0) {
+		my $amount = min($stack_limit, min( $item->{amount}, $row->{amount} || $item->{amount} ));
+		if ( ( $stack_limit - $amount ) < 0 || ($target_item->{amount} - $stack_limit) == 0) {
 			error TF("Unable to add %s to %s. You can't stack over %s of this item\n", $item->name, $row->{target}, $stack_limit);
 			redo;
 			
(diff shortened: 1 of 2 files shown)


src/AI/CoreLogic.pm
```diff
@@ -1028,18 +1028,16 @@ sub processTransferItems {
 		foreach ( [ $row->{source}, $source ], [ $row->{target}, $target ] ) {
 			my ( $name, $list ) = @$_;
 			next if $list->isReady;
-			error T( "Your inventory is not available. Unable to transfer item '%s'.\n", $row->{item} ) if $name eq 'inventory';
-			error T( "Your storage is not available. Unable to transfer item '%s'.\n",   $row->{item} ) if $name eq 'storage';
-			error T( "Your cart is not available. Unable to transfer item '%s'.\n",      $row->{item} ) if $name eq 'cart';
+			#name can be storage, inventory or cart
+			error T( "Your $name is not available. Unable to transfer item '%s'.\n", $row->{item} );
```
lututui: Interpolation with T (translation method) is forbidden.
It should go like this:
```suggestion
			error TF( "Your %s is not available. Unable to transfer item '%s'.\n", $name, $row->{item} );
```

Now that I look at it, the old way is also wrong, and should be using TF instead of T

src/AI/CoreLogic.pm
```diff
@@ -1028,18 +1028,16 @@ sub processTransferItems {
...
 		}
 
 		# Verify that the item is still in the source list and has not changed.
 		my $item = $source->get( $row->{item}->{binID} );
+		
 		if ( !$item || $item->nameString ne $row->{item}->nameString ) {
-			error TF( "Inventory item '%s' disappeared!\n", $row->{item} ) if $row->{source} eq 'inventory';
-			error TF( "Storage item '%s' disappeared!\n",   $row->{item} ) if $row->{source} eq 'storage';
-			error TF( "Cart item '%s' disappeared!\n",      $row->{item} ) if $row->{source} eq 'cart';
+			error TF( "$row->{source} item '%s' disappeared!\n", $row->{item} );
```
lututui: Same thing, interpolation is forbidden.
```suggestion
			error TF( "%s item '%s' disappeared!\n", $row->{source}, $row->{item} );
```

src/AI/CoreLogic.pm
```diff
@@ -1059,16 +1057,25 @@ sub processTransferItems {
...
 			error TF( "Inventory item '%s' is equipped.\n", $item->{name} );
 			redo;
 		}
-
+		
+		my $target_item = $target->getByName( $row->{item}->name );
+		
 		# Transfer the item!
-		$messageSender->$method( $item->{ID}, min( $item->{amount}, $row->{amount} || $item->{amount} ) );
+		if ((30000 - $target_item->{amount} ) <= 0) {
```
lututui: Is 30k ALWAYS the max stack for EVERY SINGLE stackable item?

src/AI/CoreLogic.pm
```diff
@@ -1059,16 +1057,25 @@ sub processTransferItems {
...
+			error TF("Unable to add %s to %s. You can't stack over 30,000 of the same item\n", $item->name, $row->{target});
+		} elsif ( (30000 - $target_item->{amount} ) < min( $item->{amount}, $row->{amount} ) ) {
```
lututui: min(a, b) is a List::Util method. Did you import it?

src/Commands.pm
```diff
@@ -4761,15 +4768,9 @@ sub cmdStorage {
 }
 
 sub cmdStorage_add {
-	my $items = shift;
+	my $name = shift;
+	my $amount = shift;
```
lututui: Maybe
```perl
my ($name, $amount) = @_;
```
just to keep a standard?

src/Commands.pm
```diff
@@ -4780,20 +4781,14 @@ sub cmdStorage_add {
 }
 
 sub cmdStorage_addfromcart {
-	my $items = shift;
+	my $name = shift;
+	my $amount = shift;
```
lututui: Maybe
```perl
my ($name, $amount) = @_;
```
just to keep a standard?

src/Commands.pm
```diff
@@ -4804,15 +4799,9 @@ sub cmdStorage_addfromcart {
...
-	my ( $name, $amount );
-	if ( $items =~ /^[^"'].* .+$/ ) {
-		# Backwards compatibility: "storage get Empty Bottle 1" still works.
-		( $name, $amount ) = $items =~ /^(.*?)(?: (\d+))?$/;
-	} else {
-		( $name, $amount ) = parseArgs( $items );
-	}
+	my $name = shift;
+	my $amount = shift;
+
```
lututui: Maybe
```perl
my ($name, $amount) = @_;
```
just to keep a standard?

src/Commands.pm
```diff
@@ -4823,20 +4812,14 @@ sub cmdStorage_get {
 }
 
 sub cmdStorage_gettocart {
-	my $items = shift;
+	my $name = shift;
+	my $amount = shift;
```
lututui: Maybe
```perl
my ($name, $amount) = @_;
```
just to keep a standard?

src/AI/CoreLogic.pm
Nipodemos: That's a very good question, which i don't have na answer 😢

cydh: all default for stackable items are 30000
imho, I prefer we make a new table file (can be eac server later) to use max stack limit, example
```
-1 30000
<itemid> <stack_limit>
```
the stack info for now, RK's runes are 60 in inventory and 3 for AB's Ancilla in inventory
source: https://github.com/rathena/rathena/blob/master/db/re/item_stack.txt

c4c1n6kr3m1: that's for inventory right? this for stackable storage, so no need a new table file right now ,i think

Nipodemos: @c4c1n6kr3m1 actually this piece of code is for any type of transference, 
* inventory to storage
* inventory to cart
* cart to inventory and so on

c4c1n6kr3m1: my bad, didn't realize there is new processTransferItems

Nipodemos:
tests made right now:

### get item without set amount:
image

### get item WITH amount defined
image

### when amount is already 30k:
image

### when everything is ok (no error nor warning)
image

PR #2486, opened 2018-11-06
Updates Hateffect Tables

cydh:
* from kRO 2018-10-31


lututui:
Are those grf files? If so, what are they originally named?

alisonrag:
@lututui 
last time i checked this file only exists in kRO 
data/luafiles514/lua files/hateffectinfo/hateffectinfo.lua

PR #2493, opened 2018-11-11
Storage/inventory/cart desc change

Nipodemos:
Now it shows the amount of the item in description, so it can be easier to check the amount of an especific item (which can be hard when you have 200+ items on you storage)

### this is more a stetical change
image

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -2450,40 +2450,15 @@ sub positionNearPortal {
 }
 
 ##
-# printItemDesc(int itemID, string location)
-# $location can only be: 'inventory', 'cart', 'store' or 'storage'
+# printItemDesc(obj $item)
 #
-# Print the description for $itemID.
+# Print the description for $item.
 sub printItemDesc {
-	my $itemID = shift;
-	my $location = shift;
-	
-	my $itemName = itemNameSimple($itemID);
-	my $amount;
-	if ($location eq 'store') {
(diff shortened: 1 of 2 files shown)


src/Misc.pm
```diff
@@ -2450,14 +2450,15 @@ sub positionNearPortal {
...
 sub printItemDesc {
-	my $itemID = shift;
-	my $itemName = itemNameSimple($itemID);
-	my $description = $itemsDesc_lut{$itemID} || T("Error: No description available.\n");
-	message TF("===============Item Description===============\nItem: %s (ID: %s)\n\n", $itemName, $itemID), "info";
+	my $item = shift;
+		
+	my $description = $itemsDesc_lut{$item->{nameID}} || T("Error: No description available.\n");
+	message T("===============Item Description===============\n");
+	message TF("Item: %s, ID: %s, Amount: %s\n\n", $item->{name}, $item->{nameID}, $item->{amount}), "info";
```
Fadreus:
```suggestion
	message TF("Item: %s, ID: %s, Amount in storage: %s\n\n", $item->{name}, $item->{nameID}, $item->{amount}), "info";
```

Since there is no description, adding like this might let people know what's this amount referring to or nvm :thinking:

lututui: printItemDesc is not inventory-specific, it can be used for any stackable anywhere: storage, guild storage, cart, player inventory, vending, etc

Issue #2466, opened 2018-10-30
minStorageZeny doesn't work


midoom:
hello, minStorageZeny doesn't work and make a loop on npc storage 

> storageAuto 1
> storageAuto_npc moc_ruins 59 157
> storageAuto_standpoint
> storageAuto_distance 2
> storageAuto_npc_type 3
> storageAuto_npc_steps c r0 n
> storageAuto_password
> storageEncryptKey
> storageAuto_keepOpen 0
> storageAuto_useChatCommand
> storageAuto_useItem
> storageAuto_useItem_item
> storageAuto_notAfterDeath 1
> relogAfterStorage 0
> minStorageZeny 100

 minStorageZeny [<amount>]
Disables auto storage when your zeny is less than specified.

lututui:
minStorageZeny prevents auto storage from triggering.
I believe what's happening here is that you have enough zeny, auto storage triggers, and then the character loses some (due to teleport service, auto buy, or any other reason), putting us below minStorageZeny - but auto storage is already in the ai queue.
@midoom Do you agree? Or is something else happening?

midoom:
Yes that's it

Henrybk:
maybe we should add minStorageZeny in the routing zeny calculation and also add another one before talking to the npc, so this loop never happens

PR #2500, opened 2018-11-15
Prevent loop when trying to open storage...

lututui:
and failing because we don't have enough zeny

Workaround #2466

### DIFF
--- a/src/AI.pm
+++ b/src/AI.pm
@@ -51,6 +51,7 @@ our @EXPORT = (
 	ai_skillUse
 	ai_skillUse2
 	ai_storageAutoCheck
+	ai_canOpenStorage
 	cartGet
 	cartAdd
 	ai_talkNPC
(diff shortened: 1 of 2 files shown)


cydh:
Can also add a config maybe to ignore zeny check as long as player has `Storage Free Ticket` (7059), a config maybe.. well or maybe on separated PR, also for Zeny (7060) on 'normal' warp (not airport), rent cart too (7061)

c4c1n6kr3m1:
- all job has NV_BASIC not only novice and high novice 
- doram has different basic skill

c4c1n6kr3m1:
-    doram has different basic skill
-    other job has NV_BASIC not only novice and high novice

lututui:
@cydh Added the storage ticket. 

About the warp thing, I'm pretty sure that already is taken into account: kore won't pick a route if we don't have enough budget to complete it.

About rent cart, is that event a problem? I don't think there's an "auto rent cart" feature

Nipodemos:
@lututui https://www.ragnaplace.com/bRO/item/7061/Vale-Carrinho

as he said, this part should be on another PR

PR #2426, opened 2018-10-19
Rodex send fix disconnect from map server on GGamerz RO.

CoffeeParser:
Issue: When using the rodex system on the private server GGamerz RO the openkore client always disconnected from the map server and no rodex mails could be sent.

Found bugs: send and receive packet descriptions were mismatched when I looked into them on wireshark. The send logic also has formed packets that were not supported by the server.
 
Proposed solution: 
In order to fix the issue I modified  send and receive files of RagexeRE_2015_11_04a.pm. 
 
Details: 
The functionality has been tested and works as intended without crashes on the private server GGamerz RO (http://ggamerz-ro.com/). The rodex system works and no further disconnects from the map server occur. 

this fixes #2441

PR #2530, opened 2018-12-02
Issues/2441

sctnightcore:
base code : https://github.com/OpenKore/openkore/pull/2426
ty @CoffeeParser

need to be test !

### DIFF
--- a/src/Network/Send/kRO/RagexeRE_2015_05_13a.pm
+++ b/src/Network/Send/kRO/RagexeRE_2015_05_13a.pm
@@ -13,6 +13,8 @@ package Network::Send::kRO::RagexeRE_2015_05_13a;
 
 use strict;
 use base qw(Network::Send::kRO::RagexeRE_2014_10_22b);
+use Globals qw($char $rodexWrite);
+use I18N qw(stringToBytes);
 
 sub new {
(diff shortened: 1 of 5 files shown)


src/Network/Send/kRO/RagexeRE_2015_05_13a.pm
```diff
@@ -85,4 +89,25 @@ sub new {
...
 }
+
+sub rodex_send_mail {
+	my ($self) = @_;
+
+	my $title = stringToBytes($rodexWrite->{title});
+	my $body = stringToBytes($rodexWrite->{body});
+	my $pack = $self->reconstruct({
+		switch => 'rodex_send_mail',
+		receiver => $rodexWrite->{target}{name},
```
sctnightcore: $rodexWrite->{name} ?

sctnightcore: fixed : 85c1356

Nipodemos:
the only problem is to test
what server use that ragexeRE you edited?

sctnightcore:
I have tested my server 
@nipo

Nipodemos:
fair enough

PR #2539, opened 2018-12-07
allow regex matching for pickupitems.txt

### LINKED ISSUE #2110: [items_control] [mon_control] [pickupitems] Allow regex match
Implement #1364, but with a more efficient approach to it.

Nipodemos:
try to fix #2110
DO NOT MERGE YET

for now is just a concept, for you guys to see and evaluate

the ideia here is:
instead of make a hole new hash for each regex found,
or loop the entire pickupitems hash for each item,
i save the regexes on a hash and then make a comparison of the item name with each regex

I am open to ideias of how to improve this, i know is not good yet

TODO LIST:
- [ ] Test more with more items 
- [ ] Find a better name for variables


Nipodemos:
i was thinking in maybe create only one hash that will save the regexes of mon_control, and keep a structure like this:
```perl
%regexes = {
    'mon_control' => {
        'regex1' => '1 0 1',
        'regex2' => '0 0 0 4'
    },
    'items_control' => {
        'regex1' => '1 0 1',
        'regex2' => '0 0 0 4'
    },
    'pickupitems' => {
        'regex1' => '0',
        'regex2' => '2 '
    }
};
```
what do you guys  think?

lututui:
This is still VERY BAD, complexity-wise

Nipodemos:
i am reading about complexity-wise but is still hard for me to understand
@lututui could you explain the problems on my approach?

lututui:
nvm i've read it wrong
it's quite ok actually

lututui:
This really should be in a plugin

Nipodemos:
okay i'll work in that today at night

Nipodemos:
after a small talk, i decided that is better to convert that into a plugin. When i have free time i will do this

PR #2544, opened 2018-12-11
[ Receive ] Update and Merge vender_items_list

alisonrag:
* Update on vender_items_list to avoid manual byte skip.
* Merge vender_items_list.

tested in iRO and local server (rAthena based)

image

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -3040,6 +3040,45 @@ sub vending_start {
 	$shopEarned ||= 0;
 }
 
+sub vender_items_list {
+	my ($self, $args) = @_;
+
+	my $item_pack = $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
+	my $item_len = length pack $item_pack;
+	my $item_list_len = length $args->{itemList};
+
+	my $player = Actor::get($args->{venderID});
+
+	$venderItemList->clear;
+
+	message TF("%s\n" .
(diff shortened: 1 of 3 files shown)


src/Network/Receive.pm
```diff
@@ -3040,6 +3040,45 @@ sub vending_start {
 	$shopEarned ||= 0;
 }
 
+sub vender_items_list {
+	my ($self, $args) = @_;
+
+	my $item_pack = $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
```
sctnightcore: add `$self->{vender_items_list_item_pack} = 'V v2 C v C3 a8 a25';` to `RagexeRE_2015_05_13a`

alisonrag: https://github.com/OpenKore/openkore/pull/2346/files

already added and checked

this pull is only to avoid manual unpacking and manual skip

i removed and updated this
```
my $msg = $args->{RAW_MSG};
	my $msg_size = $args->{RAW_MSG_SIZE};
	my $headerlen;
	my $item_pack = $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
	my $item_len = length pack $item_pack;

	# a hack, but the best we can do now
	if ($args->{switch} eq "0133") {
		$headerlen = 8;
	} else { # switch 0800
		$headerlen = 12;
	}
```

PR #2547, opened 2018-12-12
fix pickupitems parse

### LINKED ISSUE #2543: xconf or pickupitems not working as it should
xconf (pconf) is adding Item ID in pickupitems.txt, but Kore is not picking up items unless I put the name, like Orc Claw instead of 1043 (Orc Claw ID).

When I use pconf Orc Claw 2, it goes to pickupitems as 1043 2 # Orc Claw, but it wont pick it, unless i put the name.

100% reproduction.

I dont know if the problem is with xconf inserting item as ID, or is at pickupitem doesnt picking ID numbers.

https://i.imgur.com/HiawmTe.png

Nipodemos:
fix #2543
### before:
cmd_qg4tdmisi2

### after:
cmd_r2y6mxog3m

### DIFF
--- a/src/FileParsers.pm
+++ b/src/FileParsers.pm
@@ -31,6 +31,7 @@ use Text::Balanced qw(extract_delimited);
 use Utils;
 use Utils::TextReader;
 use Plugins;
+use Settings;
 use Log qw(warning error debug);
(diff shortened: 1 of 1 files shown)


Nipodemos:
@lututui done

PR #2563, opened 2018-12-26
Updates idRO_Renewal

cydh:
* Changes server name in servers.txt to Odin
* Item updates for MT 2018-12-18


cydh:
seems I forgot to create PR for last week maintenance

PR #2568, opened 2018-12-31
add gameGuard 0 line in config.txt

Nipodemos:
in my opinion, if poseidon is a built-in tool on openkore, then we should let the config built-in on config.txt too

but since only 1 server (that i know) will use this config, make disabled by default


lututui:
really unnecessary but harmeless

PR #2572, opened 2019-01-03
Packets Update - iRO

Fadreus:
- [x] iRO - Re:Start
- [ ] iRO - Renewal

_New year sickness.
Still writing 2018 in 2019._


alisonrag:
Tks <3

Issue #1526, opened 2017-12-21
Autobuy -> Overweight


Fvegini:
###### Summary: 

Kore keeps trying to "autobuy" when weight is full instead of do an autosell and only then do the autobuy

###### Affected configuration(s)/ file(s):
config.txt
###### Impact:
Openkore keeps stucked until manually insert an "autosell" or "autostorage" command
###### Expected Behavior:
When kore got the error from autobuy, it should try to release from the items exceed, selling or storaging then (if none are configurated, then it should keep stucked)
###### Actual Behavior:
Kore jeeps stuck in the autobuy routine

![peso](https://user-images.githubusercontent.com/32243959/34247455-b3ff3254-e619-11e7-8825-5de84b97b478.png)

sctnightcore:
Use with New openkore ?

Fvegini:
Maybe one version from two weeks ago!

I will update and try to simulate it again to see if keeps hapening!

Fvegini:
Same issue with the last version of openkore.

Kore is with 68% of weight trying to buy red potions.

I think the config is alright:

buyAuto Red Potion {
	npc payon_in01 5 49
	standpoint
	distance 5
	price 50
	minAmount 0
	maxAmount 300
	batchSize
	disabled 0
	label buyredpot
}


If I manually send him "autosell", he sells, return to the autobuy routine and go to lockmap

fadreus:
I believe this happen when;

You loot items > Then red pot reach minimum > bot autoBuy > your loots in the way = prevent buying 300 red pot.

If this is the case, I wouldn't say it's a bug or problem on kore (unless kore suppose to ignore autobuy when exceed weight).

Fvegini:
fadreus, this seems to be the exactly scenario when it happens!

In my point of view openkore don't use the weight of the items.. so it can't ignore the autobuy when it checks is "full"
But just adding an autosell/autostorage after the prompted message would be enough

One bad way I can think in do it, would be: (bad programming skills)

src/Network/Receive/ServerType0.pm Line 1352

```
} elsif ($args->{fail} == 2) {
	error T("Buy failed (insufficient weight capacity).\n")
}
```
to

```
} elsif ($args->{fail} == 2) {
	error T("Buy failed (insufficient weight capacity).\n")
	if (Autosell == 1) do autosell
	if (Autostorage ==1) do autostorage
}
```

[...]

PR #2595, opened 2019-01-22
Autosell when overweight

piichia:
#1526


piichia:
uhh I didn't think it would add my other updated fork files to this pull. I'll try fixing it

piichia:
Ok this is ridiculous I'll just reopen a new pull and not mess with things zzz

PR #2621, opened 2019-02-11
Move identical network functions to receive.pm

Henrybk:
Mostly identical functions, sub changeToInGameState was also removed from both servertype0 and sakexe_0.

vender_found had a difference in the hook calls.

That is all I remember that was different.

PR #2637, opened 2019-02-19
Follow up to #2621

lututui:
Bareword under strict subs


Fadreus:
image

Hmmm.. :thinking:

Fadreus:
image

Which is which?

PR #2642, opened 2019-02-22
System Chat & Local Broadcast

cydh:
* Added `/s` for system chat regex so can be match with newline character. In some cases, the padded chars can be newline and will ignores all regex in ifs
* Uncommented `logLocalBroadcast` config in `sub local_broadcast`

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -318,6 +318,7 @@ logPrivateChat 1
 logPartyChat 1
 logGuildChat 1
 logSystemChat 1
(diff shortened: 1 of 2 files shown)


alisonrag:
I think that you should add logLocalBroadcast to config

ya4ept:
Do not forget to add the description to the wiki

alisonrag:
http://openkore.com/index.php/LogLocalBroadcast

PR #2649, opened 2019-02-27
update iRO Re:Start files

allanon:
- [ ] QA Review


lulinha93:
Doesnt work when i run wxstart the next text appears:

Packet: Tokenizer: Switch desconhecido: 0069
Packet: Tokenizer: Switch desconhecido: F180

it happened when trying to log on account servers

PR #2652, opened 2019-03-02
Update equiptypes

cydh:
* Added missing costume garment/robe type as "Visual Cape"

**previous**
image
**now**
image


Nipodemos:
great work :D

PR #2677, opened 2019-03-18
teleportAuto_item accept NameID

OvoKore:


c4c1n6kr3m1:
limitation to 3 digit numbers or more ?

c4c1n6kr3m1:
if the items name is "blablabla"
than you will search it from all your inventory ( getByNameID ), for what ?

limit it to digit numeric
```
/^\d{3,}$/
```

PR #2685, opened 2019-03-21
add maxBase and minBase in self condition

OvoKore:


lututui:
This should be implemented as a self condition, not something buyAuto-specific

Nipodemos:
agree

PR #2692, opened 2019-03-24
Add LWP::UserAgent to openkore

sctnightcore:


lututui:
Why?

itsrachelfish:
I agree. Why? Is there a specific plugin you are using that requires this?

alisonrag:
kRO and ruRO uses authentication based in web token
https://github.com/alisonrag/Openkore-Plugins-Scripts/tree/master/ZeroAuth

lututui:
Package it with the plugin, kore doesn't need it

Nipodemos:
as mentioned, it is better to make a standalone plugin and install this package manually, it's not that hard

PR #2710, opened 2019-04-04
update iRO Re:Start files

allanon:
- [ ] QA Review


Fadreus:
These updates were from #2684

jourlez:
Did not work for me also.

PR #2719, opened 2019-04-10
fix min/maxBase in buyAuto and accept nameID

OvoKore:


Nipodemos:
what gets returned to `$amount` if user type an invalid ID or name?
depending, this should be considered

PR #2721, opened 2019-04-10
[README] FAQ Added.

Fadreus:
Tried to make it simple for anyone to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,12 +3,12 @@
 * OpenKore is an custom client & intelligent automated assistant for Ragnarok Online.
 * It is a **free**, open source & cross-platform program _(Linux, Windows & MacOS are supported)_.
 
-## Prerequisites.
+## Prerequisites
 
 To run OpenKore you will need:
 * [Read the Requirements page on our wiki](http://wiki.openkore.com/index.php/How_to_run_OpenKore#Requirements)
 
-## Quickstart.
+## Quickstart
 
 1. [Download OpenKore](https://github.com/OpenKore/openkore/archive/master.zip) & extract it. Alternatively, you could press the **Windows Key + R**, type in ``cmd`` & enter. Run the following command in the cmd to clone.
 ***Note: [Git](https://git-scm.com/) required.***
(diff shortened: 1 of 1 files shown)


Nipodemos:
hello
i just wonder why to put a dot in every title, why?

README.md
```diff
@@ -1,40 +1,56 @@
...
-
-* Before submitting an issue, ensure you are running the latest commit on branch.
-* Check if an issue already exists for your request.
+3. Run openkore.pl _(You can run start.exe or wxstart.exe if you use Windows)_.
+
+## F.A.Q. (Frequently Asked Questions).
+<!-- Source: http://forums.openkore.com/viewtopic.php?f=0&t=11287 -->
+ 1. **Have a problem?**
+    - Update your openkore.
+    - Download anew.
```
sctnightcore: new .

Fadreus: It's literally mean the same but with proper English.

Fadreus:
@Nipodemos To complete a sentence? :no_mouth:
Some are single word, some are in phrase. Having just some title with fullstop doesn't look nice.
Having fullstop on every title make it look uniform.
Or change all title in capital letters? :thinking: 

It just doesn't seems right IMO.

README.md
Nipodemos: i though that it was wrong too hahahah
i had to reasearch to know that is correct

sctnightcore: I think use `a new` ?

Nipodemos:
@Fadreus titles are not considered as phrases, so they don't need a dot in the end (i think they even can't have)

i search a few README's on GitHub to see the pattern:
[README of Flutter, a framework made by all mighty Google](https://github.com/flutter/flutter/blob/master/README.md)
[README of VS Code, a code editor made by Microsoft](https://github.com/Microsoft/vscode/blob/master/README.md)
[README of React Native, a framework made by Facebook](https://github.com/facebook/react-native/blob/master/README.md)

in all three repositories, all titles don't have dot

Fadreus:
> titles are not considered as phrases, so they don't need a dot in the end (i think they even can't have)

Fair Enough

Changed as requested. :ok_hand:

PR #2741, opened 2019-04-22
[eventMacro] Move Base Conditions to it's own folder

Nipodemos:
this is more a stetical change
with this i was able to remove a little bit of boilerplate code (all sub usable in all condition)
and maintain things more organized


plugins/eventMacro/eventMacro/Condition.pm
```diff
@@ -114,9 +114,4 @@ sub is_unique_condition {
 	0;
 }
 
-# Default: true
```
lututui: Why?

This hurts the concept of an abstract class

Nipodemos: it's because this class will not exist anymore with this Pull Request

Nipodemos: it's because this class will not exist anymore with this Pull Request

PR #2763, opened 2019-05-15
update iRO Restart files

allanon:
- [ ] QA Review

This pull request was auto-generated by patch_iro.pl: https://gist.github.com/allanon/38620af49d250005acbf


Nipodemos:
i don't know if this PR is correct :thinking:

Fadreus:
Both event already ended.
Is it necessary? :thinking:

Nipodemos:
considering it is automatic, maybe the restart server removed the info of only one of the quests. Since both quests have ended, i think its better to merge this

PR #2780, opened 2019-06-02
Add batchSize to getAuto logic

Henrybk:
Problem:
If getAuto is set to get a non-stackable item and the diference between the minAmount and maxAmount is greater than 1 openkore will try to get more than 1 of the item at the same time. This is not possible. In some servers it will just not work, on others it will disconnect the char, in some it may even get you banned.

Proof:
config:

result:

Solution:
I added batchSize logic from buyAuto to getAuto, it should now work if properly set up.

Proof:
config:

result:

Tested on RMS renewall test server.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -738,6 +738,7 @@ npcWrongStepsMethod 0
 getAuto {
 	minAmount
 	maxAmount
+	batchSize
 	passive
 	disabled 0
 }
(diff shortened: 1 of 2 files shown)


ya4ept:
Please add a description of the option to the documentation.

Henrybk:
Documentation added

PR #2781, opened 2019-06-02
Event macro slot system

Henrybk:
Adding a way for eventMacro to run multiple macros at the same time has always been something I wanted but I never made. Today I decided to come up with a simple system to make it work, this is still very much a work is progress (as the many "TODO" messages may indicate).

There are some design decisions I have want/need help with.

How should we handle the AI queue while we have 1 or more macros running? What if one is overrideAI 1 and the others are not? What if they all are overrideAI 0 or 1?

How should the commands be changed to accompany these changes?

The code as it is now only supports multiple macros if they are all "overrideAI 1".

Example:

### DIFF
--- a/control/eventMacros.txt
+++ b/control/eventMacros.txt
@@ -3,10 +3,14 @@ automacro test_slot1 {
 	exclusive 1
 	InMap prontera
 	disabled 0
+	overrideAI 1
 	slot 1
 	call {
-		log This is working in slot 1
-		pause 3
+		log This is working in slot 1 start
+		pause 1
+		do c hey
+		pause 1.5
+		log This is working in slot 1 end
 	}
 }
 
@@ -14,10 +18,14 @@ automacro test_slot2 {
 	exclusive 1
 	InMap prontera
 	disabled 0
+	overrideAI 1
 	slot 2
 	call {
-		log This is working in slot 2
+		log This is working in slot 2 start
+		pause 1.5
+		do c there
 		pause 1
+		log This is working in slot 2 end
 	}
 }
 
@@ -25,9 +33,13 @@ automacro test_slot3 {
 	exclusive 1
 	InMap prontera
 	disabled 0
+	overrideAI 1
 	slot 3
 	call {
-		log This is working in slot 3
+		log This is working in slot 3 start
 		pause 2
+		do c man
+		pause 0.5
+		log This is working in slot 3 end
 	}
 }
--- a/control/eventMacros.txt
+++ b/control/eventMacros.txt
@@ -1,0 +1,33 @@
+
+automacro test_slot1 {
+	exclusive 1
+	InMap prontera
+	disabled 0
+	slot 1
+	call {
+		log This is working in slot 1
+		pause 3
+	}
+}
+
+automacro test_slot2 {
+	exclusive 1
+	InMap prontera
+	disabled 0
+	slot 2
+	call {
+		log This is working in slot 2
+		pause 1
+	}
+}
+
+automacro test_slot3 {
+	exclusive 1
+	InMap prontera
+	disabled 0
+	slot 3
+	call {
+		log This is working in slot 3
+		pause 2
+	}
+}
--- a/plugins/eventMacro/eventMacro.pl
+++ b/plugins/eventMacro/eventMacro.pl
@@ -104,6 +104,8 @@ sub parseAndHook {
 	}
 }
 
+
+# TODO: fix the command to work with slots
 sub commandHandler {
 	### no parameter given
 	if (!defined $_[1]) {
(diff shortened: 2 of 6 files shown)


control/eventMacros.txt
```diff
@@ -0,0 +1,45 @@
+
+automacro test_slot1 {
```
sctnightcore: del  control/eventMacros.txt

Henrybk: It is there so people can see how this system would work.

ya4ept:
what is the difference between options "priority" and "slot"?

Henrybk:
@ya4ept Priority hasn`t changed at all. Slots separate macros and automacros into different "instances", as is all macros and automacros using slot 1 will be totally independent from those using slot 2 or 3 or etc.

While a macro is running at slot 1, in slot 2 there can be a few automacros checking for conditions, and another macro running at the same time in slot 7

Nipodemos:
```perl
automacro alpha {
    BaseLevel > 0
    exclusive 1
    slot 1
    call {
        $i = 0
        while ($i < 10) {
            pause 2
            $i++
            log alpha automacro running on slot 1 for &eval(2*$i) seconds
        }
    }
}

automacro beta {
    BaseLevel > 0
    exclusive 1
    slot 2
    timeout 3
    call {
        log beta automacro running on slot 2
    }
}
```
image
worked here with overrideAI 0 on both

marchewka997:
great update!

plugins/eventMacro/eventMacro/Automacro.pm
```diff
@@ -132,6 +132,9 @@ sub set_parameters {
 	if (!defined $self->{parameters}{'repeat'}) {
 		$self->{parameters}{'repeat'} = 1;
 	}
+	if (!defined $self->{parameters}{'slot'}) {
```
allanon: If this is a 1-based counting system (ie, there's no slot zero), you could do:
```
$self->{parameters}{'slot'} ||= 1;
```

plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -67,6 +67,12 @@ sub new {
 		$self->repeat(1);
 	}
 	
+	if (defined $slot && $slot =~ /^\d+$/) {
```
allanon: Should slot `0` be allowed?

plugins/eventMacro/eventMacro/Runner.pm
```diff
@@ -216,48 +222,103 @@ sub overrideAI {
...
-	debug "[eventMacro] Macro '".$self->{name}."' is now deleting itself from AI queue.\n", "eventMacro", 2;
+	debug "[eventMacro] Macro '".$self->{name}."' is now deleting eventMacro from AI queue.\n", "eventMacro", 2;
 	AI::clear('eventMacro');
 	$self->{registered} = 0;
 }
 
+# Adds the current macro to the already existent eventMacro sequence in AI queue
+sub register_add {
+	my ($self, $macro_args) = @_;
+	debug "[eventMacro] Macro '".$self->{name}."' is now adding itself to the eventMacro sequente in AI queue.\n", "eventMacro", 2;
```
allanon: `sequente` => `sequence`

allanon: Make this message match the phrasing in `register_new`? Same for `unregister` and `register_remove`.

plugins/eventMacro/eventMacro.pl
```diff
@@ -196,157 +198,222 @@ sub commandHandler {
...
 	### parameter: check
 	} elsif ($arg eq 'check') {
-		if (!defined $params[0] || (defined $params[0] && $params[0] ne 'force_stop' && $params[0] ne 'force_start' && $params[0] ne 'resume')) {
-			message "usage: eventMacro check [force_stop|force_start|resume]\n", "list";
-			message 
-				"eventMacro check force_stop: forces the stop of automacros checking\n".
-				"eventMacro check force_start: forces the start of automacros checking\n".
-				"eventMacro check resume: return automacros checking to the normal state\n";
-			return;
+        if (!defined $params[0] || (defined $params[0] && $params[0] ne 'force_stop' && $params[0] ne 'force_start' && $params[0] ne 'resume')) {
```
allanon: Tabs vs spaces.

plugins/eventMacro/eventMacro.pl
```diff
@@ -196,157 +198,222 @@ sub commandHandler {
...
+            return;
+        }
+		
+		my $new_status = $params[0];
+       
+        if (defined $params[1]) {
+            my $slot = $params[1];
+			
+            if ($slot !~ /^\d+$/) {
+                message "[eventMacro] '".$slot."' is not a valid option\n";
```
allanon: Should we allow people to target macros by the name of the running routine? So slot could be `foo` if the running macro's name is `foo`. Targeting by slots could be hard, since slots are allocated automatically.

Nipodemos:
bug report:

sometimes eventMacro is not removing itself from AI queue, making bot simply doing anything. Despite that, all automacros triggers normally.

The only solution i've found is to run `ai clear`, when immediately go back to work properly

plugins/eventMacro/eventMacro.pl
Henrybk: slots are not allocated automatically, they are defined in the automacro or in the plugin command.

Issue #2789, opened 2019-06-09
Add travis-ci.org


sctnightcore:
## ------------------ Openkore Issues Template ------------------

* **Openkore version git**: 

* **Server**: 

* **Bug Report / Feature Request**: Feature Request

* **Summary**: I think we need to add travis-ci for check error in pr 

```
language: perl
perl:
   - "5.12"

before_install:
   - git clone git://github.com/haarg/perl-travis-helper
   - source perl-travis-helper/init
   - build-perl
   - perl -V
   - sudo apt-get install -y libgmp-dev
   - sudo apt-get install -y build-essential g++
   - sudo apt-get install -y libreadline6-dev

install:
   - cpanm --quiet --notest Time::HiRes
   - cpanm --quiet --notest Compress::Zlib

script:
   - perl openkore.pl

```

alisonrag:
GitHub action CI added at:
https://github.com/OpenKore/openkore/pull/3078

PR #2790, opened 2019-06-09
Add Support travis-ci

sctnightcore:
issues #2789 

- [ ] login -> char -> map check

EX..
https://github.com/Openkore-File/openkore/pulls

### DIFF
--- a/src/functions.pl
+++ b/src/functions.pl
@@ -554,7 +554,7 @@ sub finalInitialization {
 	Log::message("\n");
 
 	Log::message("Initialized, use 'connect' to continue\n") if $Settings::no_connect;
-
+	exit if ($Settings::travis_ci);
 	Plugins::callHook('initialized');
 	XSTools::initVersion();
 }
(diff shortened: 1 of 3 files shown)


src/functions.pl
```diff
@@ -428,6 +428,7 @@ sub promptFirstTimeInformation {
 
 sub processServerSettings {
 	my $filename = shift;
+	exit(0) if ($Settings::travis_ci);
```
sctnightcore: it cannot check error in pServer_servertype

sctnightcore: I think we need check with 
login -> char -> map -> wait 5 sec -> exit

Nipodemos:
that's a nice ideia

maybe someday we can use this to test new servertypes that you send :smile:

src/functions.pl
Nipodemos: why wait 5 seconds?

Nipodemos:
i think we should merge this.

sctnightcore:
@Nipodemos  It need wait someone add travis-ci 

i cannot add it ...

koooge:
Hi, CI is a good idea to ensure reliability!
How about [Github Actions](https://github.com/features/actions) instead of 3rd party CI SaaS like travisci?

alisonrag:
we already use github actions

PR #2793, opened 2019-06-10
Add self_interruptible automacro parameter

Henrybk:
This new parameter defines if the macro should be interruptible by the automacro that called it.

exclusive 1: Cannot be interrupted by any automacro, self_interruptible does not matter.

exclusive 0 & self_interruptible 0: Cannot be interrupted by itself, can be interrupted by any other automacro.

exclusive 0 & self_interruptible 1: Can be interrupted by itself and by any other automacro.

Current master default behaviour: exclusive 0 & self_interruptible 1
New default behaviour with these changes: exclusive 0 & self_interruptible 0

### DIFF
--- a/plugins/eventMacro/eventMacro.pl
+++ b/plugins/eventMacro/eventMacro.pl
@@ -589,13 +589,15 @@ sub commandHandler {
 		
 		$eventMacro->{Macro_Runner} = new eventMacro::Runner(
 			$arg,
+			'command',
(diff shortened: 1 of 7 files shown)


ya4ept:
don't forget to add a change to the documentation

Nipodemos:
added [documentation](http://openkore.com/index.php/Self_interruptible) and merging

PR #2785, opened 2019-06-06
Review and cleaning of quest system

Henrybk:
There were some problems with openkore's quest system, mostly related to hunting missions. These changes review all (or most of) the quests packets and adapts them to better reflect the rAthena implementation.

I also updated the plugins that used this information and the console commands.

Tested shortly on RMS renewal test server, seems to be working.

PR #2794, opened 2019-06-10
Fix typo in quest system cleaning

Henrybk:
Fixes a typo in the packet switch caused by #2785


Henrybk:
@Fadreus Would you please test and confirm if this is working?

PR #2804, opened 2019-06-15
fix 'sitAuto_hp_lower' and 'sit' command for Novice 1-3 job lvl

ya4ept:
1) If a Novice (1-3 job lvl) uses the default setting `sitAuto_hp_lower 40` and its HP is reduced to 40%, then the bot just stops and does nothing.
2) If a Novice (1-3 job lvl) uses `sit`command then the bot just stops and does nothing.

I added _NV_BASIC_ check.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2554,7 +2554,7 @@ sub processSitAuto {
 		debug "HP is now > $config{sitAuto_hp_upper}\n", "sitAuto";
 		stand() if (!AI::isIdle && !AI::is(qw(follow sitting clientSuspend)) && !$config{'sitAuto_idle'} && $char->{sitting});
 
-	} elsif (!$ai_v{'sitAuto_forceStop'} && ($weight < 50 || $config{'sitAuto_over_50'}) && AI::action ne "sitAuto") {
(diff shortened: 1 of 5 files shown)


ya4ept:
see please [src\Task\SitStand.pm (line 99)](https://github.com/OpenKore/openkore/blob/master/src/Task/SitStand.pm#L99)
`	} elsif ($self->{actor}->getSkillLevel(new Skill(handle => 'NV_BASIC')) < 3 && ($self->{actor}{jobID} == 0 || $self->{actor}{jobID} == 161)) {  # Check NV_BASIC skill only for Novice and High Novice`

I think this is an extra condition:
`&& ($self->{actor}{jobID} == 0 || $self->{actor}{jobID} == 161)`

Nipodemos:
testing now

ya4ept:
@Nipodemos  how about my previous comment?

Nipodemos:
i don't know what that piece of code does and why it was not working

ya4ept:
it works, but it seems to me that there are too many unnecessary verification conditions

Fadreus:
Doram cant sit on command.

Fadreus:
Doram cant sit on command.
It still checking for NV_BASIC.

PR #2811, opened 2019-06-21
update iRO Renewal files

allanon:
- [ ] QA Review


Nipodemos:
very glad too :smile:

PR #2825, opened 2019-07-12
fix eventMacro 'orphan' parameter

ya4ept:
if i use this automacro, then the OpeKore does not swear on syntax `orphan`:

```
automacro test {
	disabled 0
	JobID 0
	run-once 1
	orphan terminateZZZ
	call {
		log test
	}
}
```

After this patch we get an error:
`[eventMacro] Ignoring automacro 'test' (orphan parameter should be 'terminate', 'terminate_last_call', 'reregister' or 'reregister_safe'. Given value: 'terminateZZZ')`

Maybe we should do it with other options too?


Nipodemos:
interesting how it works even with that ZZZ, it really shoudn't

PR #2850, opened 2019-08-22
Unable to display exp output in guild chat

class777:
i have tried googled many things but could only get to reply 1 item to display
For EG:

automacro test {
pm "item"
exclusive 1
priority 1
timeout 30
call qwer
}

macro qwer {
	$item1 = @invamount (Iron) Iron
	while (@invamount (Iron) > 0) as loop
        do pm "$.lastpm" $item1
        $item1 -= 1;
        end loop
	stop
}

is there a way to display the exp out(exp.txt) in pm or guild chat . 
I had it long back but lost to ransomeware. 
So can anyone help me with the script now tht i am old :D


sctnightcore:
move to issues not Pull request

PR #2878, opened 2019-09-12
[Receive] Quest list update

alisonrag:
added 08FE, 0AFE, 0AFF, 0B0C packets


c4c1n6kr3m1:
afaik, on iRO 08FE only send "hunting mission goal", received after new quest is taken, not mission update.

PR #2887, opened 2019-09-14
update inventoryList to work with new item_list

alisonrag:


ya4ept:
need wiki update (new servers.txt options `itemListType`)

PR #2890, opened 2019-09-15
[xKore 1] workaround to recognize the right char data

alisonrag:
if you use xKore 1 and select chars in highests slots, openkore is not able to define the correct char data.
this pull fixes this

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -6193,6 +6193,16 @@ sub received_character_ID_and_Map {
 		$map_port = $args->{mapPort};
 	}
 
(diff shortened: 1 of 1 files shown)


ya4ept:
Now we have two attempts to change the config: 

confModify

1. in [ClientReceive.pm](https://github.com/OpenKore/openkore/blob/166683911b4dc2608fb14d7d3bb6b3f7c11cbbc5/src/Network/ClientReceive.pm#L49)
2. in [Receive.pm](https://github.com/OpenKore/openkore/blob/166683911b4dc2608fb14d7d3bb6b3f7c11cbbc5/src/Network/Receive.pm#L6212)

PR #2891, opened 2019-09-15
[AI] move ignoreAll to AI

alisonrag:
while using xkore 1,3 sometimes kore send ignoreall too early, so sometimes not work.
so im moving this to AI, to wait some seconds before send the command

Now works in xkore 0,1,2,3

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -182,6 +182,7 @@ sub iterate {
 	processAutoShopOpen();
 	processAutoBuyerShopOpen();
 	processRepairAuto();
+	processSendIgnoreAll();
 	Benchmark::end("AI (part 4)") if DEBUG;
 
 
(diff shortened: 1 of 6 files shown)


src/AI/CoreLogic.pm
```diff
@@ -3449,6 +3450,20 @@ sub processRepairAuto {
...
+	return if($net->getState() != Network::IN_GAME || !$config{ignoreAll} || $ignored_all);
+
+	if(!$timeout{'ai_ignoreAll'}{'time'}) {
+		$timeout{'ai_ignoreAll'}{'time'} = time;
+		return;
+	} elsif(timeOut($timeout{ai_ignoreAll})) {
+		warning "sending ignoreAll... \n";
+		$messageSender->sendIgnoreAll(0);
+		$timeout{'ai_ignoreAll'}{'time'} = time;
+		$ignored_all = 1;
```
Nipodemos: i only ask to move this to above the `$messageSender->sendIgnoreAll(0);` line.
like yesterday PR, if someone use this packet as hook, it will get the old value instead of the updated one

ya4ept:
need wiki update (new timeout `ai_ignoreAll`)
http://openkore.com/index.php/Timeouts.txt

PR #2899, opened 2019-09-17
Fixed 'revive force' command

cydh:


cydh:
crash on `$item->name()`

alisonrag:
as i read when you put item name and dont have the item, openkore try to print/use the item even without have it

PR #2900, opened 2019-09-17
servers.txt update

cydh:
Do not ask more, it's EAC blocked. All you can use is login, see the user online, create a new char :rofl: 
Maybe later will update only the table files which can be extracted from grf and itemInfo.lub


Nipodemos:
what happens if someone actually has the EAC bypass?

cydh:
ask that someone then :thinking:

PR #2903, opened 2019-09-17
[CoreLogic] autostorage wait before action

alisonrag:
sometimes openkore try to get or add item to storage faster then server send the item list to openkore, so sometimes server disconnect us

this pull fixe this


ramochas:
when fix

ramochas:
iro timeoeut map serv

PR #2907, opened 2019-09-23
Fix homunculus map change in RMS server

Henrybk:
As the title states once you change maps in RMS restart the homunculus actor won't be destroyed because a reference to it will be kept in `$char->{homunculus}`, but `$char->{slaves}` is always cleared in map changes, which causes the homunculus to be added again in `AI::SlaveManager` via `AI::SlaveManager::addSlave`.

The problem happens when in `AI::SlaveManager::addSlave` the class of the homunculus is changed from `Actor::Slave::Homunculus` to `AI::Slave::Homunculus`. If the actor is then added again without being deleted first its class will actually be `Actor::Slave::Homunculus`, preventing it from being identified as a homunculus, and being added as a `AI::Slave`.


ya4ept:
need update [wiki](http://openkore.com/index.php/Timeouts.txt) (add new timeouts: ai_mercenary_attack, ai_homunculus_attack_auto, ai_mercenary_attack_auto)

PR #2908, opened 2019-09-23
Fixed homunculus and mercenary timeouts

Henrybk:
At the moment `AI::Slave` uses a single timeout key for all slaves attacks, this makes it so that if you have a homunculus and a mercenary at the same time they won't be able to attack at the same time. This PR fixes that and also adds the missing timeouts.


Henrybk:
Added documentation

PR #2911, opened 2019-09-25
Slave AI cleaning

Henrybk:
This PR:
1 - adds a few timeouts for the slaves.
2 - changes most (I believe all) of the wrong messages in the slave AI referring slaves as homunculus.
3 - adds a new sub for selecting the slave target based on the one used for the player.
4 - changed a few sub names to better reflect what they do.
5 - added a "teleport to master" logic to recall the slave when it gets lost.
6 - Fixes the merc and homun command, which did not work because of wrong class check
7 - Fixes the "move near homun if idle" logic which also did not work because of wrong class check

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -361,6 +361,7 @@ mercenary_attackAuto_notInTown 1
 mercenary_attackAuto_onlyWhenSafe 0
 mercenary_attackDistance 1.5
 mercenary_attackMaxDistance 2.5
+mercenary_attackDistanceAuto 0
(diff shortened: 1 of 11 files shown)


ya4ept:
need wiki update: mercenary_attackDistanceAuto, homunculus_attackDistanceAuto, ai_homunculus_standby, ai_mercenary_standby

Henrybk:
@ya4ept As soon as it gets merged

Henrybk:
Added documentation

PR #2930, opened 2019-10-08
fix wrong portals on iRO

Nipodemos:
as @bookbk said [in here](https://github.com/OpenKore/openkore/commit/5b1af26e78475486a9ff37eec9bba6d5631528cb#commitcomment-35385470) , there is some errors on portals.txt from iRO.

Credit goes to him

### DIFF
--- a/tables/iRO/official/portals.txt
+++ b/tables/iRO/official/portals.txt
@@ -66,14 +66,14 @@ ama_dun01 229 7 ama_in02 119 181 0 c r0 n
 #prontera 282 200 morocc 156 47 1200 c r2 c r3
 #prontera 282 200 gef_fild10 52 326 1700 c r2 c r4
 #prontera 282 200 alberta 117 56 1800 c r2 c r5
-prontera 35 208 izlude 94 103 2000 1 c r2 c r0
-prontera 35 208 geffen 120 39 2000 1 c r2 c r1
-prontera 35 208 payon 161 58 2000 1 c r2 c r2
-prontera 35 208 morocc 156 47 2000 1 c r2 c r3
-prontera 35 208 alberta 117 56 2000 1 c r2 c r4
-prontera 35 208 aldebaran 168 112 2000 1 c r2 c r5
-prontera 35 208 comodo 209 143 3000 1 c r2 c r6
-prontera 35 208 umbala 100 154 3000 1 c r2 c r7
+#prontera 35 208 izlude 94 103 2000 1 c r2 c r0
+#prontera 35 208 geffen 120 39 2000 1 c r2 c r1
+#prontera 35 208 payon 161 58 2000 1 c r2 c r2
+#prontera 35 208 morocc 156 47 2000 1 c r2 c r3
+#prontera 35 208 alberta 117 56 2000 1 c r2 c r4
+#prontera 35 208 aldebaran 168 112 2000 1 c r2 c r5
+#prontera 35 208 comodo 209 143 3000 1 c r2 c r6
+#prontera 35 208 umbala 100 154 3000 1 c r2 c r7
 #payon 175 226 prontera 116 72 1200 c r2 c r0
 #payon 175 226 alberta 117 56 1200 c r2 c r1
 #payon 175 226 morocc 156 47 1200 c r2 c r2
--- a/tables/iRO/official/portals.txt
+++ b/tables/iRO/official/portals.txt
@@ -8,7 +8,7 @@ lou_fild01 190 100 alberta 235 45 0 c r0
 tur_dun01 165 29 alberta 241 115 0 c r0 c c
 alberta 245 93 amatsu 197 83 10000 c r1 c c c r0 n
 amatsu 194 79 alberta 243 91 0 c r0 n
-alberta 245 69 gon_fild01 258 82 0 c r1 c c r0 n
+alberta 245 62 gon_fild01 258 82 0 c r1 c c r0 n
 gon_fild01 255 79 alberta 243 67 0 c r0 n
 izlude 201 181 izlu2dun 107 50 150 c r0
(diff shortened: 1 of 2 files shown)


Fadreus:
> you can go from prontera to yuno in iRO? oh what a luxury
> Edit: no you can't, i just checked it in Chaos
> 
> i'll make those fixes, thank you for your contribution

@Nipodemos 
Yes you can but not kafra.
See the coordinate; prontera 98 120
image

Same NPC also in Yuno, to go Prontera.
image

Btw, how about some RO using old Prontera?
Shouldn't those Kafra being added as optional rather than replacing them?
Just like VIP kafra :thinking:

Fadreus:
> you can go from prontera to yuno in iRO? oh what a luxury
> Edit: no you can't, i just checked it in Chaos
> 
> i'll make those fixes, thank you for your contribution

@Nipodemos 
Yes you can but not kafra.
See the coordinate; prontera 98 120
image

Same NPC also in Yuno, to go Prontera.
Edit: This NPC provide more option when you are VIP. Non VIP just Prontera-Yuno. Exist in Eden too.
image

Btw, how about some RO using old Prontera?
Shouldn't those Kafra being added as optional rather than replacing them?
Just like VIP kafra :thinking:

tables/iRO/official/portals.txt
```diff
@@ -66,12 +66,14 @@ ama_dun01 229 7 ama_in02 119 181 0 c r0 n
...
-#prontera 29 207 gef_fild10 52 326 1700 c r2 c r4
-#prontera 29 207 alberta 117 56 1800 c r2 c r5
+prontera 35 208 izlude 94 103 2000 1 c r2 c r0
+prontera 35 208 geffen 120 39 2000 1 c r2 c r1
+prontera 35 208 payon 161 58 2000 1 c r2 c r2
+prontera 35 208 morocc 156 47 2000 1 c r2 c r3
+prontera 35 208 alberta 117 56 2000 1 c r2 c r4
+prontera 35 208 aldebaran 168 112 2000 1 c r2 c r5
+prontera 35 208 comodo 209 143 3000 1 c r2 c r6
+prontera 35 208 umbala 100 154 3000 1 c r2 c r7
```
Fadreus: These are under VIP Warps.
Should replace under:
https://github.com/OpenKore/openkore/blob/0842ac7990a41aecc7b91130de33afea7ee59d0c/tables/iRO/official/portals.txt#L95
There will be a conflict later.

And if wanna change, should change VIP too.
But I prefer adding it as Optional rather than default.

Nipodemos: roger that sir

i wonder why the vip warps of this specific npc are exactly equal to the non-VIP warps

i'm adressing these issues now

Nipodemos:
ok @Fadreus 
* That npc you mentioned that teleports you to yuno, i've noticed that it is already there in portals.txt, so no changes were necessary
* i've seen that you said that it uses the default tables/portals.txt, so i replicated all changes to that file.
* just in case someday iRO uses the correct portals.txt, i've maintened changes there too

* moved fix to the non-vip section, and also altered the VIP section, while maintaining optional as it already is

i think i did everything

tables/portals.txt
```diff
@@ -66,12 +66,14 @@ ama_dun01 229 7 ama_in02 119 181 0 c r0 n
...
-#prontera 29 207 gef_fild10 52 326 1700 c r2 c r4
-#prontera 29 207 alberta 117 56 1800 c r2 c r5
+#prontera 35 208 izlude 94 103 2000 1 c r2 c r0
+#prontera 35 208 geffen 120 39 2000 1 c r2 c r1
+#prontera 35 208 payon 161 58 2000 1 c r2 c r2
+#prontera 35 208 morocc 156 47 2000 1 c r2 c r3
+#prontera 35 208 alberta 117 56 2000 1 c r2 c r4
+#prontera 35 208 aldebaran 168 112 2000 1 c r2 c r5
+#prontera 35 208 comodo 209 143 3000 1 c r2 c r6
+#prontera 35 208 umbala 100 154 3000 1 c r2 c r7
```
Fadreus: Supposedly to just change the coordinate, not the fee nor the destination.
Orc Village is missing or I'm not sure if it still there.
I don't have VIP right now to test.

Fadreus:
_-Just a minor stuff-_

Those fix are just for West Gate Kafra. Fixed as addressed.
However, South & East are incorrect too _(If you wanna do it)_

- `prontera 151 29` -> `prontera 151 34` (South)
- `prontera 282 200` -> `prontera 276 200` (East)

The Optional I meant is to add rather than replacing old Kafra coordinates.
So people can choose which to enable, old or new; depend on servers they are playing.
I haven't play Classic, so IDK about which Prontera is there.
Since this change is made into `tables/portals.txt`, this gonna affect servers using old Prontera.
So just hope no one gonna say "Unable to use Warp after latest download."

tables/iRO/official/portals.txt
Fadreus: VIP is IRO is like "Unlock" feature from normal stuff.
If you have VIP, Kafra just offer 2nd storage, more destination, less fee and etc etc.

Nipodemos:
unfortunaly i will be very busy these days so i can't insert that fix
the ideal was to iro to use a portals.txt to classic and other to renewal, and stop using default one, but i don't know how to fix that

tables/portals.txt
Nipodemos: i'm also don't have vip to confirm, unfortunaly.
Does kafra teleports change when you are vip? here in bRO we have a different NPC for VIP teleports

Fadreus: The destination option didn't change, except the list's order slightly changed as new destination is added when you are VIP _(Orc village, Mine Pits, Comodo Beach, etc etc)_

PR #2938, opened 2019-10-11
small fix avoid hash

ya4ept:
every time the `avoidList_talk` function is run, the name of the new player is added to the% avoid hash
Example:
```perl
#!/usr/bin/env perl
%avoid = ();
$avoid{Players}{"[GM]Example"}{disconnect_on_chat} = 0;
$avoid{ID}{"000001"}{disconnect_on_chat} = 3;
use Data::Dumper; print Dumper(%avoid);

print "====================\n";
$user = "zzzzzzzzzzzzzzzzzzzzzzzzzz";
if ($avoid{Players}{lc($user)}{disconnect_on_chat}) {
    print "if true $user\n";
}

use Data::Dumper; print Dumper(%avoid);
```
avoid
this PR corrects this behavior

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -3933,12 +3933,11 @@ sub avoidGM_near {
 		# skip this person if we dont know the name
 		next if (!defined $player->{name});
 
-		# Check whether this "GM" is on the ignore list
-		# in order to prevent false matches
-		last if (existsInList($config{avoidGM_ignoreList}, $player->{name}));
+		# Check whether this "GM" is on the ignore list in order to prevent false matches
+		next if (existsInList($config{avoidGM_ignoreList}, $player->{name}));
 
 		# check if this name matches the GM filter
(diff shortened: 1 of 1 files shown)


src/Misc.pm
```diff
@@ -3988,6 +3990,16 @@ sub avoidList_near {
 	for my $player (@$playersList) {
 		my $avoidPlayer = $avoid{Players}{lc($player->{name})};
 		my $avoidID = $avoid{ID}{$player->{nameID}};
+
+		# Exit if the player is not on the avoid list
+		last if (!$avoidPlayer and !$avoidID);
```
lututui: This is wrong. The first player not on `playersList` breaks the avoid function for all other users. This is supposed to be `next`, not `last`

ya4ept: yes, you are absolutely right!

Mortimal:
seems good to me too

PR #2964, opened 2019-11-01
change wx npc portal dot color

c4c1n6kr3m1:
just a test
@Nipodemos ask this on https://github.com/OpenKore/openkore/pull/2429  , so why not
don't merge until code review by an elder


c4c1n6kr3m1:
did u guys read this
```
zeny => $args[7],
```
zeny on 7th element , really ??? do we need this , i don't see any code relate to this
looks like wasting a hash key for me

Nipodemos:
> did u guys read this
> 
> ```
> zeny => $args[7],
> ```
> 
> zeny on 7th element , really ??? do we need this , i don't see any code relate to this
> looks like wasting a hash key for me

it seems like who CTRL+C CTRL+V the sub didn't bother to remove
but why the sub is there? why it is not using the normal one in FileParser.pm?

PR #2965, opened 2019-11-01
change wx npc portal dot color

c4c1n6kr3m1:
@Nipodemos ask this on https://github.com/OpenKore/openkore/pull/2429
don't merge until code review by an elder
i don't know what color to use , but i put cyan

PR #2966, opened 2019-11-03
Wx npc portal

ya4ept:
this patch fixes: #2429 and #2965:
- now function `parsePortals` uses global hash `%portals_lut`
- Warp NPCs are now cyan

Thanks @c4c1n6kr3m1 

fix

### DIFF
--- a/src/Interface/Wx/MapViewer.pm
+++ b/src/Interface/Wx/MapViewer.pm
@@ -55,6 +55,7 @@ sub new {
 	$self->{textColor}{npc}     = new Wx::Colour (127, 0, 127);
 	$self->{brush}{portal}      = new Wx::Brush(new Wx::Colour(255, 128, 64), wxSOLID);
 	$self->{textColor}{portal}  = new Wx::Colour (191, 95, 47);
+	$self->{brush}{portalNpc}   = new Wx::Brush(new Wx::Colour(0, 255, 255), wxSOLID);
 	$self->{brush}{slave}       = new Wx::Brush(new Wx::Colour(0, 0, 127), wxSOLID);
 	
 	$self->{brush}{gaugeBg}     = new Wx::Brush(new Wx::Colour(63, 63, 63), wxSOLID);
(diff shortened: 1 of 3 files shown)


Nipodemos:
oh wow i like like
i'm gonna test it now!

tables/portals.txt
```diff
@@ -574,20 +574,6 @@ bat_room 169 211 bat_room 154 150
...
-moc_fild01 56 384 prt_fild08a 55 21
-prontera 156 22 prt_fild08a 170 378
-prt_fild07 383 239 prt_fild08a 16 239
-prt_fild07 385 186 prt_fild08a 16 187
-prt_fild08a 16 187 prt_fild07 385 186
-prt_fild08a 16 239 prt_fild07 383 239
-prt_fild08a 55 21 moc_fild01 56 384
-prt_fild08a 170 378 prontera 156 22
-prt_fild08a 233 16 moc_fild01 239 382
-prt_fild08a 371 212 izlude 24 98
```
Nipodemos: why you removed those lines?

Nipodemos:
if i understand correctly, this needs to be merged AFTER the change dot color PR right?

alisonrag:
yep orange for portals, blue for npcs

src/Interface/Wx/MapViewer.pm
```diff
@@ -310,30 +311,33 @@ sub setMapDir {
...
+					destination => {
+						field => $portals_lut{$portal}{dest}{$dest}{map},
+						x => $portals_lut{$portal}{dest}{$dest}{x},
+						y => $portals_lut{$portal}{dest}{$dest}{y},
+					},
+				});
+			}
 		}
 	}
-	close FILE;
```
alisonrag: will openkore keep the file open?

Nipodemos: the is no file being opened anymore

tables/portals.txt
ya4ept: 1. because these portals seem to be used only on iRO server
2. because of this, portals are duplicated, for example:
```
prontera 156 22 prt_fild08 170 378
prontera 156 22 prt_fild08a 170 378
```

ya4ept:
> if i understand correctly, this needs to be merged AFTER the change dot color PR right?
no, this patch includes a color change

src/Interface/Wx/MapViewer.pm
ya4ept: yes, now we do not reopen the file. We use a ready-made global hash `%portals_lut`

ya4ept:
@Nipodemos, @alisonrag  can we now merge?

Nipodemos:
to me yes

PR #2978, opened 2019-11-20
[CoreLogic.pm - processPartySkillUse] Timeout absence fix.

Mortimal:
Fixed timeout problem.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2692,6 +2692,7 @@ sub processPartySkillUse {
 		for (my $i = 0; exists $config{"partySkill_$i"}; $i++) {
 			next if (!$config{"partySkill_$i"});
 			next unless checkSelfCondition("partySkill_$i");
+			$ai_v{"partySkill_$i"."_time"} = time;
 			$party_skill{skillObject} = Skill->new(auto => $config{"partySkill_$i"});
 			$party_skill{owner} = $party_skill{skillObject}->getOwner;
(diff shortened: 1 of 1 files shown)


Mortimal:
> 
> 
> I think this is not the best solution. I have a 1 line solution.

show me

src/AI/CoreLogic.pm
```diff
@@ -2692,6 +2692,7 @@ sub processPartySkillUse {
 		for (my $i = 0; exists $config{"partySkill_$i"}; $i++) {
 			next if (!$config{"partySkill_$i"});
 			next unless checkSelfCondition("partySkill_$i");
+			$ai_v{"partySkill_$i"."_time"} = time;
```
c4c1n6kr3m1: is this the right  place? what will happen  if succes through checkSelfCondition but fail on checkPlayerCondition ?

src/AI/CoreLogic.pm
```diff
@@ -2802,6 +2804,7 @@ sub processPartySkillUse {
 				$party_skill{isSelfSkill} ? $party_skill{owner} : $party_skill{targetActor},
 				$party_skill{prefix},
 			);
+			$timeout{"partySkill_".$party_skill{skillTimeoutId}."_timeout"}{time} = time;
```
c4c1n6kr3m1: sorry,but why do we need this?

ya4ept:
@Mortimal see #2982

PR #2979, opened 2019-11-20
[Misc.pm - useTeleport] Inventory state cheking.

Mortimal:
I think this will fix most problems with teleporting with wings/items when inventory is still not loaded.

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -3495,7 +3495,13 @@ sub useTeleport {
 			return 1;
 		}
 	}
-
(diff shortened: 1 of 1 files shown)


Nipodemos:
hello mortimal, thanks for this pull request
i want to ask, how can i test this?

i'm asking because i don't know what error you are fixing.
I can happily test this as soon as you answer

Mortimal:
When you change map and Openkore need to teleport immediately it will spam no items in inventory. This is to fix that.

PR #2994, opened 2019-12-06
update Readme.md

alisonrag:


Fadreus:
I'm not aware of those servers but their websites and game still exist and running.
Usually it was removed because server closed; for this case is it because other *openkore* is supporting them?

alisonrag:
The only server that i removed was mRO because was merged into phRO

Other servers i only re ordered

I still hope someday someone will cane here with a solution against the shielda

Fadreus:
> Other servers i only re ordered

Uh, mmm.. My bad.. Some rotten eyes I have. :expressionless:
I only checked RO from Extreme Electronics company website.
<details>
<summary>I think I saw some servers with their *openkore* missing there just now. /swt</summary>
Btw, it would be nice to listing RO based to their region. Something like.

iRO
bRO

euRO
rRO

kRO
jRO
cRO
twRO

tRO
phRO
idRO

Western continent > Middle continent > Eastern continent.
In that order or something.
</details>

PR #3027, opened 2020-01-31
[tRO] support

alisonrag:
still need bypass e a c


sylvester2677:
wow untill this time no one can bypass EAC    T_T

Fadreus:
@sylvester2677 'no one can bypass' is misleading.
'No one with zero knowledge can bypass' is more accurate.
People with appropriate skills and knowledge can by bypass.

PR #3032, opened 2020-02-03
[Parser] Fix missing unpack string into buying_store_items_list

alisonrag:
fix missing unpackstring into buying_store_items_list caused in: #3026


oleokoong:
this will fix Not trigger buyshop call hook?

alisonrag:
Nope, is #3016

PR #3057, opened 2020-02-23
Add euRo server info

koooge:
Hi. just add euRO server info(not working yet).


ya4ept:
why do we need this commit?

PR #3064, opened 2020-02-24
fix: Parse item including '#'

koooge:
Hi.
FileParses failed to parse name in items_control.txt like "Unknown #502" because the "#" was regarded comment

### DIFF
--- a/src/FileParsers.pm
+++ b/src/FileParsers.pm
@@ -169,7 +169,7 @@ sub parseCommandsDescription {
 
 	undef %{$r_hash} unless $no_undef;
 	my ($key, $commentBlock, $description);
-	
+
 	my $reader = new Utils::TextReader($file);
 	while (!$reader->eof()) {
 		my $line = $reader->readLine();
@@ -202,11 +202,11 @@ sub parseCommandsDescription {
 
 		} elsif ($line =~ /^(.*?)\t+(.*)$/) {
 			push @{$r_hash->{$key}}, [$1, $2];
-		
+
 		} elsif ($line !~ /\t/) {	#if a console command is without any params
 			push @{$r_hash->{$key}}, ["", $line];
 		}
-		
+
 	}
 	return 1;
 }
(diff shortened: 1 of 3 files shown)


Nipodemos:
ok, really it is a problem using # for comments
but in your test, you still put a comment in the end of the line.
Just to be sure, with this PR merged it is not possible anymore to put comments in front of a line that has content, right? 
then i think you should remove this on the test file

koooge:
It's ok if the line begins with '#' by this line `next if $line =~ /^\s*#/;`.
https://github.com/OpenKore/openkore/pull/3064/files#diff-deeee5b3577ccb4f13fcc9088e4cb01bL514


FYI) OK pattern
```
# comment
   # comment
"something" 1 2 3 4 5 # comment 
"Unknown #502" 1 2 3 4 5
```

NG pattern
```
Unknown #502 1 2 3 4 5
"something" 1 2 # coment 
```

koooge:
These file parsers are not simple and not easy-understand. So, structured data such as json or yaml would be better and should use existing CPAN lib though.

alisonrag:
Sometime in future we need to move table data to yaml or json

ya4ept:
It seems to me that this is a bad name for the item. Have you met him anywhere?
```"Unknown #502" 1 2 3 4 5```

my items_control.txt:
```all 0 1 0 1 1 # test
505 0 1 0 # Blue Potion
506 0 1 0 1 1
```
before this patsh:
изображение

after this patch:
изображение

alisonrag:
> It seems to me that this is a bad name for the item. Have you met him anywhere?
> `"Unknown #502" 1 2 3 4 5`
> 
> my items_control.txt:
> 
> ```
> 505 0 1 0 # Blue Potion
> 506 0 1 0 1 1
> ```
> 
> before this patsh:
> изображение
> 
> after this patch:
> изображение

People still using this kind of name instead of update their items.txt so this is true, i found some cases in the last year

ya4ept:
> People still using this kind of name instead of update their items.txt so this is true, i found some cases in the last year
it seems to me that we should teach users to update this file correctly

alisonrag:
> > People still using this kind of name instead of update their items.txt so this is true, i found some cases in the last year
> > it seems to me that we should teach users to update this file correctly

they dont know / want to learn get it from grf, and convert it to UTF-8, just imagine if they will install lua, then download script, the unpack the .lub from system folder, then use the script annd convert the file to UTF-8

PR #3068, opened 2020-02-27
[Feature] Attendance System

alisonrag:
Added support to Attendance System (daily login)

New commands
`attendance<open|request>`

image


alisonrag:
when system is unavaliable:

image

PR #3078, opened 2020-02-29
feat: Add github actions

koooge:
Hi there, this is the proposal OpenKore's CI using GitHub Actions. yet another #2790 but GitHub Actions's auth would be easier since it's all in GitHub.
For now the CI make sure to pass  just `make test`.

samples
ok: https://github.com/koooge/openkore/pull/1
ng: https://github.com/koooge/openkore/pull/2

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+    tags-ignore:
+      - '*'
+  pull_request:
+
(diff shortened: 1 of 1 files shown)


alisonrag:
@sctnightcore @allanon @itsrachelfish

Nipodemos:
i am absolutely in favor of this.
If this works only by merging, then it is already good

sctnightcore:
cool ! 
it' can check src .

PR #3112, opened 2020-04-13
Fix homunculus not getting stats after map change

Henrybk:
euRO seems to have stopped sending 'homunculus_property' after a map change, this makes it so now  openkore has to keep the information itself. This change makes it so homunculus_info won't delete the homunculus object after a map change if it still holds the same ID.

This introduces another problem, AI::SlaveManager::addSlave requires the actor type to be Actor::Slave::Homunculus, but after a map change it will have been changed to AI::Slave::Homunculus, so I also added a check to change it back to Actor::Slave::Homunculus.

Previous fix: #2907

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -2765,10 +2765,19 @@ sub homunculus_info {
 	if ($args->{state} == HO_PRE_INIT) {
 		my $state = $char->{homunculus}{state}
 			if ($char->{homunculus} && $char->{homunculus}{ID} && $char->{homunculus}{ID} ne $args->{ID});
-		$char->{homunculus} = Actor::get($args->{ID});
+		
+		# Some servers won't send 'homunculus_property' after a teleport, so we don't delete $char->{homunculus} object
+		if ($char->{homunculus}{ID} ne $args->{ID}) {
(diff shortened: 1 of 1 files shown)


src/Network/Receive.pm
```diff
@@ -2765,10 +2765,19 @@ sub homunculus_info {
 	if ($args->{state} == HO_PRE_INIT) {
 		my $state = $char->{homunculus}{state}
 			if ($char->{homunculus} && $char->{homunculus}{ID} && $char->{homunculus}{ID} ne $args->{ID});
-		$char->{homunculus} = Actor::get($args->{ID});
+		
+		# Some servers won't send 'homunculus_property' after a teleport, so we don't delete $char->{homunculus} object
+		if ($char->{homunculus}{ID} ne $args->{ID}) {
```
ya4ept, line 2770: I propose to replace
```perl
$char->{homunculus} = Actor::get($args->{ID}) if ($char->{homunculus}{ID} ne $args->{ID});

ya4ept:
Tested on eoRO - everything works well

Henrybk:
@ya4ept changed

PR #3072, opened 2020-02-27
[FIX] Change rodex body print

alisonrag:
image

this pull fixes #2927

PR #3130, opened 2020-05-15
[Fix] Rodex Content, Target Job and Type

alisonrag:
fixed no transition to the next line after the body of the letter and not showing Item count and Zeny, as reported here:
https://github.com/OpenKore/openkore/pull/3072#issuecomment-626248702

added content type as suggested here:
https://github.com/OpenKore/openkore/pull/3072#issuecomment-626323067

also updated openkore to show job name instead of job id after receive target info in rodex.

image

thanks @ya4ept for the report and @c4c1n6kr3m1 for the suggestion

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -8327,7 +8327,7 @@ sub rodex_check_player {
 
 	my $print_msg = center(" " . "Rodex Mail Target" . " ", 79, '-') . "\n";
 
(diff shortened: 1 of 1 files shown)


ya4ept:
we need to increase the line width (-----) to 119 characters. It doesn’t look so beautiful:
```
------------------------------ Rodex Mail Page  -------------------------------
0    From:  stew4                     Read:   Yes  ID:  530726    Title:  test thema                Content: Text
1    From:  stew4                     Read:   Yes  ID:  530657    Title:  Title                     Content: Text
2    From:  stew4                     Read:   Yes  ID:  530641    Title:  Title                     Content: Text
-------------------------------------------------------------------------------
```

PR #3145, opened 2020-05-29
[FIX] check for monster before try to back to route

alisonrag:
add checkmonster to ai queue give attackauto time to find other monster beforeback to route again, this avoid openkore to lost main route and walk away from monsters then back to kill

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -174,6 +174,7 @@ sub iterate {
 	Misc::checkValidity("AI part 3");
 	processAutoEquip();
 	processAutoAttack();
+	processCheckMonster();
 	processItemsTake();
 	processItemsAutoGather();
 	processItemsGather();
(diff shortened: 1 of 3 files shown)


hubertyang88:
Nice!! Tested for 30 minutes , no funny movement before or after attck both in random or planLockMapWalk mode. 
Now, random walk can keep the previous destination after attck instead of choosing a new one ,  Good!

src/AI/CoreLogic.pm
```diff
@@ -174,6 +174,7 @@ sub iterate {
 	Misc::checkValidity("AI part 3");
 	processAutoEquip();
 	processAutoAttack();
+	processCheckMonster();
```
c4c1n6kr3m1, line 177: i think it's better on MANUAL AI  ???

src/AI/CoreLogic.pm
```diff
@@ -3584,4 +3583,17 @@ sub processPartyShareAuto {
 	}	
 }
 
+sub processCheckMonster {
+	return if AI::inQueue("attack");
+	return if !AI::inQueue("checkMonsters");
```
c4c1n6kr3m1, line 3588: no need for this because next line is AI::is ???

onstop:
capture-20200531-032247
3516  + if (!$field->isCity && !AI::inQueue("storageAuto", "buyAuto") && $config{dcOnPlayer}

alisonrag:
@onstop

hubertyang88:
about timeout , Should there be no "if" , or it will only time out for the first time?

$timeout{'ai_check_monster_auto'}{'time'} = time;

PR #3186, opened 2020-06-18
fix 09DB, 09DC and 09DD packets

ya4ept:
removed "body_style" from packages 09DB (actor_moved), 09DC (actor_connected) and 09DD (actor_exists)
NPC
09DD dont have body_state
is used in kRO from 20131223 to 20150513
body_state implemented at >=20150513


ioaltisgithub:
21/6/2020: when I logged in (tRo), the character couldn't move anywhere and disconnected from MapServer, well, how to deal with this? Thanks you

PR #3209, opened 2020-06-28
Add plugin hook for packet_msgstring

sctnightcore:


src/Network/Receive.pm
```diff
@@ -8889,11 +8889,17 @@ sub skill_msg {
...
+	if ($msgTable[$index]) { # show message from msgstringtable.txt
+		message "$msgTable[$index]. Value: $args->{paral}\n", "info";
 	} else {
-		warning TF("Unknown msgid:%d paral:%d. Need to update the file msgstringtable.txt (from data.grf)\n", $args->{index}, $args->{paral});
+		warning TF("Unknown msgid:%d paral:%d. Need to update the file msgstringtable.txt (from data.grf)\n", $index, $args->{paral});
 	}
+
+	Plugins::callHook('packet_msgstring', {
+		index => $index,
+		val => $args->{para1}
```
alisonrag, line 8901: paraL, not para1

src/Network/Receive.pm
```diff
@@ -8889,11 +8889,17 @@ sub skill_msg {
 sub msg_string {
 	my ($self, $args) = @_;
 
-	if ($msgTable[++$args->{index}]) { # show message from msgstringtable.txt
-		message "$msgTable[$args->{index}]. Value: $args->{paral}\n", "info";
+	my $index = ++$args->{index};
+	if ($msgTable[$index]) { # show message from msgstringtable.txt
+		message "$msgTable[$index]. Value: $args->{paral}\n", "info";
```
Nipodemos, line 8894: i've noticed that before, it had the `++` only in one place. Both in warning and message it was displayed the index without the `++`. But now every part has the plus sign. Can you confirm if the message remains as intended?

src/Network/Receive.pm
ya4ept: https://github.com/OpenKore/openkore/blob/master/src/Network/Receive/ServerType0.pm#L479
we use `para1`

alisonrag: so need to fix it in #8896

ya4ept: what is "paral"?

ya4ept:
how can we generate this event?

src/Network/Receive.pm
Nipodemos: that #8896 doesn't exist. Is it a typo @alisonrag ?

alisonrag:
> how can we generate this event?

the only that i can record is use teleport while teleport screen is opened or try to use skill while talking NPC, both should generate the following message:
```
"any work in progress (NPC dialog, manufacturing...)quit and try again.#"
```

src/Network/Receive.pm
alisonrag: line 8896* ^^
ya4ept already fixd this

src/Network/Receive.pm
alisonrag: at this point $index already have the correct value

ya4ept:
we have two similar functions [msg_string](https://github.com/OpenKore/openkore/blob/master/src/Network/Receive.pm#L8889):
```perl
sub msg_string {
	my ($self, $args) = @_;

	my $index = ++$args->{index};
	if ($msgTable[$index]) { # show message from msgstringtable.txt
		message "$msgTable[$index]. Value: $args->{paral}\n", "info";
	} else {
		warning TF("Unknown msgid:%d paral:%d. Need to update the file msgstringtable.txt (from data.grf)\n", $index, $args->{paral});
	}

	Plugins::callHook('packet_msgstring', {
		index => $index,
		val => $args->{paral}
	});
}
```
and [message_string](https://github.com/OpenKore/openkore/blob/master/src/Network/Receive.pm#L10458):
```perl
# not only for mercenaries, this is an all purpose packet !
sub message_string {
	my ($self, $args) = @_;

	if ($msgTable[++$args->{msg_id}]) { # show message from msgstringtable.txt
		warning "$msgTable[$args->{msg_id}]\n";
		$self->mercenary_off() if ($args->{msg_id} >= 1267 && $args->{msg_id} <= 1270);
	} else {
		warning TF("Unknown message_string: %s. Need to update the file msgstringtable.txt (from data.grf)\n", $args->{msg_id});
	}
}
```
 We need to leave only one

Issue #3198, opened 2020-06-22
Exp doubled after relog


hubertyang88:
## ------------------ Openkore Issues Template ------------------

* **Openkore version git**: 

* **Server**: 
cRO

* **Bug Report / Feature Request**:
Bug Report
After relog , bot will get a very high exp , I don't know if it's send by server or a bug of openkore .
after relog bot shows:
[Jun 22 22:51:43 2020.66] You have gained 38930192/-74401562 (51.68%/-56.51%) Exp
this is the total exp bot get before relog, so after relog exp is wrong.

* **Summary**:

c4c1n6kr3m1:
always broken or recently update? (i don't play official coz i can't penetrate the security heh, not programer here)

hubertyang88:
It is recently update , I don't update frequently , but I update openkore in April for cRO update Dorum, before this unpate , exp works fine.

c4c1n6kr3m1:
than test with your old openkore first, because it could be server's update

c4c1n6kr3m1:
may be relevant, months ago i got 64 bits exp problem https://github.com/OpenKore/openkore/issues/3017 than another problem is exp counting. Latest private server i tested, claimed latest version, using different mechanism
old server send level first than exp ( may be reverse, i don't remember well )
but this server send the opposite, that's how i got another exp problem  ( or reverse, as i said i don't remember well )

but it's depend on server
if any one want to take a look .... not me

Issue #2788, opened 2019-06-08
zeny amount not updating after relog/charselect


xDuds:
## ------------------ Openkore Issues Template ------------------

* **Openkore version git**: 

* **Server**: 

* **Bug Report / Feature Request**:

* **Summary**:

ServeR: bRO
version versionctime:2019_05_31
bug report
The zeny amount is not updating if you spend the zeny on another openkore console or on the official ragnarok client. After relogin or charselect, the amount still shows as if you never spent, unless you do trade with a player or buy/sell with a npc.

Fadreus:
Just restart ur bot.

Fadreus:
Related to this issue
-Base level is not updated if level up with different client.
-EXP is not updated if using exp gained using different client.
*(There are more but I couldn't care less or even remembering it).*

Anyways, I noticed this early 2019.
I thought this was intended but seeing this issue is still open, so I  add few more that related to  this issue.

It appear to me as openkore only update once; when client started or when there are changes rather than updating character data every time logged in. I assume it to reduce the load on openkore/server?

몰라
![FB_IMG_15625044282660696](https://user-images.githubusercontent.com/32188494/61588427-ca2b6e00-abcd-11e9-99c6-8b7d81707450.jpg)

moneyisgod:
Maybe something about not design pattern on 'player' object.

PR #3214, opened 2020-07-01
[fix] update character data

alisonrag:
this pull fixes: #3198
this pull fixes: #2788

* update the received_characters function to save correct character information
* fix exp function 

EXP in character selection screen:
client:
exp client

openkore:
exp at char server

exp gain:
exp at 175

------

zeny before relog:
zeny before relog

zeny lost in client:
zeny client

zeny after relog:
zeny after relog


alisonrag:
i checked exp from my 170+ characters in official servers.
1 - seems like exp and exp_job is set to 0 after gain a level
2 - even after 175 there is no monster or exp that can overflow the exp or exp_job

hubertyang88:
#3198 fixed! thx!

PR #3219, opened 2020-07-04
[fix] follow up 3215

alisonrag:


alisonrag:
don't approve yet, seems like pincode is also triggering 0066 send

i will check the src

alisonrag:
tested at bRO, iRO RT/Renewal, tRO, localserver rAthena

would be good if someone can test in ruRO / euRO or other server

PR #3224, opened 2020-07-07
[fix] solveMessage itemLink

alisonrag:
image


ya4ept:
@alisonrag @sctnightcore 
why can't we use the same condition for both options?

alisonrag:
there are 2 or 3  versions of itemlink.
if u can add a regex to work at same time to all, fell free to create a pull request.
this solution worked to the current issue, should has a better option

ya4ept:
as far as I can see - the second expression includes the first template
I see two recording formats:
1. one item:
```lua
Sell<ITEML>0000y1kk&0g)00)00)00)00+05,00-01+0B,00-0m</ITEML>500k
```
2. multiple items:
```lua
S><ITEML>0000y1nC&05)00)00+0h,00-0C+0F,00-0h</ITEML><ITEML>000021jM&01)00)00+0h,00-0N+0B,00-0g</ITEML><ITEML>000021jM&01)00)00+0f,00-02+0B,00-0e</ITEML>
```
i think we can use one expression:
```lua
<ITEML>([a-zA-Z0-9%&(),+\-*]*)<\/ITEML>
```
see: https://regex101.com/r/0lel32/1

PR #3242, opened 2020-07-26
Fix headgears

ya4ept:
1. we must not skip empty lines
2. added messages when a nearby player changes headgear (low, top, mid)
3. added a lot of new headgears


ucodelukas:
here the headgear file was corrupted

ya4ept:
@ucodelukas already fixed: #3276

PR #3246, opened 2020-07-29
fixed the unshift function

ya4ept:
```
 macro unshift {
     &unshift(@monsters, Porint)
     &unshift(@monsters, Volf)
     &unshift(@monsters, Maya)
     log The array \@monsters has size @monsters and its members are $monsters[0], $monsters[1] and $monsters[2]
 }
```
if the array is empty then OpenKore gets a fatal error:
image

this patch fixes a bug:
image

### DIFF
--- a/plugins/eventMacro/eventMacro/Core.pm
+++ b/plugins/eventMacro/eventMacro/Core.pm
@@ -971,7 +971,7 @@ sub unshift_array {
 		undef $new_member;
 	}
 
-	my @old_array = @{$self->{Array_Variable_List_Hash}{$variable_name}};
+	my @old_array = @{$self->{Array_Variable_List_Hash}{$variable_name}} if $self->{Array_Variable_List_Hash}{$variable_name};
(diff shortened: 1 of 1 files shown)


ya4ept:
yes, I have tested other functions. The problem only occurs with this function

> btw, nice idea of make a em command to eventMacro, i wish i had thought that back in the days

use config: [alias_em eventMacro](https://openkore.com/wiki/Alias)

alisonrag:
> yes, I have tested other functions. The problem only occurs with this function
> 
> > btw, nice idea of make a em command to eventMacro, i wish i had thought that back in the days
> 
> use config: [alias_em eventMacro](https://openkore.com/wiki/Alias)

years using and developing openkore and i never see this haahah

ya4ept:
maybe we can fix the original eventMacro command?

Nipodemos:
both `em` and `eventMacro` commands should be official

PR #3251, opened 2020-08-01
update map of tro

comdevx:
add fields missing.
update maps had multi channel.


ya4ept:
All our maps are packed in *.GZ archive.
I see a lot of duplicate maps ..

Nipodemos:
there is only one problem with that update (and it is not your fault)

those map files are shared for every server that openkore opens, so the maps that are different in other servers will make openkore confused

To be honest that was already happening for years, but it makes me wonder wheter this will make things worse or better for other servers

the only way was to make a fields folder for each server, but i don't know how to code that

if anyone else has your oppinion, i don't know if this should be merged or not

comdevx:
Ok. I accept everyone's decision.

Nipodemos:
@comdevx, @ya4ept mentioned that there is a lot of duplicate maps, you should take a look at that

comdevx:
Can I have a duplicate sample?

comdevx:
i will remove *.field and merge again.

PR #3255, opened 2020-08-02
Fix event macros listlength

ya4ept:
the keyword `&listlength` did not work due to a typo in the code.
Example:
```
 macro listlength {
    log listlength: &listlength(a, b, c)
 }
```


ya4ept:
deleting because of extra commits

PR #3260, opened 2020-08-05
Win32 interface update

alisonrag:
based in #3247 
this pull need to be aproved after 3247

i'm pulling this as it is cause the lib win32::GUI don't have support to unicode characters, so i dont have any plans to update this in the future. i'm moving to TK.

before:
image

after:
image


ya4ept:
image
the player and the NPC are still flickering on the map..
can you fix so you can click on the map?

ya4ept:
deleting the branch?

PR #3265, opened 2020-08-28
Added configuration for hercules private servers

ervinpm:


Nipodemos:
Hello!
do you have a screenshot or video of this data working?

ervinpm:
> Hello!
> do you have a screenshot or video of this data working?

Yes, here are a couple of screen captures of login and fight logs. I masked the ipaddress of our hercules server though:

openkore1
openkore2

Nipodemos:
great! I have one more question, it is problably silly but i never messed with hercules or athena so here it goes:

is it possible that some of these values change depending on the configuration made in the server? because if it does, i don't know if it is really worth to keep configs that only work in a specific condition

ervinpm:
I don't think it will, I'm no Hercules or Athena expert too but I have left the settings of hercules server to their default values. At least these configuration will work on them if someone was to have the same default server settings that I did.

ya4ept:
I think that the localhost server is superfluous in our settings

ahoyaharr:
Unfortunately this config is not working for me on my localhost Hercules server which is running all of the default values:

image

ervinpm:
Are you on the latest version?

Nipodemos:
> I think that the localhost server is superfluous in our settings

why?

ahoyaharr:
Hi, sorry for the late follow up -- the config works, but it seems that packet version `20180621` is more popular on Athena, and many people on Hercules use `20180418`, though I do not know the reason. It might be worth it to make a note somewhere even if it is likely to change in the future.

Anyways, good work! I found it helpful and support this being merged.

sctnightcore (approved):

ya4ept:
> why?

i think very few people use localhost

Nipodemos:
> > why?
> 
> i think very few people use localhost

even so i think is still worth it since it will help begginers by already having it

my only suggestion is to then put a name on these configs to show that is for hercules private servers, as it looks the aethena already changes things and break you configs

ya4ept:
@alisonrag, what do you think about this?

alisonrag:
> @alisonrag, what do you think about this?

i think now need to update the info, cause if i'm not wrong current default clientdate is different

ya4ept:
Who has Hercules private servers?

PR #3270, opened 2020-09-19
fix "quest list" command

ya4ept:
the questID field is now 6 digits

see: https://github.com/OpenKore/openkore/commit/f5441d2a695ae8331f425124d65474e9cde8633b#commitcomment-42449551:

> Can you fix the quest list command that only show quest ID only 5 digit, because the official server i play the quest id have 6 digit

before:
```
quest list
--------------------------------- Quest List ----------------------------------
  0 7124  Battle Basics                   active
  - Lunatic                         5
  1 7126  Selling items                   active
  2 7123  Battle Basics                   active
  - Picky                           0
  3 7127  Battle Basics                   active
  - Willow                          0
-------------------------------------------------------------------------------
```
after:
```
quest list
--------------------------------- Quest List ----------------------------------
  0 7126   Selling items                   active
  1 7123   Battle Basics                   active
  - Picky                           1
  2 7127   Battle Basics                   active
  - Willow                          2
-------------------------------------------------------------------------------
```

### DIFF
--- a/src/Commands.pm
+++ b/src/Commands.pm
@@ -6784,7 +6784,7 @@ sub cmdQuest {
 		my $msg .= center(" " . T("Quest List") . " ", 79, '-') . "\n";
 		foreach my $questID (keys %{$questList}) {
 			my $quest = $questList->{$questID};
-			$msg .= swrite(sprintf("\@%s \@%s \@%s \@%s \@%s", ('>'x2), ('<'x4), ('<'x30), ('<'x10), ('<'x24)),
+			$msg .= swrite(sprintf("\@%s \@%s \@%s \@%s \@%s", ('>'x2), ('<'x5), ('<'x30), ('<'x10), ('<'x24)),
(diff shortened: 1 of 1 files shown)


src/Commands.pm
```diff
@@ -6784,7 +6784,7 @@ sub cmdQuest {
 		my $msg .= center(" " . T("Quest List") . " ", 79, '-') . "\n";
 		foreach my $questID (keys %{$questList}) {
 			my $quest = $questList->{$questID};
-			$msg .= swrite(sprintf("\@%s \@%s \@%s \@%s \@%s", ('>'x2), ('<'x4), ('<'x30), ('<'x10), ('<'x24)),
+			$msg .= swrite(sprintf("\@%s \@%s \@%s \@%s \@%s", ('>'x2), ('<'x5), ('<'x30), ('<'x10), ('<'x24)),
```
Nipodemos, line 6787: I'm problably wrong but, if this is the number of caracters the the mission number will ocupy, shouldn't be six? or it is couting starting from zero?

Nipodemos (changes requested):

src/Commands.pm
oleokoong: I change this line to x5 and its work

ya4ept: it `s Magic

Nipodemos (approved):
good them 😄

PR #3293, opened 2020-10-15
[tRO] 2020-10-15 update + new charblocksize (175)

alisonrag:


alisonrag:
not full tested, i will update later

PR #3294, opened 2020-10-18
fixed parameters of automacros

ya4ept:
- fixed regular expressions for stricter validation of incoming data
this is a bad example of a macro that shouldn't work
```
automacro MyAuto3 {
  InMap morocc
  delay test5
  call myMacro
 }
```
this fix will now get the error:
изображение


Nipodemos (approved):
very nice man 😄

PR #3301, opened 2020-11-11
[Connection] xKore add option to inject Dll or not

alisonrag:
This pull add a option so user can choose if he want to inject the dll in xkore 1 mode, usefull for users that use their own method to inject dll

Default is 1 (inject DLL)


ya4ept (approved):

alisonrag:
https://openkore.com/wiki/XKore_injectDLL

PR #3199, opened 2020-06-22
[fix] add check to skill lvl in skill use

alisonrag:
this fixes #3197

PR #3324, opened 2020-12-13
[FIX] auto select the skill lvl when lvl in config is higher than the character have

alisonrag:
follow up to: #3199

before:
image

after:
image

### DIFF
--- a/src/AI.pm
+++ b/src/AI.pm
@@ -603,8 +603,8 @@ sub ai_skillUse {
 	}
 
 	if ($char->{skills}{$args{skillHandle}}{lv} < $args{lv}) {
-		debug "Attempted to use skill (".$args{skillHandle}.") level ".$args{lv}." which you do not have, adjusting to level ".$char->{skills}{$args{skillHandle}}{lv}.".\n";
-		return;
+		debug "Attempted to use skill (".$args{skillHandle}.") level ".$args{lv}." which you do not have, adjusting to level ".$char->{skills}{$args{skillHandle}}{lv}.".\n", "ai";
+		$args{lv} = $char->{skills}{$args{skillHandle}}{lv};
 	}
 
 	AI::queue("skill_use", \%args);


sctnightcore (approved):

ya4ept:
@alisonrag
what config did you use?

alisonrag:
> @alisonrag
> what config did you use?

```
useSelf_skill AL_HEAL {
    lvl 10
    hp < 90%
    sp > 10%
    whenStatusInactive EFST_POSTDELAY
}

useSelf_skill AL_INCAGI {
    lvl 10
    inLockOnly 1
    notWhileSitting 1
    whenStatusInactive EFST_INC_AGI, EFST_POSTDELAY
    sp > 10%
}

useSelf_skill AL_BLESSING {
    lvl 10
    inLockOnly 1
    notWhileSitting 1
    whenStatusInactive EFST_BLESSING, EFST_POSTDELAY
    sp > 10%
}

useSelf_skill MO_CALLSPIRITS {
    lvl 5
    sp > 10%
    spirit < 5
    notWhileSitting 1
    whenStatusInactive EFST_POSTDELAY
    inLockOnly 1
}

attackComboSlot MO_CHAINCOMBO {
    lvl 5
    sp > 50%
    afterSkill MO_TRIPLEATTACK
    waitBeforeUse 0.2
    dist 1.5
    isSelfSkill 1
}
```
A friend just copied and pasted
Blessing is lvl 10 in config but in monk is lvl 3

alisonrag:
Also blessing block was before of other skills block, before of this pull openkore was in a infinity looping and openkore was not using the next skill block

ya4ept:
@alisonrag see:
https://github.com/OpenKore/openkore/pull/3199#issuecomment-647667237
```
2. We should not automatically substitute a level if it is already registered in our config! The level should be set automatically only if it is not specified in the config
```

alisonrag:
> @alisonrag see:
> [#3199 (comment)](https://github.com/OpenKore/openkore/pull/3199#issuecomment-647667237)
> 
> ```
> 2. We should not automatically substitute a level if it is already registered in our config! The level should be set automatically only if it is not specified in the config
> ```

In this case this config block was blocking next skill to be used

Also level is lower then was set in config, in my point of view we need to pick the max available 

If use this comment as base if someone set wrong level, more then the character have, the skill will be not used

I dont know if dont use the skills just because the level you have is lower is a good thing

As user even if i put higher and wrong level i want to openkore use it in the avaliable level

alisonrag:
The scenario was:
Config: Blessing lvl 10
Character: Blessing lvl 3

openkore was not using blessing and not using the next skill configured

ya4ept:
You can use this block:
```
useSelf_skill AL_BLESSING {
    inLockOnly 1
    notWhileSitting 1
    whenStatusInactive EFST_BLESSING, EFST_POSTDELAY
    sp > 10%
}
```

ya4ept:
see the skill of TaeKwon Kid: [Mild_Wind](https://irowiki.org/wiki/Mild_Wind)
изображение
```
useSelf_skill Mild Wind {
    lvl 4
# fire element
    inLockOnly 1
    notWhileSitting 1
    whenStatusInactive EFST_BLESSING, EFST_POSTDELAY
    sp > 10%
}
```
If i use this block and my character has only level 3 available, then a completely different element will be used.
This way the bot will do 0 damage..

c4c1n6kr3m1:
i agree with title
sorry @ya4ept , for me, it's a mistake to setup config like that
it's uncommon


wait, i ever did something like that....so i edited checkSelfCondition to make it work with eval

```
attackSkillSlot GS_RAPIDSHOWER {
	inInventory Bloody Shell >= 10
	eval ($::char->{skills}{GS_RAPIDSHOWER}{lv} >= 5)
	label useRS
	disabled 0
	lvl 5
	dist 19
	sp > 25%
	notOnAction skill_use,teleport
	whenStatusInactive EFST_POSTDELAY
```

PR #3336, opened 2021-01-10
add new TableFile "pets.txt"

ya4ept:
1. this file contains the names of the monsters that the pets look like
2. this file can be obtained from "data\luafiles514\lua\datainfo\petinfo.lua"
3. this is an optional file

If a bot meets a pet "Alice", but before that time it has never seen this monster, then the pet type will always be "Unknown".
This patch fixes this problem.


ya4ept:
I have an idea for a better one:
1.we can add pet names to monsters.txt file.
2.add a function to overwrite the monsters.txt file if the monster name is different from the default

alisonrag:
> I have an idea for a better one:
> 1.we can add pet names to monsters.txt file.
> 2.add a function to overwrite the monsters.txt file if the monster name is different from the default

need to check the code, but monster.txt uses the name given from server, sometimes some monsters has different names (event/quest)
Example:
assassin test quest

also take a look at:
https://github.com/OpenKore/openkore/blob/29fe641ca35879b62cfb970a9bf9c2c091694b3b/src/Misc.pm#L2320-L2323

alisonrag (approved):

ya4ept:
[current logic:](https://github.com/OpenKore/openkore/blob/f1be711b022bdd14658c7abe66050b2201968f0c/src/Network/Receive.pm#L2589)
We only add the name of the monster if its id is not found in monters.txt

If I write `1002 abracadabra` to monsters.txt, then when poring is encountered, this value will never change.

I believe we should use the current logic (my second option) instead of adding a new file.
@alisonrag what do you think?

alisonrag:
pet have their own actor type, 

> [current logic:](https://github.com/OpenKore/openkore/blob/f1be711b022bdd14658c7abe66050b2201968f0c/src/Network/Receive.pm#L2589)
> We only add the name of the monster if its id is not found in monters.txt
> 
> If I write `1002 abracadabra` to monsters.txt, then when poring is encountered, this value will never change.
> 
> I believe we should use the current logic (my second option) instead of adding a new file.
> @alisonrag what do you think?

will this function be safe? what about assassin test? in this test a lot of same monster have different names
for every monster in the screen openkore will re-write the name in monsters.txt

to me the file is better, just like the ragexe

ya4ept:
> in this test a lot of same monster have different names
for every monster in the screen openkore will re-write the name in monsters.txt

Now the ml command reads the original names of the monsters without overwriting the monsters.txt file.
After my second patch we will rewrite monster names if the current name does not match monsters.txt

After the thief quest, most likely monsters.txt will contain incorrect monster names, but this will be corrected when the bot meets a real monster. Or we can add a dirty hack for this location

alisonrag:
If logic use name given from server to me is ok update monster

c4c1n6kr3m1:
long long  time ago, me :
- made default monsters.txt from rathena
- made option inside config to turn on/off force actor name Given from server to handle this assasin quest

sctnightcore (approved):

ya4ept:
do not merge this request yet

PR #3343, opened 2021-02-05
fixed cash shop

ya4ept:
- added description to 'cash' and 'cashbuy' commands
- now the 'cash list' command does not require running the 'cash open' command. After all, we receive this information immediately after loading the map
- fixed the 'cash buy <item_name>' command
- added clearing of the variable '$ cashShop {points}' after the store was closed, otherwise the 'cash open' command did not work after the store was closed
- updated the 'cash_shop_list' function. Added support for items 4 bytes ($ self -> {cash_shop_list_pack})
- updated the 'cash_shop_buy_result' function
- update wiki:
https://openkore.com/wiki/Cash
https://openkore.com/wiki/Cashbuy
https://openkore.com/wiki/WhenInGame_requestCashPoints

before:
изображение
after:
изображение

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -3705,7 +3705,7 @@ sub shop_sold_long {
 sub vending_start {
 	my ($self, $args) = @_;
 
-	my $item_pack = $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
+	my $item_pack = $self->{vender_items_list_item_pack_self} || $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
 	my $item_len = length pack $item_pack;
 	my $item_list_len = length $args->{itemList};
 	#started a shop.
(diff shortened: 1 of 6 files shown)


src/Network/Receive.pm
```diff
@@ -3705,7 +3705,7 @@ sub shop_sold_long {
 sub vending_start {
 	my ($self, $args) = @_;
 
-	my $item_pack = $self->{vender_items_list_item_pack_self} || $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
+	my $item_pack = $self->{vender_items_list_item_pack} || 'V v2 C v C3 a8';
```
alisonrag, line 3708: this will cause parser errors and close openkore, in some servers like bRO, unpack string to self vending items is different compared to other vending items.
and some servers like kRO de unpack string is the same

alisonrag (changes requested):

src/Network/Receive.pm
alisonrag: reference:
https://github.com/OpenKore/openkore/pull/2346/files

ya4ept: I'm sorry, I haven't found this code for some reason before

alisonrag (approved):

PR #3365, opened 2021-03-04
[Config] [XKore] add option to auto reply pincode request

alisonrag:
let user decide if openkore or client will reply the pincode request

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -3912,7 +3912,7 @@ sub login_pin_code_request {
 	# This is ten second-level password login for 2013/3/29 upgrading of twRO
 	my ($self, $args) = @_;
 
-	if($args->{flag} ne 0 && ($config{XKore} eq "1" || $config{XKore} eq "3")) {
+	if($args->{flag} ne 0 && ($config{XKore} eq "1" || $config{XKore} eq "3") && !$config{XKore_autoReplyPinCode}) {
 		$timeout{master}{time} = time;
 		return;
 	}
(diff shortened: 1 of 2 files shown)


ya4ept:
What about:
~~} elsif ($args->{flag} == 2) {~~
```} elsif ($args->{flag} == 2 || $args->{flag} == 4) {```

**Review by pointmaster2 — APPROVED** [review]

**Comment by ya4ept** [comment]
under what scenario is it useful to use XKore 1 mode and turn off `XKore_autoReplyPinCode` ?
how about deleting this code?
```perl
	if(($config{XKore} eq "1" || $config{XKore} eq "3") && !$config{XKore_autoReplyPinCode}) {
		$timeout{master}{time} = time;
		return;
	}
```

alisonrag:
> under what scenario is it useful to use XKore 1 mode and turn off `XKore_autoReplyPinCode` ?
> how about deleting this code?
> 
> ```perl
> 	if(($config{XKore} eq "1" || $config{XKore} eq "3") && !$config{XKore_autoReplyPinCode}) {
> 		$timeout{master}{time} = time;
> 		return;
> 	}
> ```

uKore bypass auto send pincode using client, if openkore send too it cause DC, this is why i added this key, if user want to use openkore or the bypass to send the pincode

ya4ept:
> uKore bypass auto send pincode using client

the user enters the PIN code by himself?

alisonrag:
> > uKore bypass auto send pincode using client
> 
> the user enters the PIN code by himself?

user configure a file and the bypass send the inputs to client insted of openkore

src/Network/Receive.pm
```diff
@@ -3912,7 +3912,7 @@ sub login_pin_code_request {
 	# This is ten second-level password login for 2013/3/29 upgrading of twRO
 	my ($self, $args) = @_;
 
-	if($args->{flag} ne 0 && ($config{XKore} eq "1" || $config{XKore} eq "3")) {
```
ya4ept, line 3915: why did you decide to remove this check?
`$args->{flag} ne 0`

PR #3367, opened 2021-03-06
[alertSound] fixed eventList item

ya4ept:
Fixed a bug in alertSound in which the conditions "eventList item name1, item name2" did not work.

```
eventList  item Red Potion, item cards, item *Card*, item *cotton*, item *potion*
```
the `item *cotton*` condition does not work, since parsed incorrectly:
```
foreach (split /\,/, $eventList) {
```

### DIFF
--- a/plugins/alertSound/README.txt
+++ b/plugins/alertSound/README.txt
@@ -1,4 +1,4 @@
-Version 9
+Version 10
 
 alertSound($event)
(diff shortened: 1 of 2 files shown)


alisonrag (approved):

alisonrag:
As i can see before was parsing just the last one
Now can parse the entire list

PR #3376, opened 2021-03-18
Improved runFromTarget and slave dancing

Henrybk:
1 - Fixed wall avoiding in runFromTarget that was broken since the A* pathfinding update.
2 - Added runFromTarget_inAdvance, which makes it so you start avoiding/kiting the mob as soon as you attack, using the same idea as the dancing logic, which sends attack and move commands at the same time.
3 - Used the path prediction algorithmn used in meetingPosition in runFromTarget to better predict the target path and avoid it.
4 - Reworked the dancing algorithmn by testing the best combination of attack/move packets sent using this plugin I created 
https://github.com/Henrybk/openkore/blob/a0e07c73c11ca9397bb315f8c8e68d9b5c2cce10/plugins/testAttackDelay/testAttackDelay.pl
Reducing the delay between attack from 0.52s to 0.37s (From 1.05s not dancing), tested with a lvl 97 Amistr.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -216,9 +216,10 @@ route_tryToGuessMissingPortalByDistance 1
 route_reAddMissingPortals 1
 
 runFromTarget 0
+runFromTarget_inAdvance 0
 runFromTarget_dist 5
 runFromTarget_minStep 7
-runFromTarget_maxStep 9
+runFromTarget_maxStep 11
 runFromTarget_maxPathDistance 13
 
 saveMap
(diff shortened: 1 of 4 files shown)


alisonrag (approved):

ya4ept:
Please add a description of the new config parameters:

1. runFromTarget_inAdvance

alisonrag:
> Please add a description of the new config parameters:
> 
> 1. runFromTarget_inAdvance

https://openkore.com/wiki/RunFromTarget_inAdvance

Issue #3361, opened 2021-02-26
monsters name problem of lastest version


hubertyang88:
## ------------------ Openkore Issues Template ------------------

* **Openkore version git**: 
Latest 
* **Server**: 
cro
* **Bug Report / Feature Request**:
something wrong with mon name
when set in mon_control
1081 0 0 0
kore will leave Yellow Plant alone
but when set
黄草 0 0 0 #“Yellow Plant 0 0 0”
kore will attck Yellow Plant , monsters name don't work

this also affect attackSkillSlot when use monsters name ,  only monster ID works

attackSkillSlot skills {
	lvl 2
	dist 1
	maxDist 3
	whenStatusInactive status
	monsters 1113,1009,1049
	timeout 2
}

Strangely , name of some new event monster works , kore didn't read monsters.txt only read new monsters ?
#3299  is this same problem?
* **Summary**:

alisonrag:
just to have sure, can u delete the monsters.txt?

alisonrag:
can u add this line at CoreLogic line 3074
```message "we are trying to check : " . $monster->{name} . " \tlc: ".lc($monster->{name})." \tID: ".$monster->{nameID}."\n";```

![image](https://user-images.githubusercontent.com/10372732/109331097-2db25e80-783b-11eb-9b16-47e5ce293e43.png)

and send log here?

alisonrag:
you will see messages like this:
![image](https://user-images.githubusercontent.com/10372732/109331427-9699d680-783b-11eb-828d-6dd402d0931b.png)

alisonrag:
also send your mon_control here

hubertyang88:
As your request , I remove monsters.txt , and run kore , this file is generated again , see zip file , console and mon_control included.
[mon.zip](https://github.com/OpenKore/openkore/files/6053340/mon.zip)
In console , same monster first check -1 and then check 1.  As I set in mon_control, non of those mons should be killed twice.
For 2218 event monser , always check -1, kore seems ok with this one...

PR #3401, opened 2021-04-16
bug fix for mon_control

ya4ept:
In the packages "actor_display" and "actor_info" there is a space in the name of the monster at the end of the word.
изображение
This is typical for Russian monsters.
Because of this openkore stores the monster name as "Дикая кошка " instead of "Дикая кошка" (Wild Rose)
Thus, if you write the name of the monster instead of its ID in the mon_control.txt file, then these settings do not work.
Also, NPC names contain a space in the name.

Now in the file mon_control.txt you can specify Russian names of monsters. This error could be with other non-English languages.

PS: i don't understand this code
```perl
# New actor_display packets include the player's name
if ($args->{switch} eq "0086") {
	$actor->{name} = $args->{name};
} else {
	$actor->{name} = bytesToString($args->{name}) if exists $args->{name};
}
```
"0086" package does not contain the actor name:
`'0086' => ['actor_display', 'a4 a6 V', [qw(ID coords tick)]],`
I am sure that we can easily shorten this code:
изображение

### DIFF
--- a/src/Network/Receive.pm
+++ b/src/Network/Receive.pm
@@ -1737,7 +1737,9 @@ sub actor_display {
 	#### Initialize ####
 
 	my $nameID = unpack("V", $args->{ID});
-
+	my $name = bytesToString($args->{name});
+	$name =~ s/^\s+|\s+$//g;
+	
 	if ($args->{switch} eq "0086") {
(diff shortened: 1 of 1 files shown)


alisonrag:
this also fixes #3361

alisonrag (approved):

PR #3411, opened 2021-04-25
fix pauseCharLogin

ya4ept:
- now the behavior of "pauseCharLogin" is similar to "pauseCharServer" and "pauseMapServer"
- now the "char_login_pause" timeout is configured in one place - in the config

I have not tested this change, but it seems to me that it is now easier to configure.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -45,7 +45,7 @@ inGameAuth 0
 
 macAddress
 
-pauseCharLogin 1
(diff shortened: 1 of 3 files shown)


alisonrag (approved):

ya4ept:
fixed the documentation: https://openkore.com/wiki/PauseCharLogin

PR #3412, opened 2021-04-25
code cleanup

ya4ept:
adminPassword "xxxxxxxxxx" no longer topical

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -50,7 +50,6 @@ pauseCharServer 0
 pauseMapServer 0
 ignoreInvalidLogin 0
 ignoreAntiCheatWarning 0
-secureLogin_requestCode
 
 # Opening cash shop when connected to map server (recv/ST0)
 whenInGame_requestCashPoints 0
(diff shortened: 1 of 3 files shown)


ya4ept:
@alisonrag, any idea why the `secureLogin_requestCode` parameter is duplicated in our config? I suggest deleting it from config.txt

alisonrag:
> @alisonrag, any idea why the `secureLogin_requestCode` parameter is duplicated in our config? I suggest deleting it from config.txt

https://github.com/OpenKore/openkore/blob/e8b5f9e02bb2ba6c5d75b68834d2ea707253376b/src/Network/DirectConnection.pm#L378-382

i think old openkore developers used to like to duplicate keys between config.txt and servers.txt hahahah

if i'm not wrong, long time ago `gameguard`used to work in servers.txt and config.txt

alisonrag (approved):

ya4ept:
updated the documentation: https://openkore.com/wiki/secureLogin_requestCode

PR #3415, opened 2021-04-26
uppercase macAddress

alisonrag:


src/Network/Send.pm
```diff
@@ -321,6 +321,7 @@ sub reconstruct_master_login {
 
 	$args->{ip} = '192.168.0.2' unless exists $args->{ip}; # gibberish
 	$args->{mac} = $config{macAddress} || '111111111111' unless exists $args->{mac}; # gibberish
```
ya4ept, line 323: What about:
```perl
	$args->{mac} = uc($config{macAddress}) || '111111111111' unless exists $args->{mac}; # gibberish
```

ya4ept (changes requested):

src/Network/Send.pm
alisonrag: i need to test, idk what perl return if we use uc in a not defined var

alisonrag:
test:
```
use strict;
use warnings;

my $test;
my $output = uc($test) || 'test';
print $output."\n";
```

output:
```
$perl main.pl
test
Use of uninitialized value $test in uc at main.pl line 5.
```

this will still works, but will drop a warning in user console, what do you think? @ya4ept should we keep like it is?

ya4ept:
```
$args->{mac} = uc($args->{mac});
```

alisonrag:
what about now?

ya4ept:
`unless exists $args->{mac};` it is not equivalent `if (exists $args->{mac})`
why did you decide to change it?

alisonrag:
> `unless exists $args->{mac};` it is not equivalent `if (exists $args->{mac})`
> why did you decide to change it?

sorry, my mistake,  updated

ya4ept (approved):
looks nice

PR #3363, opened 2021-03-02
Exchange checkLineWalkable for canMove and fix follow sitAuto Idle

Henrybk:
Right now even when openkore is following another player it will activate sitAuto Idle, that is not correct based on our wiki.
This pull fixes that and also fixes some wrong distance checks and walkable line checks (canMove was based on rathena/hercules checks)

PR #3416, opened 2021-04-26
remove followCheckLOS

alisonrag:
now openkore always check the LOS
follow up to #3363


ya4ept:
what about `attackCheckLOS` ?

alisonrag:
> what about `attackCheckLOS` ?
`$config{attackCheckLOS}` is still in use together with `$field->checkLOS()` in 3 or 4 lines of code, i need to check but i think we can update just to canMove()

@Henrybk these 2 functions `checkLOS` and `canMove` do the same thing?

alisonrag:
reading the functions:
checkLOS = check if all cells to target is walkable
canMove = checkLOS + check if the distance between source and target is equal or less to 14

ya4ept (approved):

PR #3421, opened 2021-04-30
add missing partySkill keys

alisonrag:
add missing keys added at #1662
whenPartyMembersNear
whenPartyMembersNearDist

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -664,6 +664,8 @@ partySkill {
 	target_timeout 0
 	target_deltaHp
 	target_dead 0
+	whenPartyMembersNear
+	whenPartyMembersNearDist
 	inInventory
 	isSelfSkill 0
 }
--- a/control/config.txt
+++ b/control/config.txt
@@ -564,6 +564,8 @@ attackComboSlot {
 	monsters
 	notMonsters
 	monstersCount
+	whenPartyMembersNear
+	whenPartyMembersNearDist
 }
 
 doCommand {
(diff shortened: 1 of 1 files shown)


ya4ept:
> Both are self coditions, therefore can be used in partySkill, useSelfSkill, useSelf_item, buyAuto, etc.

we can add to other blocks too

alisonrag:
added to blocks where i found usefull, do you think need to add to some other?

ya4ept:
What about `attackComboSlot`?

alisonrag:
> What about `attackComboSlot`?

added to attackComboSlot.

ya4ept:
@alisonragб have an idea why we need to check for the total number of people [here](https://github.com/OpenKore/openkore/blob/master/src/Misc.pm#L5050):
```perl 
if ($config{$prefix."_whenPartyMembersNear"}) {
	# Short circuit if there's not enough players nearby, party members or not
	# +1 account for self
	return 0 unless inRange(scalar @{$playersList} + 1, $config{$prefix."_whenPartyMembersNear"});

	# Short circuit if there's not enough players in our party
	return 0 unless inRange(scalar @partyUsersID, $config{$prefix."_whenPartyMembersNear"});
	...
```
I think this is an extra check:
``` perl
return 0 unless inRange(scalar @{$playersList} + 1, $config{$prefix."_whenPartyMembersNear"});
```

ya4ept:
I looked more closely. I think everything is fine:
1. `scalar @{$playersList}` - this shows the number of people around, including the bot itself
2. `scalar @partyUsersID` - this always shows the number of players in the party (online and offline), including the bot itself
This code allows you to understand at an early stage that there are very few players around

PR #3444, opened 2021-06-13
added new option "itemsGatherAuto_notInTown"

ya4ept:
now we can ignore objects on the ground only in the city

```
itemsGatherAuto_notInTown <boolean>
If this option is set, disables item gathering when in a city.
```


alisonrag (approved):

ya4ept:
https://openkore.com/wiki/itemsGatherAuto_notInTown

PR #3450, opened 2021-06-30
party_bug_autovivification

ya4ept:
The "onlyWhenSafe" condition did not work correctly.
Fixed a defect when the bot considered all the people around it to be members of its party (like #3449)

Steps to reproduce the problem:

1. Use config:
```lua
doCommand pl;;eval Log::warning "I'm safe \n" {
    timeout 3
    onlyWhenSafe 1
}
partySkill 142 {
    isSelfSkill 1
    dist 3
    timeout 5
    disabled 0
    notPartyOnly 1
}
```
2. find several other players

https://user-images.githubusercontent.com/7117363/123915275-00142800-d989-11eb-9724-96185e13bb44.mp4

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2800,10 +2800,6 @@ sub processPartySkillUse {
 						# party member should be online, otherwise it's another character on the same account (not in party)
 						next unless $char->{party}{users}{$ID} && $char->{party}{users}{$ID}{online};
 					}
-
-					# if that intended to distinguish between party members and other characters on the same accounts, then it didn't work
-					my $player = $playersList->getByID($ID);
-					next if (($char->{party}{users}{$ID} && $char->{party}{users}{$ID}{name} ne $player->{name}) && !$config{"partySkill_$i"."_notPartyOnly"});
 				}
 
 				my $player = Actor::get($ID);
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -2794,17 +2794,16 @@ sub processPartySkillUse {
 					next if ((!$char->{slaves} || !$char->{slaves}{$ID}) && !$config{"partySkill_$i"."_notPartyOnly"});
 					next if (($char->{slaves}{$ID} ne $slavesList->getByID($ID)) && !$config{"partySkill_$i"."_notPartyOnly"});
 				} elsif ($playersList->getByID($ID)) {
-					my $partyUsersID = $char->{party}{users}{$ID};
 					unless ($config{"partySkill_$i"."_notPartyOnly"}) {
-						next unless $char->{party}{joined} && $partyUsersID;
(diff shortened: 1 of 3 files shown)


alisonrag (approved):

src/AI/CoreLogic.pm
```diff
@@ -2798,12 +2798,12 @@ sub processPartySkillUse {
 						next unless $char->{party}{joined} && $char->{party}{users}{$ID};
```
ya4ept, line 2798: we have three checks:
1) `next unless $char->{party}{joined} && $char->{party}{users}{$ID};`
2) `next unless $char->{party}{users}{$ID} && $char->{party}{users}{$ID}{online};`
3) `next if (($char->{party}{users}{$ID} && $char->{party}{users}{$ID}{name} ne $player->{name}) && !$config{"partySkill_$i"."_notPartyOnly"});`

I don't like the third check. It seems to me that she's never used.
This code can only be called when there is another player near OpenKore
We have several OpenKore states:
1) OpenKore is not in a party + "notPartyOnly 0" condition - the (1) condition will be triggered `$ char -> {party} {joined} = undef`
2) OpenKore is not in a party + "notPartyOnly 1" condition - all three conditions are ignored
3) OpenKore in a party + "notPartyOnly 0" condition + a regular player next to it - the (1) condition will be triggered `$ char -> {party} {users} {$ ID} = undef`
4) OpenKore in a party + "notPartyOnly 1" condition + a regular player next to it - all three conditions are ignored
5) OpenKore in a party + "notPartyOnly 0" condition + next to the party member - all three conditions are ignored
6) OpenKore in a party + "notPartyOnly 1" condition + a party member next to it - all three conditions are ignored
7) OpenKore in a party + "notPartyOnly 0" condition + next to another char from the party member's account - the (2) condition will be triggered `$ char -> {party} {users} {$ ID} {online} = undef`
8) OpenKore in a party + the "notPartyOnly 1" condition + next to another char from the party member's account - all three conditions are ignored

So I suggest removing the third condition

alisonrag:
```
6. OpenKore in a party + "notPartyOnly 1" condition + a party member next to it - all three conditions are ignored
```
$ID is accountID, characters in same account have the same accountID so it check if is the same character of party or other of the same account.
the third condition seems that should active when notPartyOnly is 1, in my point of view the second condition should not be called because it is called only when notPartyOnly 0

also this third condition is wrong because:
it only is called when notPartyOnly is 1 and this piece of code `!$config{"partySkill_$i"."_notPartyOnly"}` will always be false and next if will never be called

so or remove or try to understand the real intention of this piece of code and fix it

ya4ept: let's delete it

alisonrag: seems ok to me delete it, this condition is never called =p

ya4ept: done

alisonrag (approved):

PR #3459, opened 2021-07-12
vRO_auth plugin update

ya4ept:
- changed the hook to get new data every time I reconnect
- added auto logout if serverType is not "vRO"


abc9900:
vRO: always disconnect from server when idle for a few sec, or use bot follow. Mage attack skill can not config, only can use by console command.

alisonrag (approved):

PR #3467, opened 2021-07-14
[twRO] support

sctnightcore:


tables/twRO/recvpackets.txt
```diff
@@ -1,1548 +1,1583 @@
...
+09FE -1 83 1
+09FF -1 84 1
+0A00 269 269 1
+0A01 3 3 0
+0A02 4 4 1
+0A09 55 55 1
+0A0A 57 57 1
+0A0B 57 57 1
+0A0C 66 66 1
+8B00000A -1 4 1
```
alisonrag, line 1010: need to check this packet

sctnightcore: fixed

alisonrag (approved):

tables/twRO/recvpackets.txt
```diff
@@ -1,1548 +1,1583 @@
...
-0A04 6
-0A05 63
-0A06 6
-0A07 9
-0A08 26
-0A09 55
-0A0A 57
-0A0B 57
-0A0C 66
-0A0D -1
```
ya4ept, line 1200: `inventory_items_nonstackable 0A0D`
we need to check this package. It is missing from the new version of the file

ya4ept (changes requested):

tables/twRO/recvpackets.txt
ya4ept: @alisonrag do we only need to check the last packet or all three?

control/config.txt
```diff
@@ -839,4 +839,4 @@ logToFile_Errors
 logToFile_Messages
 logToFile_Warnings
 history_max 50
-gameGuard 0
+gameGuard 1
```
alisonrag, line 842: this should not be here

ya4ept: We will remove this when we merge with the master. I am planning to remove this setting from config.txt

ya4ept (approved):

PR #3480, opened 2021-07-19
update attack unfail

alisonrag:
timeout to unfail is too low, depending in how many monsters openkore cant reach it will lead to a infinity looping trying to leave and attack same monster.
image

this pull "fix" this by ignoring the monster for more time, this should be enough to avoid stuck

### DIFF
--- a/control/timeouts.txt
+++ b/control/timeouts.txt
@@ -66,7 +66,7 @@ ai_homunculus_attack_waitAfterKill 0.3
 ai_mercenary_attack_waitAfterKill 0.3
 
 ai_attack_unstuck 2.75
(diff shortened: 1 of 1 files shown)


ya4ept:
my friends say it didn't help

kou90258:
@alisonrag
attackChangeTarget 0
attackCheckLOS 0
The problem is reduced after the above two are closed.
Because these two items are long-range attack functions, the close-range attacker accidentally opens them, and there will be error problems?
Can't these two functions be directly limited to the attack distance <2, can you use the program to limit it?

alisonrag:
implemented at #3493

PR #3481, opened 2021-07-19
update fields up to kRO July, 2021

alisonrag:
keep in a separated folder old maps:
prontera (is the same, but just in case)
alberta
izlude
payon
morroc


alisonrag:
> what have you done with these fields?
> image

i just took all maps from kRO and replaced in fields, this maps should be the same and still in openkore

alisonrag:
image

alisonrag:
altered to draft, portals.txt still has 300+ npc portals to review

alisonrag:
reverted unrelated changes made, and now only fields is updated

ya4ept:
> reverted unrelated changes made, and now only fields is updated

why?

alisonrag:
> > reverted unrelated changes made, and now only fields is updated
> 
> why?

Check #3493 and #3494

PR #3492, opened 2021-07-28
Compile perl 5.22

vutung2311:
Because of the EOL of PDK as in: 
`https://www.activestate.com/blog/perl-dev-kit-pdk-now-end-of-life/ `
and complication of adding icon to Windows executable as in:
`https://stackoverflow.com/questions/22607334/create-exe-with-custom-icon-using-perl-par-packer`
`https://www.perlmonks.org/?node_id=1192706`

I want to drop the way we're shipping binaries now and shipping `.cmd` file instead.
To ship `.cmd` file, I added a dependencies installation `.bat` file for user to install all dependencies for running openkore. This is not ideal because it's not about downloading and run anymore but with this we save more maintenance cost on developer's side.
If we still want to pursue the way of downloading and run then we can think about doing local lib vendoring.
Also I fixed few issues about Perl XS which cause access violation, adding `-msse3` with the hope that it will increase performance when dealing with matrix (although I'm not sure that we have any logic about dealing with matrix). And lastly I added some improvement for scons scripts.


alisonrag:
i can't see the changes, is a bug at github?

vutung2311:
> i can't see the changes, is a bug at github?

Oh I've just rebase it. Lemme update the branch.

PR #3496, opened 2021-07-29
fix hook packet_pre/sendSkillUse

Sean820729:
- fix hook packet_pre/sendSkillUse doesn't pass any arguments.
- remove unused code
image


ya4ept:
Your request does not make sense, I have to reject it.
This hook should not have any default arguments.
This hook is needed in order to change the packet before sending it to the server.
Example: [pp_workaround.pl](https://github.com/OpenKore/openkore/blob/master/plugins/needs-review/Obsolete/legacy/pp_workaround/trunk/pp_workaround.pl)

PR #3497, opened 2021-07-30
generator.exe is now working

ya4ept:
generator.exe does not work. At startup, it gives an error:
image
because it by default looks for file `1.pl` which doesn't exist.

The `generator.exe` should actually run the `generator.pl`:
I fixed it:
image

--------
**Known issues:**
- [ ] The resulting file cannot pass command line parameters


ya4ept:
```
mingw32-make.exe
g++ -Wall launcher.cpp utils.cpp core.cpp hstring.c ui.o -o launcher.exe -mno-cygwin -s
launcher.cpp: In function 'int main(int, char**)':
launcher.cpp:71: warning: deprecated conversion from string constant to 'char*'
```
but the launcher.exe file is generated

alisonrag:
some dlls is missing, maybe need to add flags in compile to include it
image

alisonrag:
> some dlls is missing, maybe need to add flags in compile to include it image

@ya4ept

FrozenCube:
I do recommend use library instead dlls. About import!

PR #3500, opened 2021-07-31
fixed build for windows + perl 5.32

ya4ept:
our artifacts are getting messed up for building windows + perl 5.22
this is a temporary solution

- [x] we have to add perl version check to SConstruct


ya4ept:
let's merge

alisonrag (approved):

PR #3502, opened 2021-08-07
Packetdescriptions updated

ya4ept:
- just updated packetdescriptions.txt
- minor package fixes


tables/packetdescriptions.txt
```diff
@@ -413,7 +428,6 @@
 080B Booking Delete
 080E Party User HP Info
 080F Deal Item Added (Other)
-081B Buying Store Item Update
```
alisonrag, line 416: this seems that u just removed and not added again

ya4ept: this line is duplicated below

alisonrag (approved):

PR #3503, opened 2021-08-08
Merge master to compile_perl_5.22

vutung2311:
This is for reducing amount of changes in https://github.com/OpenKore/openkore/pull/3492


ya4ept:
I think this is superfluous in the current branch

PR #3518, opened 2021-09-10
Best spot support current combat character

kou90258:


ya4ept:
@alisonrag note, there are many off-topic fixes like "captcha_image", "guild_storage_log", etc.

PR #3557, opened 2022-01-23
jRO support 

hexhexd:
Let's goooo


ya4ept (dismissed):
I think the `STATUS_id_handle.txt` file is redundant. You can use the common file `tables\STATUS_id_handle.txt`

PR #3566, opened 2022-03-01
Captcha support

alisonrag:
Add Support to Macro Detection Client Interface

Detection UI:
image

captcha command:
image

thanks to @macabeus (BMP uncompress)

PR #3614, opened 2022-08-25
feat: implement macro detector

alisonrag:
remake of #3566

Add Support to Macro Detection Client Interface

Detection UI:
image

captcha command:
image

thanks to @macabeus (BMP uncompress)


ya4ept:
https://github.com/HerculesWS/Hercules/pull/3051

ya4ept (approved):
looks correct

ya4ept:
we need to update the documentation
https://openkore.com/wiki/captcha

alisonrag:
> * [x]   we need to update the documentation
> 
> 
> https://openkore.com/wiki/captcha

updated the above and added the bellow:
https://openkore.com/wiki/Captcha

they are different system

CatOnMyLeg:
What if the captcha is Chinese words?
Because I play in twro. And now they changed to Chinese words.

I can recognize it .  And I try to type in the console :  captcha 出來走走(出來走走 is the answer of the captcha)
The bot will crash.   How should I solve this problem?

PR #3624, opened 2022-09-08
Add missing casting key deletion

Henrybk:
Currently skill location use and skill failed do not delete the {casting} key.

This adds the deletion.


alisonrag (approved):

Henrybk:
Added doc
https://openkore.com/wiki/whileCasting

PR #3626, opened 2022-09-08
add attackAuto_considerDamagedAggressive

Henrybk:
Currently if the bot has damaged a monster but the monster has not yet attacked the bot it won't be considered as aggresive.

This new config key when set to 1 makes openkore consider all monster in the screen who have been damaged by the bot as aggresives

Also deleted the duplicate timeout attack_failed -> ai_attack_unfail check


alisonrag (approved):

Henrybk:
doc: https://openkore.com/wiki/attackAuto_considerDamagedAggressive

PR #3628, opened 2022-09-09
Move run from target to meeting position

Henrybk:
Removes get_kite_position, uses meetingPosition instead.

Also changes another attacking behavior for the situation that a ranged character has already damaged its target but curently is out range 
Right now openkore will use meetingposition to get closer at attackMaxDist range
But if we already damaged the mob that means it is coming at us
So more often than not we will get closer to the mob while it is getting closer to us
This adds a config key that if '1' will make the bot wait for the target instad of going to it.

Video showcasing the branch: https://www.youtube.com/watch?v=ig9J9NiGgp8

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -63,7 +63,6 @@ allowedMaps
 allowedMaps_reaction 1
 
 attackAuto 2
-attackBeyondMaxDistance_waitForAgressive 1
 attackAuto_party 1
 attackAuto_onlyWhenSafe 0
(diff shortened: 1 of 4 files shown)


Henrybk:
Working with slaves: https://youtu.be/1k2f6ailsD4

alisonrag (approved):

ya4ept:
Updated:
https://openkore.com/wiki/runFromTarget_maxStep

PR #3631, opened 2022-09-11
fix runFromTarget_maxPathDistance

Henrybk:
currently runFromTarget_maxPathDistance is not working.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -382,7 +382,7 @@ mercenary_attackDistanceAuto 0
 mercenary_attackMaxRouteTime 4
 mercenary_attackCanSnipe 0
 mercenary_attackCheckLOS 1
-mercenary_attackRouteMaxPathDistance 13
+mercenary_attackRouteMaxPathDistance 20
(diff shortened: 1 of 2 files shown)


ya4ept:
1) is it correct that the parameter `runFromTarget_maxPathDistance=13`, but for the mercenary and homunculus this value is `20`?
2) is it correct that the parameter `attackRouteMaxPathDistance=13`, but for the mercenary and homunculus this value is `20`?

Henrybk:
1 and 2: Yes, the character cannot move more than 14 tiles away from the target or it will lose aggro/get out of sight, however for slaves they can go from one extreme of the screen (say -14/-14 relative to the character) to the other extreme (+14/+14) without the target moving out of the character screen.
In fact they could be 14 for the character and 28 for the slaves, I choose it to be a bit more restrictive and kept it at 13/20.

3: Will do.

Henrybk:
Added description

alisonrag (approved):

PR #3645, opened 2022-09-19
Add onlyIdentified to buyauto and equipAuto

Henrybk:
With this key buyauto will only sum identified items.

equipAuto will now properly equip identified items when there are non identified varients in inventory


ya4ept:
@Henrybk please add this parameter to the documentation:
https://openkore.com/wiki/buyAuto

PR #3663, opened 2022-09-25
Fix search_store_result in RagexeRE_2020_04_01b

Henrybk:
Before:
d47627ff8aec5921bd29581f53637e51

After:
fbabe7f477087f917dd39f3c78f2c23d


alisonrag:
what about rewrite the function in Receive.pm and add just the unpack string to 2020_04_01b servertype?
like the following:
https://github.com/OpenKore/openkore/blob/2b73507a696af43af76b402b7be20348815a3280/src/Network/Receive.pm#L3786

https://github.com/OpenKore/openkore/blob/2b73507a696af43af76b402b7be20348815a3280/src/Network/Receive/bRO.pm#L48

ya4ept:
i think alisonrag is right

PR #3665, opened 2022-09-27
Rework AI clear

Henrybk:
For some strange reason sometimes AI::clear leaves @ai_seq and @ai_seq with different sizes with causes a critical error, very hard to reproduce however we never really should use delete on arrays, this is still unfinished.


Henrybk:
Seems to be working well, runing it since I made this PR with no problems, also I was not able anymore to reproduce the bug.

alisonrag (approved):

PR #3667, opened 2022-09-27
add clientSight_removeBeyond

Henrybk:
Currently openkore removes (actully just never adds) actors that are sent by the server with a distance greater than clientSight, however this comparison is made using {pos_to} and not actual position which can sometimes be greater than clientSight even when close.

This change actually calculates the position of both actors and compares them.
The second issue is that calcPosition is not perfect and usually gets 1 cell wrong, accounting for both actors the distance could be 2 greater of lower than the real one, so I added a config key to account for that.

Now only removes actors that are at least clientSight_removeBeyond beyong the clientSight limit.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -157,6 +157,7 @@ cachePlayerNames_duration 900
 cachePlayerNames_maxSize 100
 
 clientSight 15
+clientSight_removeBeyond 2
 
 dcPause 1
 dcOnDeath 0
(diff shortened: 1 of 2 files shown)


alisonrag (approved):

ya4ept:
@Henrybk, please update the documentation
https://openkore.com/wiki/clientSight_removeBeyond

ya4ept:
@Henrybk @alisonrag 
actually i think "clientSight_removeBeyond" is an extra parameter. We can add guidelines to the description of the "clientSight" parameter: use clientSight + some cells

PR #3676, opened 2022-10-08
feat: Pathfinding v3 + Attack rework

Henrybk:
This is a rework of the current pathfinding algorithm based on my finding looking into the hercules emulator.

This PR actually has a lot of changes so I will be updating this comment little by little while rewiewing the code, for now this is only a draft.

Main changes:
1- Implements the Drunk plugin logic directly into the pathfinding algorithm, in the form of the config key route_randomFactor
(example: https://www.youtube.com/watch?v=sLRZoqQKMeQ)

2- Separates the avoidWalls logic from the weight logic, creating a new variable exclusive for custom weight maps, making, possible and easy the creation of plugins such as my NewAStarAvoid (included) (example: https://www.youtube.com/watch?v=Rov7DZ8wL9s&t=5s)

3- Implements a second pathfinding mode that simulates the client pathfinding by using the same heuristic as it to aquire the most precise position for a given character (calcPosFromPathfinding), this is used to calculate interception points in meetingPosition and alike.

4- Implements 2 new checks based on the client code to define one an actor can move to another spot (Field::canMove) or attack a target at another spot (Utils::canAttack), should never fail.

5- Actually uses client distance to calculate interactions, which is diferent from distance and blockDistance, removing the diference between melee and ranged.

6- Uses the emulator chase logic to approach targets, how this works:
When you send an attack packet to the emulator it checks if you are moving to the target, if you are then it stores the attack intention and checks in the emulator itself after every tick if the attack can be started and starts it when possible.
By using this information we can send after every move packet a new attack packet to start the chase logic and make the emulator attack for us.

Extra changes:
1- AI::findAction rework, added skip parameter.
2- Moved checkMonsterCleanness , provoke check and sitting->stand to before attack::main
3- Unified giveUp logic (except provoke and KS. Maybe TODO?)
4- Unified range and LOS checks in Utils::canAttack and removed them from AI::Attack
5- runFromTarget now loops distance from runFromTarget_dist up to clientSight instead of 0->14
6- Added attackBeyondMaxDistance_waitForAgressive for melee
7- Recalculate slave route whenever the character moves
8- Never move slave to the same cell as the character
9- Add setRoute2 to Wx to print both attack and move route
10- Optimization in meetingPosition to never check the same cell twice

TODO:

- [x] add checks for config key useweapon
- [x] add checks for use skill ( openkore is using autoattack everytime even when only skill is necessary)
- [x] Test route_randomFactor (drunk walk)
- [x] Test melee character without skills
- [x] Test melee character with skills
- [x] Test ranged character without skills and without runFromTarget
- [x] Test ranged character without skills and with runFromTarget
- [x] Test ranged character with skills and without runFromTarget
- [x] Test ranged character with skills and with runFromTarget
- [x] Test ranged character with ONLY skills and without runFromTarget
- [x] Test ranged character with ONLY skills and with runFromTarget
- [x] After above is tested, move new Atack.pm logic to SlaveAttack.pm

BUGS:

1. Even though you can only attack from 1 diagonal cell away or 2 ortogonal cells away, and can't from 1 diagonal + 1 ortogonal, sometimes if the mob is moving when it is attacked the game/bot will glitch and they will attack each other from (1 diagonal + 1 ortogonal) like the example bellow, currently this bugs kore. We should add a way to check for these situations.
This seems to be caused by [rathena range increase when following and continous attacking](https://github.com/rathena/rathena/blob/f4f60a3e915881c4f0b3bee8189bd1d2ce116ef2/src/map/unit.cpp#L2864)
4069ea746e82717bdc8aead0674c796f
- [x] Fixed

Future insights:
- Attacking with aspd bound skills (double strafe, mamonite) is slower than normal attacking by ~20%, maybe that can be improved

Video showing it working:
[Youtube](https://youtu.be/PK9ckS35uNg)


Henrybk:
av_top_index() was introduced in Perl version 5.18, I'll fix later

Henrybk:
I am satisfied with the code as of now, so I am changing this PR from draft to Ready for review.
Needs testing.

hexhexd:
I have been using this patch for few days on my bot that's running 24/7. It's rock solid. 👍

LastRoze:
> I have been using this patch for few days on my bot that's running 24/7. It's rock solid. 👍

I use Henry's repo with his pathing branch and the XSTool.dll failed to load, can you share the working repo, so I can test this too, thanks.

LastRoze:
Haven't got time to test it, will do this Sunday, seems promising 🎉

alisonrag:
> 

install strawberry perl 5.32 32bits:
https://strawberryperl.com/download/5.32.1.1/strawberry-perl-5.32.1.1-32bit.msi

run openkore.pl

if not work just go with cmd until openkore folder type `gmake` and copy the generated dlls to openkore folder

Henrybk:
Updated to master and added a few more checks

Henrybk:
The issue with attackUseWeapon/Skills and attackSendAttackWithMove should be fixed now.

hexhexD:
Maybe it's a good idea to also change how we set maxdistance here: https://github.com/OpenKore/openkore/pull/3676/files#diff-59b062e5109c3e29fefe3eb97e72c6567debf55a12713597e5c3c3445fd1fdfeR364

I had a issue where a attackskillslot has a timeout and that skill will not be selected and the bot will behave like a melee and try to approach the monster.

Henrybk:
Updated to master, lets do this

Henrybk:
> Maybe it's a good idea to also change how we set maxdistance here:
> 
> https://github.com/OpenKore/openkore/blob/21aba245ae05c3533d71f6f3689f5b7d09bdc6d4/src/AI/Attack.pm#L391
> 
> I had a issue where a attackskillslot has a timeout checkSelfCondition returns 0 and the subsequent block will not select that skill. then the bot will behave like a melee and try to approach the monster.

I believe that should not happen since attackmethod remains undefined in that case, however it would probably be smart to set it to something like "none" and take actions based on that.

Henrybk:
Added check list and known bugs to the original post.
Tested melee character without skills, a lvl 30 dagger thief which lvled on bigfoot for 4 hours, no major issues, just the very rare BUG 1.

src/Utils.pm
```diff
@@ -110,67 +150,294 @@ sub calcPosFromTime {
...
+
+# Only God and gravity developers know why this is done this way, but I tested in the client and it works 100% of the time
+#
+# Reference: hercules src\map\path.c distance_client
+sub getClientDist {
+	my ($pos1, $pos2) = @_;
+	my $xD = abs($pos1->{x} - $pos2->{x});
+	my $yD = abs($pos1->{y} - $pos2->{y});
+	my $temp_dist = sqrt(($xD*$xD) + ($yD*$yD));
+	$temp_dist -= 0.0625;
```
alisonrag, line 270: can u confirm this?
https://github.com/rathena/rathena/pull/6950

Henrybk: In Hercules that is the case, I would have to search some more to find if it is correct.
Source:
https://github.com/HerculesWS/Hercules/blob/69edb0cf6de382dfa128b8582617e3bc126514d4/src/map/path.c#L500C1-L500C2

Henrybk: Changed to agree with rathena

Henrybk:
Fixed bug 1 and tested a hunter farming in hu_fild06 using only bow, no skills and no runfromtarget for a few hours, no bugs found.

Henrybk:
Reworked meeting position a bit to simplify it a lot at the cost of a bit more calculations, tested again Thief and Hunter, no issues, tested runfromtarget on hunter, bow only, no skills, seems to be worling well

Henrybk:
Finished testing all use cases described in the original post. Gonna update slaveattack and test it a bit and then I believe this is ready to merge code-wise, will still need a bit of cleaning in the messages/warnings department

ya4ept:
you are a wizard!

Henrybk:
Will also make some optimizations by moving some heavily used functions to XS/C++, initially only checkLOS, maybe later also distances, canMove and canAttack

Henrybk:
Tested homunculus basic attacking and char random_walking while the homunculus is killing mobs, working well enough, can be improved a bit in the future, added TODO messages for the future.

Still needs to be tested:
Homunculus skills
Homunculus runfromtarget
Homunculus dancing (melee and ranged)
Mercenary basic
Mercenary skills
Mercenary runfromtarget

Henrybk:
Oops I did not meant to push the test branch merge of the fix slaves PR, guess I'll have to put this one in a pause until that one is merged.

Henrybk:
All testing complete (albeit done with the slave PR merged here) should be good to merge after a bit of cleaning, commenting and tagging messages.

If anyone intends to test this now would be a good time.

Henrybk:
Ready for testing, changin it from Draft to Pull Request

alisonrag:
after merge change calcPosition to calcPositionFromPathfind https://github.com/OpenKore/openkore/pull/3923/files

alisonrag (approved):

ya4ept:
@Henrybk @alisonrag please complete the documentation:
https://openkore.com/wiki/attackBeyondMaxDistance_sendAttackWhileWaiting
https://openkore.com/wiki/attackSendAttackWithMove
https://openkore.com/wiki/attackWaitApproachFinish
https://openkore.com/wiki/runFromTarget_noAttackMethodFallback
https://openkore.com/wiki/timeouts.txt#AI_Attack (ai_attack_after_skill)

PR #3687, opened 2023-01-02
Fix attckAuto 0 and improve teleportAuto_search check fix #3625

### LINKED ISSUE #3625: attackAuto 0 won't retaliate attacks
According to the wiki:
https://openkore.com/wiki/attackAuto

attackAuto 0: Do not attack any monsters unless they attack you first. (Act like passive monsters.)

However attackAuto 0 will never retaliate

c7a52584e06350b2688391b98e30dc38

hexhexd:
this fixes #3625 
@aggressives wasn't populated due to incorrect predicate use. It makes no sense to check attackAuto value while populating @aggressives

"Perl has no boolean data type.
A scalar in an if statement evaluates to boolean "false" if and only if it is one of the following:
    undef
    number 0
    string ""
    string "0"
"

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -582,7 +582,7 @@ sub processPortalRecording {
 		});
 		$recorded = 1;
 	}
-	
+
 	if ($recorded && $config{portalRecord_recompileAfter}) {
 		Settings::loadByRegexp(qr/portals/);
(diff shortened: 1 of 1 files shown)


hexhexd:
How can I link this pull request with issue #3625?

alisonrag (approved):

PR #3699, opened 2023-01-30
fix: correct reuse of char in character select screen

alisonrag:


ya4ept:
how can we reproduce this problem?

alisonrag:
add a test message at /src/inventorylist/inventory.pm
image

before:
image

after:
image

LastRoze (approved):
Tested on aRO, and seems stable.

ya4ept (approved):

PR #3700, opened 2023-01-30
fix: add support to old items type in item list

alisonrag:


ya4ept:
Updated documentation:
https://openkore.com/wiki/itemListUseOldType

ya4ept (approved):

PR #3739, opened 2023-03-30
feat: move wx alias menu interface to its own menu bar

LastRoze:
**Story:** This pull will change the location of the Alias menu in the WX interface to have its menu bar and add a separator to each self-generated sub-item based on the "alias_" configuration in the (config.txt) file.

**Purpose:** When the user has many aliases config and wants to have quick access to it they need to go to the _Command_ menu first then hover to the _Alias_ sub-menu to be able to run whatever their aliases are and the list of that item is too condense it even worse if your monitor has a large resolution, to mitigate it we changed the Alias menu into its menu bar, and create a separator for each item-list in aliases to make it more readable and can quickly activate any aliases user wants from there.

WX Alias Menu Before
WX Alias Menu Before

WX Alias Menu After
WX Alias Menu After


alisonrag (approved):

ya4ept (approved):
It's OK

PR #3757, opened 2023-05-23
fix: avoid hash ref error when login twice

alisonrag:
before:
https://github.com/OpenKore/openkore/assets/10372732/3a736406-671a-4f29-bfdf-04e527498042

after:
https://github.com/OpenKore/openkore/assets/10372732/dd038130-c292-45cd-9d53-d7ade76f3ae3


ya4ept (approved):
looks good

PR #3765, opened 2023-06-19
chore: update clientsuspend timeout

alisonrag:
openkore was waiting 10 seconds after map load to do some action, thisp pull request reduce this value to 1 second


hexhexD:
I experienced the same thing and changed it to 1 sec as well, and xkore1 has been working great!

ya4ept (approved):
let it be 1 but i use the value 5

PR #3766, opened 2023-06-19
Fix ranged attack distance

copiousfish:
My ranged bots were constantly getting stuck trying to attack immovable enemies with range 9 skills unless their attack distance was 6 or below. This behavior is very prevalent when attacking at a diagonal angle to the monster. With an attack distance of 9, bot would stand 9 horizontal AND vertical blocks away but the monster list would show the distance as around 12. The issue is that the block distance calculation does not do well with angles and keeps them out of range so the game will not let them attack.

This change fixes part of the attack and move calculation to use the more accurate distance subroutine. The change was like night and day and now my bots can attack at the intended 9 range.

My testing on this has been limited to a soul linker with a bard and dancer using instruments.

### DIFF
--- a/src/AI/Attack.pm
+++ b/src/AI/Attack.pm
@@ -298,7 +298,7 @@ sub main {
 	my ($realMyPos, $realMonsterPos, $realMonsterDist, $hitYou);
 	my $realMyPos = calcPosition($char);
 	my $realMonsterPos = calcPosition($target);
-	my $realMonsterDist = blockDistance($realMyPos, $realMonsterPos);
+	my $realMonsterDist = distance($realMyPos, $realMonsterPos);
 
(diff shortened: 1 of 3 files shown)


alisonrag:
can u share config.txt to test?
just attack and skill block

copiousfish:
I was seeing the issue with my Soul Linker spells with these skill blocks and attackMaxDistance:
```
attackMaxDistance 9

attackSkillSlot Esma {
	whenStatusActive Esma
	disabled 0
}

attackSkillSlot Estin {
	whenStatusInactive Esma
}
```

And my Bard with Musical Strike:
```
attackSkillSlot Musical Strike {
	sp > 30%
	maxDist 9
	disabled 0
}
```

copiousfish:
Polluted my master branch with other changes. I'll reopen this later.

PR #3801, opened 2023-10-04
hotfix: avoid infinity loop in processNameRequestQueue

alisonrag:
avoid openkore hang

### DIFF
--- a/src/Misc.pm
+++ b/src/Misc.pm
@@ -3034,7 +3034,7 @@ sub processNameRequestQueue {
 			next;
 		}
 
-		if(defined $actor && $actor->{avoid}) {
+		if (defined $actor && $actor->{avoid}) {
 			debug TF("[NameRequestQueue] Removed from list actor %s (flag avoid).\n", $actor);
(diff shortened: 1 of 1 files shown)


src/Misc.pm
```diff
@@ -3034,7 +3034,11 @@ sub processNameRequestQueue {
 			next;
 		}
 
-		next if ($actor->{avoid});
```
ErikDahlinghaus, line 3037: I was also experiencing this hang. I completely removed the line and have had no issues but showing an error message is probably preferred.

ErikDahlinghaus (approved):

src/Misc.pm
```diff
@@ -3034,7 +3034,11 @@ sub processNameRequestQueue {
...
+		if(defined $actor && $actor->{avoid}) {
```
copiousfish, line 3037:
```suggestion
		if (defined $actor && $actor->{avoid}) {
```

copiousfish (approved):

ya4ept (approved):

PR #3874, opened 2024-08-16
Fix clearing timeouts after auto teleport

maikelvf:
Openkore sets $ai_v{temp}{clear_aiQueue} =1 when doing any form of auto teleporting. This flag in turn clears the entire ai on map change, among which %ai_v. %ai_v contains timeouts for any item, skill or command block from config.txt. If these get cleared, Openkore uses every item or skill that only has a timeout after every auto teleport. This fix still clears the ai, but keeps any timeouts in %ai_v so Openkore doesn't spam skills or items after every auto teleport.

### DIFF
--- a/src/AI.pm
+++ b/src/AI.pm
@@ -134,7 +134,6 @@ sub clear {
 	if ($total == 0) {
 		undef @ai_seq;
 		undef @ai_seq_args;
-		undef %ai_v;
 
 	# If 1 arg was given find it in the queue
 	} elsif ($total == 1) {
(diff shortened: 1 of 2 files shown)


alisonrag:
did u checked if there is any other important key being deleted?
maybe revert this pull too https://github.com/OpenKore/openkore/pull/3191/files?

maikelvf:
Right now on master $ai_v just gets deleted entirely, i've only excluded any key that ends with '_time'. So far while looking through the source code and while testing i have not encountered any other key that ends with '_time' that should be deleted and is now excluded. Even if there were, i don't think it would have any implications if that doesn't get deleted on teleport.

I'm fine with reverting the PR that you linked, but you probably did that for a reason right? This way at least we're clearing anything that is not a timeout.

alisonrag:
i just did thinking that it was the best at that time, but now i know that sometimes is better keep tracking.
just remove the `undef %ai_v; from AI::clear` and i will approve

maikelvf:
@alisonrag Done

alisonrag (approved):

PR #3895, opened 2024-12-01
add new portals and cities

ya4ept:
1. add added portals for "moc_castle"
2. add new cities


tables/cities.txt
```diff
@@ -122,7 +122,7 @@ que_bingo.rsw#Bingo Game Room #
 pvp_n_room.rsw#PvP : Waiting Room#
 p_track01.rsw#Monster Race Arena#
 p_track02.rsw#Monster Race Arena#
-#cmd_fild07.rsw#Beacon Island, Pharos#
+cmd_fild07.rsw#Beacon Island, Pharos#
```
alisonrag, line 125: there is some monster, including agressive ones on this map:
https://www.divine-pride.net/database/map/cmd_fild07

personally i dont care, in 20 years playing ragnarok i never saw anyone using this map to leveling or farm.

some configs will be affected like:
`route_randomWalk_inTown` its disabled by default
`attackAuto_notInTown` its enabled by default

so after this change, openkore will not random walk or retaliate agressive monsters

alisonrag (approved):

ya4ept:
I don't remember the real reasons, but I specifically marked this map as a city..

PR #3906, opened 2024-12-17
Flame barrier logic alpha showcase

Henrybk:
OK so this is much more of a showcase of a project I had last year (or was it 2022?) than a pull request. 
I decided to put this in here in case someone finds it interesting and decides to test/help/develop in the future.

[Video showcase](https://www.youtube.com/watch?v=7Y5FLy_i5ro)

The code is full of typos and unedited comments and config edits, as I said, more of a showcase.

Can be closed if wanted.


Henrybk:
> looks interesting, thanks.
> 
> I wonder if it can be made into a separate plugin, without editing the main code?

I believe it could be done but it would require a refactor of the Attack code to use many consecutive checks interposed with hook calls instead of a elsif chain like it os done today.

Something like:
Callhook (attack1)
If (foo) { Run } 

Callhook (attack2)
If (bar) { Attack }

Much like buyauto/sellauto works

ya4ept:
If you add this code to the main branch, you can make it more versatile. For example, you can use the trap skill

Henrybk:
As soon as Path v3 is done and merged I will try to convert this one to plugin form then

PR #3914, opened 2024-12-23
Fix slave skills

### LINKED ISSUE #3913: Homunculus skills not working
The 'homun skills' command is not working as shown bellow.
73b504aad4a49c79aa383dfef93ae29c

It show no level or sp for the skills.
This is because $char->{skills} is not saving the slave skills, and $slave->{'slave_skillsID'} only saves the handles as shown bellow
db43550d3e0ec85c4a26d32054d12ed0

Also the sub that checks ownership of the skill is just empty.
460a2988a0cb81b9a676a5bc80239431

Henrybk:
Fixes #3913 

Each actor now holds their own skill list.

Possible problem is that I don't know actually how skills are handled in the server, if they are coded as a character skill then their use must remain in AI::Corelogic, if however slaves are their own separate entity with cooldowns of their own then it would probably be wise to move their skill usage to their AI.

58121ea4deb7469df1977f96e9d7dff3

EDIT:
It does seem char and slave are completely independent, I'll move the skill use logic to their AI

EDIT2:
Done, just gotta test it

Edit 3:

Video showing it working:
[Youtube](https://youtu.be/PK9ckS35uNg)

### DIFF
--- a/src/AI.pm
+++ b/src/AI.pm
@@ -593,8 +593,12 @@ sub ai_skillUse {
 		debug "[$owner] Attempted to use skill (".$args{skillHandle}.") level ".$args{lv}." which you do not have, adjusting to level ".$lvl.".\n", "ai";
 		$args{lv} = $lvl;
 	}
-
-	AI::queue("skill_use", \%args);
+	
+	if ($skill->getOwnerType == Skill::OWNER_CHAR) {
+		AI::queue("skill_use", \%args);
+	} else {
+		$owner->queue("skill_use", \%args);
+	}
 }
 
 ##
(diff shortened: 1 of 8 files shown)


Henrybk:
Tested both mercenary and homunculus for about 4 hours each, using both basic attacks and attack skills, got no errors during that time, I believe this is safe for merge.

During testing however I noticed that in new clients the server not always sends the proper packets for homunculus info, which can lead to crashes, in my testing that happens only when homunculus is dead and you interact with another slave however.
I will try to fix that also but it is beyond the scope of this pull request.

Also, in the future I will move slave selfskills to their AI in another PR.

alisonrag (approved):
LGTM

Issue #3919, opened 2024-12-26
Homunculus state crash


Henrybk:
In the new clients and servers when you log in and your homunculus is dead or vaporized sometimes the server won't send any info about the homunculus, not even that it is dead or vaporized, that is a problem in itself that need a solution, however there is a bigger issue, that it can cause crashes and incorrect checks in openkore as seen below.

![3cfe631a4e1393c0fb4c751d10ed295e](https://github.com/user-attachments/assets/2b366eb4-d75c-4f73-a7ce-17b56d9ba267)

In this example I had an alchemist farming using a mercenary bowman lv5 in metaling and randomly received a homunculus_property packet causing openkore to crash.

So there are 2 issues:

1. What to do when the server sends no info about the homunculus? Considerar it dead? vaporized? neither and don't bother? create a config key for the user to decide what to do in this case?
2. Fix kore so it doesn't crash in this scenario

I will try to fix both issues in the future

PR #3920, opened 2024-12-26
Fix homunculus state crashes

Henrybk:
This fixes issue 2 of #3919

Issue 1 must still be fixed.

I think a good solution would be to add 2 new checks for checkSelfCondition:
`_homunculus_noinfo_dead: is only positive when $char->{homunculus_info}{dead} is undefined`
`_homunculus_noinfo_resting: is only positive when $char->{homunculus_info}{vaporized} is undefined`

So we could make some useSelfSkills cleverly to fix the issue like:
useSelf_skill Resurrect Homunculus {
	homunculus 0
	homunculus_noinfo_dead 1
}

If it works, great, homun is alive, if it fails we now know $char->{homunculus_info}{dead} is 0. Then we can use:

useSelf_skill Call Homunculus {
	homunculus 0
	homunculus_noinfo_resting 1
}

If it works, great, homun is alive, if it fails we now know $char->{homunculus_info}{vaporized} is 0.

### DIFF
--- a/control/config.txt
+++ b/control/config.txt
@@ -535,6 +535,8 @@ attackSkillSlot {
 	homunculus_sp
 	homunculus_dead
 	homunculus_resting
+	homunculus_noinfo_dead
+	homunculus_noinfo_resting
 	homunculus_onAction
 	homunculus_notOnAction
 	homunculus_whenIdle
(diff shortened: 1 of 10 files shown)


alisonrag (approved):

ya4ept:
@Henrybk please add a description on this page:
https://openkore.com/wiki/homunculus_noinfo_dead

PR #3936, opened 2025-01-09
Fix eventmacro inventory

Henrybk:
Inventory conditions were permanently in stand by since the new inventory list ready packets.


ya4ept (approved):
looks good

Issue #3804, opened 2023-10-05
[eventMacro] QuestHuntCompleted not work if there is no specific monster


gghwanderland:
i use this event macro for world bounty, but its not triggering to make the event macro run.. can someone share idea or help me with this? thanks
[eventMacros.txt](https://github.com/OpenKore/openkore/files/12812741/eventMacros.txt)

automacro test1 {
exclusive 1
priority 1

macro_delay 0.1
delay 0.1

timeout 15
BaseLevel > 10
QuestHuntCompleted 8662 1004
QuestHuntCompleted 8662 1051
QuestHuntCompleted 8662 1048

call {
	if (&questCompleteCount(8662) == 1 && &questCompleteCount(8662) == 1 && &questCompleteCount(8662) == 1) call test1
	log Quest not Completed yet
	}
}

macro test1 {
	log All quests completed!
	do e flg3
	pause 3
	do move prt_fild05 366 209
	pause 1
	do talknpc 367 209 c n
	pause 1
	do talknpc 367 209 c r0 c c c r0 c r0 n
	release test1
}

rbizaare:
Would you mind sharing GGH's recvpackets? I will cross check it with their sister server (landverse).

gghwanderland:
i think you cant check it on landverse

AutoLoot03:
> i think you cant check it on landverse

can you lend him so he can try checking?

gghwanderland:
i dont know how to share  recvpackets.. where can i find it?

gghwanderland:
> Would you mind sharing GGH's recvpackets? I will cross check it with their sister server (landverse).

where can i find this?

PR #3938, opened 2025-01-09
Fix eventMacro quests

### LINKED ISSUE #3934: Quest missions undefined key
Quest missions has as undefined key sometimes (always?), tested in idRO.
This makes eventMacro and other similar checks fail

43e1a58c220a053e175eb56da0e00bae

Henrybk:
All eventMacro quest conditions relied on packet_mapChange and inventory_ready which either don't work always in xkore 3 (packet_mapChange) or don't work at all in new clientes (inventory_ready), this should fix them all.

Fixes #3934

### DIFF
--- a/plugins/eventMacro/eventMacro/Condition/QuestTimeOverdue.pm
+++ b/plugins/eventMacro/eventMacro/Condition/QuestTimeOverdue.pm
@@ -7,7 +7,7 @@ use eventMacro::Utilities qw( find_variable );
 use base 'eventMacro::Condition';
 
 sub _hooks {
-	['quest_list','quest_mission','quest_added','quest_update_mission','quest_delete','quest_active'];
+	['quest_all_list_end','quest_all_mission_end','quest_added','quest_update_mission_hunt_end','quest_delete','quest_active'];
 }
 
 sub _parse_syntax {
--- a/plugins/eventMacro/eventMacro/Condition/QuestTimeOverdue.pm
+++ b/plugins/eventMacro/eventMacro/Condition/QuestTimeOverdue.pm
@@ -6,9 +6,8 @@ use eventMacro::Utilities qw( find_variable );
 
 use base 'eventMacro::Condition';
 
-# Using 'inventory_ready' is not optimal, but it works.
 sub _hooks {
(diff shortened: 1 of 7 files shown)


src/Network/Receive.pm
```diff
@@ -4737,6 +4739,8 @@ sub quest_all_mission {
 			});
 		}
 	}
+
+	Plugins::callHook('quest_mission');
```
ya4ept, line 4743: I suggest renaming it to "quest_all_mission_end"

src/Network/Receive.pm
```diff
@@ -4906,13 +4910,17 @@ sub quest_update_mission_hunt {
 			goal => $quest_mission->{mob_goal}
 		});
 	}
+	
+	Plugins::callHook('quest_update_mission');
```
ya4ept, line 4914: I suggest renaming it to "quest_update_mission_hunt_end"

plugins/eventMacro/eventMacro/Condition/QuestTimeOverdue.pm
```diff
@@ -6,9 +6,8 @@ use eventMacro::Utilities qw( find_variable );
 
 use base 'eventMacro::Condition';
 
-# Using 'inventory_ready' is not optimal, but it works.
 sub _hooks {
-	['packet_mapChange','inventory_ready','packet/quest_all_list','packet/quest_all_list2','packet/quest_all_list3','packet/quest_all_mission','packet/quest_add','packet/quest_delete','packet/quest_update_mission_hunt','packet/quest_active'];
+	['quest_list','quest_mission','quest_added','quest_update_mission','quest_delete','quest_active'];
```
ya4ept, line 10: suggest renaming "quest_list" to "quest_all_list_end"

ya4ept (changes requested):

ya4ept:
@Henrybk  Does this fix issue #3804?

Henrybk:
QuestHuntCompleted by definition checks if a specific hunt mission is completed (count == goal), so it requires both quest ID and either mob ID or hunt ID. I could make another condition like QuestHuntCompletedAll which just returns true if all hunt missions of a specific quest are completed.

Both QuestHuntCompleted and QuestInactive were not working correctly so yes it fixes both of those issues also.

ya4ept (approved):
looks good

PR #3944, opened 2025-01-11
Prevent repeated move packets

Henrybk:
Currently Task Route sends multiple packets of character move to the same cell in a repeated manner, some servers really don't like that like idRO.

This change prevents the Task from sending a new move packet if it's destination is the same as the current character movement.

Also when this happens and you are in a meeting route to a targer, give back control to AI::Attack.


alisonrag (approved):

Henrybk:
A distance check in maproute was causing some issues, updated and it, cleaned the hook call and added a similar one to talknpc

PR #3953, opened 2025-01-16
chore: move skill target time to ai_v temp

### LINKED ISSUE #3948: ai_v dont clear attackSkillSlot_x_target_time after monster die, disappear or teleport
keep acumulating data:
image

alisonrag:
now this key is deleted on map change

fixes #3948


Henrybk:
Maybe we should also remove it when the target dies on screen?

Henrybk (approved):

hubertyang88:
it seems target_timeout doesn't work now

Issue #3949, opened 2025-01-14
Openkore dont attack the same target if drop it to attack other aggressive monster


alisonrag:
https://github.com/user-attachments/assets/c905ce04-5004-44f8-b692-33317b3fd230

PR #3954, opened 2025-01-16
fix(attack): updating get target algorithm

alisonrag:
- fix pick target by distance
- fix not pick target when it was dropped

fix: #3949


Henrybk (approved):

hubertyang88:
image
target changed several times, is this normal?

alisonrag:
should not be that normal, are u using bot in party?

hubertyang88:
Those bots are in a party, but they will not help each other
in config.txt
attackAuto_party 0
when not in a party , those attacking info were still showed

PR #3980, opened 2025-04-06
feat: adds go command support

alisonrag:
portals at portals_commands.txt are commented by default, if want to enable the atcommand go just remove the comment '#'


Henrybk:
Maybe we could also add the 'tele 2' method when saveMap is configured and we have a way to use it (Butterfly wing or teleport skill).
We should just add a check that if we teleported using tele 2 and we didn't end up at the configured saveMap we should unset it in config

Henrybk (approved):

Henrybk:
Tested and working well

ya4ept:
documentation:
https://openkore.com/wiki/routeweights.txt
https://openkore.com/wiki/Category:tables
https://openkore.com/wiki/portals_commands.txt

PR #3992, opened 2025-04-23
fix: standardizes some domain messages

alisonrag:


Henrybk (approved):

ya4ept:
I suggest replacing "xkoreProxy" with "XKoreProxy"

PR #4010, opened 2025-05-29
ROla

OvoKore:


src/Network/Receive.pm
```diff
@@ -1075,6 +1075,12 @@ sub parse_account_server_info {
 			types => 'a4 v Z20 v5',
 			keys => [qw(ip port name state users property sid unknown)],
 		};
+	} elsif ($args->{switch} eq '0C32' && $masterServer->{serverType} eq "ROla") { # ROla
```
alisonrag, line 1078: is 0c32 different in other server? if not remove the 'rola' check

OvoKore: imagem_2025-05-29_181823276

Issue #4057, opened 2025-07-20
teleportAuto_useSkill 0 ignored when character has teleport skill


nizdi:
### Openkore version git

https://github.com/OpenKore/openkore

### XKore Mode

XKore 3

### RO Server Name

roLA

### Link to RO server

https://www.gnjoylatam.com/pt

### Result

When teleportAuto_useSkill is set to 0, OpenKore still uses the Teleport skill if the character has it, instead of consuming items like Fly Wings. This happens even when the configuration explicitly disables skill usage.

In src/Task/Teleport/Random.pm and src/Task/Teleport/Respawn.pm, the function canUseSkill does not check the value of teleportAuto_useSkill. It only verifies if the actor has the skill and is not muted.

Update the canUseSkill function in both files to:

```
sub canUseSkill {
    my ($self) = @_;

    # Respect teleportAuto_useSkill
    if (defined $config{teleportAuto_useSkill}) {
        return 0 if $config{teleportAuto_useSkill} == 0;
        my $lvl_min = ($config{teleportAuto_useSkill} == 1) ? 1 : 0;
        my $have    = $self->{actor}->getSkillLevel(new Skill(handle => 'AL_TELEPORT')) || 0;
        return 0 if $have < $lvl_min;
    }

    return 0 if $self->{actor}->{muted};
    return $self->{actor}->getSkillLevel(new Skill(handle => 'AL_TELEPORT')) ? 1 : 0;
}
```

Why the current code ignores teleportAuto_useSkill 0

```
sub canUseSkill {
	my ($self) =  @_;
	return 0 if ($self->{actor}->{muted});
	return $self->{actor}->getSkillLevel(new Skill(handle => 'AL_TELEPORT')) ? 1 : 0;
}
```

The issue lies in the canUseSkill function inside both Task::Teleport::Random.pm and Task::Teleport::Respawn.pm. This function only checks whether the character is muted and whether they have the AL_TELEPORT skill. It does not check the teleportAuto_useSkill setting at all. As a result, if the character has the Teleport skill, the function will return true (1), causing the teleport task to always use the skill first — even if the config explicitly forbids it by setting teleportAuto_useSkill 0. The intended priority (e.g., preferring item use) is bypassed completely.

### Expected Result

The bot should never use the Teleport skill when teleportAuto_useSkill 0 is set, regardless of whether the skill is available.

### How to Reproduce

Create an Acolyte with Teleport skill.
Add to config.txt:
```
[...]

PR #4065, opened 2025-07-30
fix: adds check for useskill config key when teleport

alisonrag:
this fixes: #4057 

image


alisonrag:
how to test:
1- config.txt
```
teleportAuto_item1 Fly Wing
teleportAuto_useSkill 0|1
```
2- teleport skill enabled in character

then type `tele 1` in console

PR #4099, opened 2025-09-19
ROla storageAuto_npc prontera/payon

OvoKore:


src/AI/CoreLogic.pm
```diff
@@ -1490,6 +1490,9 @@ sub processAutoStorage {
 						if ($config{'storageAuto_npc_type'} eq "" || $config{'storageAuto_npc_type'} eq "1") {
 							warning T("Warning storageAuto has changed. Please read News.txt\n") if ($config{'storageAuto_npc_type'} eq "");
 							$steps = "c r1";
```
alisonrag, line 1492: add this to else or it will be set 2 times in some cases

alisonrag (approved):

PR #4129, opened 2025-12-03
fix: improve error handling in gat_to_fld2.pl

alisonrag:
Replaces exit calls with exceptions for file open errors and adds try/catch logic to continue processing maps even if one fails. Also adds more informative logging for conversion progress and unknown block types.


ya4ept (approved):
look good

PR #4150, opened 2026-03-02
Add saveMap warp, min-distance and route checks

boscv:
Introduce configurable warp-to-saveMap, min-distance behavior and route-based checks to avoid unnecessary teleports. Added new config entries (saveMap_warp, saveMap_warp_minDistance) and centralized warp eligibility into shouldUseWarpToSaveMapForBuyOrSell in CoreLogic. Implemented distance calculation and caching (getDistanceToSaveMap / getDistanceToSaveMapFromCurrentPosition) and improved Task::CalcMapRoute with helpers to resolve saveMap spawn destinations, estimate walk distances, and honor the min-distance threshold when populating warp options. Changes reduce redundant short-distance warps and improve routing accuracy while keeping fallbacks when route computation fails.

### DIFF
--- a/src/AI/CoreLogic.pm
+++ b/src/AI/CoreLogic.pm
@@ -1425,8 +1425,7 @@ sub processAutoStorage {
 				}
 
 				# If warpToBuyOrSell is set, warp to saveMap if we haven't done so
-				if ($config{'saveMap'} ne "" && $config{'saveMap_warpToBuyOrSell'} && !$args->{warpedToSave}
-				    && !$field->isCity && $config{'saveMap'} ne $field->baseName) {
+				if (shouldUseWarpToSaveMapForBuyOrSell($args)) {
 					if ($char->{sitting}) {
 						message T("Standing up to auto-storage\n"), "teleport";
 						ai_setSuspend(0);
@@ -1888,8 +1887,7 @@ sub processAutoSell {
 				undef $args->{'warpedToSave'};
 			}
 
-			if ($config{'saveMap'} ne "" && $config{'saveMap_warpToBuyOrSell'} && !$args->{'warpedToSave'}
-			&& !$field->isCity && $config{'saveMap'} ne $field->baseName) {
+			if (shouldUseWarpToSaveMapForBuyOrSell($args)) {
 				if ($char->{sitting}) {
 					message T("Standing up to auto-sell\n"), "teleport";
 					ai_setSuspend(0);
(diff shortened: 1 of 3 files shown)


src/AI/CoreLogic.pm
```diff
@@ -4054,6 +4047,78 @@ sub processFeed {
 	}
 }
 
+sub shouldUseWarpToSaveMapForBuyOrSell {
+	my ($args) = @_;
+
+	return 0 if (!$config{'saveMap'} || !$config{'saveMap_warpToBuyOrSell'} || $args->{warpedToSave});
+	return 0 if ($field->isCity || $config{'saveMap'} eq $field->baseName);
```
Henrybk: I know this '$field->isCity' has already been here for many years, but why? If we have sellauto set to prontera and we are in geffe, and our savemap is prontera, why shouldn't we teleport to savemap?

Henrybk:
Another thing that does not need to be resolved in this PR (but maybe you are inspired and want to do it), we add go commands and tele2 right at the start of openlist creation, however there might be situations were we are in a place where teleport is prohibited (eg. a guild dungeon), in these instances kore will try to TP/GOcommand out of the dungeon 5 times and fail, then walk all the way to savemap/target instead of going "dungeon > map right outside dungeon (eg. pay_gld) > teleport to savemap".

The solution would be to create a list, probably in tables, with all maps where teleport/GOcommand is prohibited and then only add in calcMapRoute the option to teleport/GOcommand when it is possible to, something like:

Start:
`Add pay_glddun > pay_gld portal to openlist`
`fail to add tele2 to openlist bc pay_glddun is a prohibited map`
`expand pay_glddun > pay_gld portal (removing it from openlist)`
`add all portals of pay_gld to openlist`
`if (can teleport in this map AND not yet added Tele2 in this route) { add Tele2 to openlist }`

billabong93:
> Another thing that does not need to be resolved in this PR (but maybe you are inspired and want to do it), we add go commands and tele2 right at the start of openlist creation, however there might be situations were we are in a place where teleport is prohibited (eg. a guild dungeon), in these instances kore will try to TP/GOcommand out of the dungeon 5 times and fail, then walk all the way to savemap/target instead of going "dungeon > map right outside dungeon (eg. pay_gld) > teleport to savemap".
> 
> 
> 
> The solution would be to create a list, probably in tables, with all maps where teleport/GOcommand is prohibited and then only add in calcMapRoute the option to teleport/GOcommand when it is possible to, something like:
> 
> 
> 
> Start:
> 
> `Add pay_glddun > pay_gld portal to openlist`
> 
> `fail to add tele2 to openlist bc pay_glddun is a prohibited map`
> 
> `expand pay_glddun > pay_gld portal (removing it from openlist)`
> 
> `add all portals of pay_gld to openlist`
> 
> `if (can teleport in this map AND not yet added Tele2 in this route) { add Tele2 to openlist }`
> 
> 

yeah, didnt thought that.. i'm aiming on doing exactly these kind of behavior improvements, i'll look into it. thanks again Henry!

Henrybk (approved):

PR #97, opened 2016-05-11
Corrected 'New' Char Deletion

cydh:
- [x] Code review
- [x] QA testing on idRO
- [x] QA testing on iRO

* Fixed `deleteDate` variable that should store formatted date by [current_time] + [deleteDate]
* Added `deleteDateTimestamp` variable for comparing delete date and current time before selecting/deleting char
* Implemented send packets:
  * `'0829' => ['char_delete2_accept', 'a4 a6', [qw(charID code)]], # 12`: Supposed for all clients.
  * `'098f' => ['char_delete2_accept', 'v a4 a*', [qw(length charID code)]]`: idRO.
* Implemented received packet:
  * `'082A' => ['char_delete2_accept_result', 'V V', [qw(charID result)]], # 10`: As answer of char_delete2_accept
* Fixes #93 

Signed-off-by: Cydh Ramdh <cydh@pservero.com>

PR #4222, opened 2026-08-09
Fix confusing "lefting" wording in character delete message

### LINKED ISSUE #128: Confusing language used when deleting characters
In the current character delete code, there is a lot of broken english and prompts that don't necessarilly make sense. For example, when a character is marked for deletion it says "your character is lefting [timestamp]" as in... "your character won't be deleted until [timestamp]"? At least that's what I think it means...

Also when you are prompted to confirm the deletion, you are asked for a birthday / security code, however, when you actually submit the information, it says the email you entered is incorrect.

prajwal-raj:
## Problem
When a character is scheduled for deletion but can't be deleted yet, OpenKore shows:

> It will be deleted lefting [date]!

"lefting" isn't valid English and confuses users into thinking deletion is imminent, when it actually means the character *cannot* be deleted until that date. This was reported in #128 back in 2016 and a prior attempt (#97) was never merged.

## Fix
Changed the message to clearly state the character cannot be deleted until that date, matching the tone of the sibling message right above it ("Deleting is possible since %s.").

Fixes #128


prajwal-raj:
Hi @sctnightcore  small, low-risk wording fix closing a long-standing issue (#128) reported back in 2016. Would appreciate a review when you have a moment!

alisonrag (approved):

